PR #43, opened 2022-03-30
Shibb v1

hectorcorrea:


hectorcorrea:
We are going with CAS instead of Shibboleth: https://github.com/pulibrary/pdc_describe/pull/45

Issue #33, opened 2022-03-28
A User can have an Orcid ID


bess:
- [x] A user object can have an orcid id as an attribute
- [x] After a user authenticates via CAS, they can reach a screen with their own information on it. One of the options on their home screen is to record their ORCID.

PR #48, opened 2022-04-07
Allow a logged in user to add an ORCiD

bess:
Work toward #33 

Screen Shot 2022-04-07 at 4 17 41 PM


jrgriffiniii (approved):
Thank you for this, this looks excellent to me! https://coveralls.io/builds/48101047/source?filename=app%2Fcontrollers%2Fusers_controller.rb seems to indicate that the coverage is failing, might this be a Coveralls/SimpleCov bug?

PR #54, opened 2022-04-13
Functionality to creates new datasets and assign a new ARK to them

hectorcorrea:
Allows a user to create a new dataset, links the dataset to the user and a default collection, and assigns a new ARK to the dataset.

I created rudimentary Dataset list and show views to allow the process to work through the UI.

I hard coded a couple of sample collections and notes that we want to revisit this to define real default collections (perhaps using the ones in DataSpace )

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class Collection < ApplicationRecord
+  def self.create_defaults
+    return if Collection.count > 0
+    # TODO: Define proper default collections, perhaps from DataSpace?
+    Collection.create(title: "Collection 1")
+    Collection.create(title: "Collection 2")
+  end
+end
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -4,10 +4,11 @@
 class Collection < ApplicationRecord
   def self.create_defaults
     return if Collection.count > 0
-    # TODO: Define proper default collections, perhaps from DataSpace?
     Rails.logger.info "Creating default Collections"
-    Collection.create(title: "Sample Collection 1")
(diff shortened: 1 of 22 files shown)


app/models/collection.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# rubocop:disable Style/NumericPredicate
+class Collection < ApplicationRecord
+  def self.create_defaults
+    return if Collection.count > 0
+    # TODO: Define proper default collections, perhaps from DataSpace?
```
bess, line 7: Per our architecture diagram, I think we have four default collections:
1. Research Data
2. PPPL
3. ETDs
4. Library Resources

bess (approved):

app/models/collection.rb
```diff
@@ -0,0 +1,14 @@
...
+
+# rubocop:disable Style/NumericPredicate
+class Collection < ApplicationRecord
+  def self.create_defaults
+    return if Collection.count > 0
+    Rails.logger.info "Creating default Collections"
+    Collection.create(title: "Research Data", code: "RD")
+    Collection.create(title: "Princeton Plasma Physics Laboratory ", code: "PPPL")
+    Collection.create(title: "Electronic Theses and Dissertations", code: "ETD")
+    Collection.create(title: "Library Resources", code: "LIB")
```
hectorcorrea, line 11: @bess - default collections. Notice that I added `code` field for us to refer to them regardless of the final name that they get. I did not add code to link the user to the proper default collection as part of this PR, though.

bess: 👍🏻

bess (approved):
🚀

Issue #28, opened 2022-03-22
Per collection user authorization


bess:
- [x] For every collection, a user is either a submitter, an approver, or neither
- [x] As a user, when I visit my own home screen I can see what collections I have submitter or admin rights on 
- [x] If I am NOT a submitter in a given collection then I can NOT create a new object there

PR #56, opened 2022-04-14
User default collection

hectorcorrea:
Adds support for users to have a default collection.

We currently calculate the default collection based on the department number for the user as reported by CAS.

Part of #28

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -10,5 +10,22 @@ class Collection < ApplicationRecord
     Collection.create(title: "Electronic Theses and Dissertations", code: "ETD")
     Collection.create(title: "Library Resources", code: "LIB")
   end
+
+  def self.default
+    Collection.where(code: "RD").first
+  end
+
+  def self.default_for_department(department_number)
+    # Reference: https://docs.google.com/spreadsheets/d/1_Elxs3Ex-2wCbbKUzD4ii3k16zx36sYf/edit#gid=1484576831
+    if department_number.nil?
+      default
+    elsif department_number >= "31000" && department_number <= "31026"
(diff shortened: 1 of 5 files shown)


app/models/collection.rb
```diff
@@ -10,5 +10,25 @@ def self.create_defaults
...
+  # Returns the default collection for a given department number.
+  def self.default_for_department(department_number)
+    # Reference: https://docs.google.com/spreadsheets/d/1_Elxs3Ex-2wCbbKUzD4ii3k16zx36sYf/edit#gid=1484576831
+    if department_number.nil?
+      default
+    elsif department_number >= "31000" && department_number <= "31026"
+      Collection.where(code: "PPPL").first
+    elsif department_number >= "41000" && department_number <= "41032"
+      Collection.where(code: "LIB").first
+    else
```
hectorcorrea, line 30: We'll tweak this logic as needed.

app/models/user.rb
```diff
@@ -16,19 +11,28 @@ def self.from_cas(access_token)
...
+      #   access_token.extra.departmentnumber (e.g. "41006")
       #   access_token.extra.givenname (e.g. "Harriet")
       #   access_token.extra.displayname (e.g. "Harriet Tubman")
       #
       user = User.new
       user.provider = access_token.provider
       user.uid = access_token.uid # this is the netid
       user.email = access_token.extra.mail
-      user.save
+      user.default_collection_id = Collection.default_for_department(access_token.extra.departmentnumber)&.id
```
hectorcorrea, line 22: Integration with CAS

bess (approved):

PR #57, opened 2022-04-15
Adds per collection user authorization

### LINKED ISSUE #28: Per collection user authorization

hectorcorrea:
Allows users to be designated as submitters, administrators, or none for collections. Users can only add new datasets to the collections that they are submitters (or administrators). 

Created a edit form for Datasets that allows users to change title, ark, and collection of a given dataset. The user can only selects from the collections where they are submitters (or admins). Superadmin users can view all collections.

Closes #28

A few screenshots showing the general idea

## Home page
home

## User's dashboard (aka my datasets)
dashboard

## Dataset edit
edit

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -41,3 +41,6 @@ RSpec/DescribeClass:
 
 RSpec/ExampleLength:
   Enabled: false
+
+Style/NumericPredicate:
+  Enabled: false
--- a/app/views/datasets/_form.html.erb
+++ b/app/views/datasets/_form.html.erb
@@ -1,0 +1,31 @@
+<%= form_with(model: dataset) do |form| %>
+  <% if dataset.errors.any? %>
+    <div id="error_explanation">
+      <h2><%= pluralize(dataset.errors.count, "error") %> prohibited this dataset from being saved:</h2>
+      <ul>
+        <% dataset.errors.each do |error| %>
+          <li><%= error.full_message %></li>
+        <% end %>
+      </ul>
+    </div>
+  <% end %>
+
+  <div class="field">
+    <%= form.label :title %><br>
+    <%= form.text_field :title %>
+  </div>
+
+  <div class="field">
+    <%= form.label :collection %><br>
+    <%= form.collection_select :collection_id, current_user.submitter_collections, :id, :title, {:include_blank => true}, {:class => 'form-control'} %>
+  </div>
+
+  <div class="field">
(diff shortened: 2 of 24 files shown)


.rubocop.yml
```diff
@@ -41,3 +41,6 @@ RSpec/DescribeClass:
 
 RSpec/ExampleLength:
   Enabled: false
+
+Style/NumericPredicate:
```
hectorcorrea, line 45: This is so that we can use `object.something > 0` like the rest of the world does.

app/views/datasets/_form.html.erb
```diff
@@ -0,0 +1,34 @@
...
+  <% end %>
+
+  <div class="field">
+    <%= form.label :title %><br>
+    <%= form.text_field :title %>
+  </div>
+
+  <div class="field">
+    <%= form.label :collection %><br>
+    <%= form.collection_select :collection_id, current_user.submitter_collections, :id, :title, {:include_blank => true}, {:class => 'form-control'} %>
```
hectorcorrea, line 20: Notice that the dropdown with the possible collections that a user can assign a dataset to is filtered to whatever they have access to submit to via the `current_user.submitter_collections`

spec/controllers/datasets_controller_spec.rb
```diff
@@ -10,21 +10,52 @@
     it "handles the index page" do
       sign_in user
       get :index
-      expect(response.status).to eq(200)
+      expect(response).to render_template("index")
```
hectorcorrea, line 13: The gem `rails-controller-testing` that I added on this PR allows to check that the proper view is rendered rather than just an HTTP 200 (I stole this from Orangelight, I think)

app/views/datasets/_form.html.erb
bess: 👍🏻

spec/controllers/datasets_controller_spec.rb
bess: 💜

.rubocop.yml
bess: 👍🏻

bess (approved):
🚀

PR #72, opened 2022-04-21
60 user routes

### LINKED ISSUE #60: User home page route
Create a route to go a user page by using their netid instead of the internal Rails id, e.g. `/user/hc8719` instead of `/user/2`

kelynch:
Closes #60

### DIFF
--- a/spec/factories/user.rb
+++ b/spec/factories/user.rb
@@ -2,8 +2,8 @@
 
 FactoryBot.define do
   factory :user do
-    sequence(:uid) { FFaker::Internet.user_name }
-    sequence(:email) { FFaker::Internet.email }
+    sequence(:uid) { FFaker::InternetSE.login_user_name }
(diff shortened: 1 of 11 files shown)


spec/factories/user.rb
```diff
@@ -2,8 +2,8 @@
 
 FactoryBot.define do
   factory :user do
-    sequence(:uid) { FFaker::Internet.user_name }
-    sequence(:email) { FFaker::Internet.email }
+    sequence(:uid) { FFaker::InternetSE.login_user_name }
```
kelynch, line 5: I switched to this as it removes the period from the username, which is breaking the `user_url` method.  This alternative method in the FFaker gem accounts for it - https://github.com/ffaker/ffaker/blob/main/lib/ffaker/internet_se.rb#L33

hectorcorrea: wow! what a catch!

PR #86, opened 2022-04-27
Collection management

hectorcorrea:
Added the pages to list, view, and edit collections. Added functionality to add/remove submitters and administrators to a collection along with the proper checks for who can do what and where.

Added a rake task to populate the `User` and `UserCollection` tables with the data in `collection_defaults.yml` so that we can initialize a new database with some sensible defaults. Also added rake task to output to the console the information required to regenerate the `collection_defaults.yml` with the current information on the database (so that we can re-import this data into another database if needed, e.g. from staging to production.)

Screen Shot 2022-04-28 at 2 20 39 PM


app/models/user.rb
```diff
@@ -13,30 +13,75 @@ class User < ApplicationRecord
...
+        UserCollection.add_admin(user.id, collection.id)
+      end
+
+      Rails.logger.info "Setting up submitters for collection #{collection.title}"
+      collection.default_submitters_list.each do |uid|
+        user = User.new_for_uid(uid)
+        UserCollection.add_submitter(user.id, collection.id)
+      end
+    end
+  end
```
hectorcorrea, line 83: When we create the default user access we create the proper User and UserCollection records in the database with the data in the `collection_defaults.yml`. These are the records that won't have a "provider" value and will be updated with the data from CAS when the user log into the system.

app/views/collections/show.html.erb
```diff
@@ -0,0 +1,186 @@
...
+        <a class="delete-user-from-collection" data-collection-id="${collectionId}" data-uid="${uid}"
+          href="#" title="Revoke user's ${role} access this collection">
+          <i class="bi bi-trash delete_icon" data-collection-id="${collectionId}" data-uid="${uid}"></i>
+        </a>
+      </span>`;
+    }
+    $(elList).append(html);
+  }
+
+  // Issues the HTTP POST to add a user to a collection and updates the UI
```
hectorcorrea, line 109: Adding and removing submitters and administrators to the collection is done via AJAX calls, so the screen does not jump/refresh every time a new submitter or administrator is added. These options are only available when the user has the proper admin rights to the collection.

app/views/collections/show.html.erb
```diff
@@ -0,0 +1,186 @@
...
+      error: function(x) {
+        alert(x.responseJSON.message);
+      }
+    });
+  }
+
+  // Adds the information for a given user to the lists of administrators/submitters
+  // for the collection.
+  // `elList` is the reference to the <ul> HTML element that hosts the list.
+  var addUserHtml = function(elList, uid, collectionId, role, canDelete, isYou) {
```
hectorcorrea, line 90: It might be a bit unorthodox to add the list of submitters and administrators on the page via JavaScript, but with this approach we have a *single logic* to display this information regardless of whether we are displaying the current data on the database vs the new records added on the fly by the user (this is in contrast to displaying the information sometimes with Ruby code and other times with JavaScript code.)

kelynch (approved):

app/models/user.rb
```diff
@@ -13,30 +13,75 @@ class User < ApplicationRecord
   end
 
   def self.from_cas(access_token)
```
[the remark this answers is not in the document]
bess: 🚀

bess (approved):

Issue #78, opened 2022-04-22
Display file metadata on dataset show page


bess:
For datasets, on the show page, display metadata for any files associated with the DOI that are stored in S3. 

- [x] Filenames
- [x] File sizes
- [x] Last edited date

bess:
![Screen Shot 2022-04-28 at 3 41 45 PM](https://user-images.githubusercontent.com/65608/165836076-36ea5216-328b-493f-91d8-8814025d2786.png)

PR #88, opened 2022-04-28
Show files in S3 on dataset show page

bess:
Also: Add configuration so we do not update ARK url unless we explicitly
say we're ready.

Work toward #78


bess:
Screen Shot 2022-04-28 at 3 41 45 PM

PR #92, opened 2022-04-28
Implements support for importing Dublin Core metadata into Works from the OAI-PMH (for handling Library PDF Items in DataSpace)

### LINKED ISSUE #80: Make fixtures for Library PDFs
Use the samples identified here - https://docs.google.com/document/d/18ZkBldqWxIIR1UA6qMY87RnGFTKU9HG3EJzodzzFf2A/edit#bookmark=id.o3vcemdfz6qb

jrgriffiniii:
Resolves #80 with the following:

- Dublin Core metadata into Works from the OAI-PMH
- Implementing the Metadata::Base, Metadata::Document, Metadata::DublinCore, Metadata::OaiPmhDocument for parsing the XML metadata
- Adding a `json` attribute to the Work Model for storing Dublin Core metadata (along with a decorating DublinCore Class)


app/values/metadata/document.rb
```diff
@@ -0,0 +1,28 @@
...
+      @attributes = {}
+
+      self.class.attribute_xpaths.each_pair do |attr, xpath|
+        elements = root.xpath(xpath, self.class.namespaces)
+        attributes[attr] = elements.map(&:content)
+      end
+    end
+
+    delegate :each_pair, to: :attributes
+    delegate :root, to: :@document
```
hectorcorrea, line 18: I did not know you can delegate to a class attribute. Nice.

db/schema.rb
```diff
@@ -77,6 +77,7 @@
     t.integer "created_by_user_id"
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
+    t.json "dublin_core"
```
hectorcorrea, line 80: I wonder if this should have gone into the `Dataset` table rather than `Work`...or perhaps we don't even need the `Dataset` table.

hectorcorrea: The `Dataset` table has a field to indicate the profile (e.g. DublinCore vs Data Cite) and I am wondering if we should move that field to the `Work` table.

hectorcorrea (dismissed):
Left a few comments for discussion but nothing blocking.

db/schema.rb
hectorcorrea: See ERD: https://docs.google.com/drawings/d/1q2sfj8rrcNVgqQPK5uT_t79A9SYqncinh3HbnCSGMyQ/edit

kelynch (approved):

kelynch (approved):

PR #95, opened 2022-04-29
Better handle the creation of the default UserCollection records

### LINKED ISSUE #90: Login process and default_collection
Bess ran into an issue in which her user record had a `default_collection_id = nil` that prevented her from logging in.

The culprit seems to be that the default Collections were not added to the database at the time her record was created and therefore the system could not determine what to assign.

We should update the logic to handle more gracefully when the Collections are not in the database (e.g. create then on the spot).

See stacktrace here: https://app.honeybadger.io/projects/99214/faults/85153455

hectorcorrea:
Closes #90

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -103,6 +103,12 @@ class User < ApplicationRecord
 
   # Adds the user to the collections that they should have access by default
   def setup_user_default_collections
+    # No need to add records for super admins.
+    return if superadmin?
+
+    # Nothing to do in this case (this should never happen, but it did once so...)
+    return if default_collection_id.nil?
+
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -103,6 +103,12 @@ def default_collection
 
   # Adds the user to the collections that they should have access by default
   def setup_user_default_collections
+    # No need to add records for super admins.
+    return if superadmin?
+
+    # Nothing to do in this case (this should never happen, but it did once so...)
+    return if default_collection_id.nil?
```
jrgriffiniii, line 110: 👍

jrgriffiniii (approved):

PR #119, opened 2022-05-12
Edit form with Title and Creators

hectorcorrea:
Allows adding more than one title and more than one creator.

titles

creators

Moves #112 and #114 forward.

### DIFF
--- a/app/controllers/datasets_controller.rb
+++ b/app/controllers/datasets_controller.rb
@@ -13,7 +13,7 @@ class DatasetsController < ApplicationController
   def show
     @dataset = Dataset.find(params[:id])
     work = Work.find(@dataset.work_id)
-    @datacite = Datacite::Resource.new_from_json_string(work.data_cite)
+    @datacite = Datacite::Resource.new_from_json(work.data_cite)
 
     if @dataset.doi
       service = S3QueryService.new(@dataset.doi)
@@ -24,7 +24,7 @@ class DatasetsController < ApplicationController
   def edit
     @dataset = Dataset.find(params[:id])
     work = Work.find(@dataset.work_id)
-    @datacite = Datacite::Resource.new_from_json_string(work.data_cite)
+    @datacite = Datacite::Resource.new_from_json(work.data_cite)
   end
 
   # rubocop:disable Metrics/AbcSize
@@ -32,23 +32,7 @@ class DatasetsController < ApplicationController
   def update
     @dataset = Dataset.find(params[:id])
     respond_to do |format|
-      # Populate the data_cite field with the data from the data entry
-      work = Work.find(form_params[:work_id])
-      work_data = work_params
-      resource = Datacite::Resource.new(title: form_params[:title])
-
-      if params["alternative_title"].present?
-        resource.titles << Datacite::Title.new(title: params["alternative_title"], title_type: "AlternativeTitle")
-      end
-
-      if params["new_given_name"].present?
-        resource.creators << Datacite::Creator.new_person(params["new_given_name"], params["new_family_name"], params["new_orcid"])
-      end
-
-      work_data[:data_cite] = resource.to_json
-      work.update(work_data)
-      work.save!
-
+      update_work_record
       # And then update the dataset
       if @dataset.update(dataset_params)
         format.html { redirect_to dataset_url(@dataset), notice: "Dataset was successfully updated." }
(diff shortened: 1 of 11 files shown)


app/controllers/datasets_controller.rb
```diff
@@ -1,4 +1,9 @@
 # frozen_string_literal: true
+
+# rubocop:disable Metrics/ClassLength
+# rubocop:disable Metrics/AbcSize
+# rubocop:disable Metrics/MethodLength
+# rubocop:disable Style/For
```
hectorcorrea, line 6: I disabled a few warnings for now since I wanted to dump all the code in a single place for now. If this PR is accepted and the code is merged we'll re-activate the warnings once we start moving some of this code out of the controller.

app/controllers/datasets_controller.rb
```diff
@@ -59,21 +68,75 @@ def resubmit
     redirect_to dataset_path(dataset)
   end
 
+  # Outputs the Datacite XML representation of the dataset
+  def datacite
+    @dataset = Dataset.find(params[:id])
+    work = Work.find(@dataset.work_id)
+    resource = Datacite::Resource.new_from_json(work.data_cite)
+    render xml: resource.to_xml
+  end
```
hectorcorrea, line 77: This was a bonus feature, it outputs the Datacite XML for a given dataset, which turns out will be very beneficial for Matt and Neggin.

app/controllers/datasets_controller.rb
```diff
@@ -59,21 +68,75 @@ def resubmit
...
+      end
+
+      for i in 1..params["new_creator_count"].to_i do
+        creator = new_creator(params["new_given_name_#{i}"], params["new_family_name_#{i}"], params["new_orcid_#{i}"])
+        resource.creators << creator unless creator.nil?
+      end
+
+      work_data[:data_cite] = resource.to_json
+      work.update(work_data)
+      work.save!
```
hectorcorrea, line 136: This is really the important part of this PR. It process the data from the form (including multiple titles and multiple authors), pushes the data to a Datacite::Resource object. This resource is serialized to JSON and saved to Postgress.

app/views/datasets/_form.html.erb
```diff
@@ -21,39 +35,103 @@
...
+    <!--
+      We use these hidden fields to keep track of how many creators and titles are on the form
+      initially and how many are added on the fly. This helps to process the data when the form
+      is submitted.
+    -->
+    <div>
+      <input type="text" id="existing_title_count" name="existing_title_count" value="<%= @datacite.titles.count %>" class="hidden" />
+      <input type="text" id="new_title_count" name="new_title_count" value="0" class="hidden" />
+      <input type="text" id="existing_creator_count" name="existing_creator_count" value="<%= @datacite.creators.count %>" class="hidden" />
+      <input type="text" id="new_creator_count" name="new_creator_count" value="0" class="hidden" />
```
hectorcorrea, line 72: `existing_title_count` and `new_title_count` are used to track how many titles were on the form and how many the user added. This way when the user clicks Save the logic in the controller is pretty simple to parse the data that comes in the HTML form. The `existing_creator_count` and `new_creator_count` are used the same way but to track creators.

app/controllers/datasets_controller.rb
```diff
@@ -20,16 +28,17 @@ def show
 
   def edit
     @dataset = Dataset.find(params[:id])
+    work = Work.find(@dataset.work_id)
+    @datacite = Datacite::Resource.new_from_json(work.data_cite)
```
hectorcorrea, line 32: FUTURE UPDATE: I am thinking that our current database structure with a `Work` and a `Dataset` table might be adding extra complexity to our code. 

I am thinking that we should ditch the `Dataset` table (since it has a 1-1 relationship with the Work) and just keep the `Work` table. We could move the few fields in the Dataset table to the Work table and keep the code simpler.

app/controllers/datasets_controller.rb
bess: 👍🏻

app/controllers/datasets_controller.rb
bess: +1 for this idea.

app/controllers/datasets_controller.rb
bess: 💯

bess (approved):

PR #126, opened 2022-05-17
Dropping Dataset in favor of Work table

hectorcorrea:
First pass at dropping the Dataset table and moving the functionality to the Work table. System runs, tests have not been updated.


config/routes.rb
```diff
@@ -12,11 +12,11 @@
...
-  post "dataset/:id/approve", to: "datasets#approve", as: :approve_dataset
-  post "dataset/:id/withdraw", to: "datasets#withdraw", as: :withdraw_dataset
-  post "dataset/:id/resubmit", to: "datasets#resubmit", as: :resubmit_dataset
-  get "datasets/:id/datacite", to: "datasets#datacite", as: :dataset_datacite
-  resources :datasets
+  post "work/:id/approve", to: "works#approve", as: :approve_work
+  post "work/:id/withdraw", to: "works#withdraw", as: :withdraw_work
+  post "work/:id/resubmit", to: "works#resubmit", as: :resubmit_work
+  get "works/:id/datacite", to: "works#datacite", as: :dataset_work
+  resources :works
```
hectorcorrea, line 19: I am not sure if we want to the URLs to read `/dataset/x` or `/work/x` but for now I made them match the new Work controller to keep things Rails-friendly. We could adjust this if we need to.

spec/factories/work.rb
```diff
@@ -5,54 +5,72 @@
     factory :shakespeare_and_company_work do
       title { "Shakespeare and Company Project Dataset: Lending Library Members, Books, Events" }
       collection { Collection.research_data }
+      doi { "https://doi.org/10.34770/pe9w-x904" }
+      ark { "ark:/88435/dsp01zc77st047" }
```
hectorcorrea, line 9: All the `Work` factories now have these two fields. Notice that I deleted the `Dataset` factories that were the one to one mapping to these Works.

app/models/work.rb
```diff
@@ -10,20 +13,59 @@ class Work < ApplicationRecord
...
+    work = Work.new(
+      title: title,
+      created_by_user_id: user_id,
+      collection_id: collection_id,
+      work_type: "DATASET",
+      state: "AWAITING-APPROVAL",
+      profile: "DATACITE"
+    )
+    work.save!
+    work
```
hectorcorrea, line 68: This is just a convenience method (a sibling to the existing `created_skeleton()`) that defaults the parameters for Datacite datasets. Simplifies the calls to create a "dataset" which is really just a Work of type DATASET with the DATACITE profile.

app/controllers/works_controller.rb
```diff
@@ -4,109 +4,88 @@
...
+    work.resubmit(current_user)
+    redirect_to work_path(work)
   end
 
-  # Outputs the Datacite XML representation of the dataset
+  # Outputs the Datacite XML representation of the work
   def datacite
-    @dataset = Dataset.find(params[:id])
-    work = Work.find(@dataset.work_id)
+    work = Work.find(params[:id])
```
hectorcorrea, line 75: This is example of the logic that is not needed anymore. We don't have to look up the `dataset` and then the `work`. The work is all we need.

app/models/ark.rb
```diff
@@ -21,6 +21,8 @@ def self.find(ezid)
   # @param [ezid] [String] the EZID being validated
   # @return [Boolean]
   def self.valid?(ezid)
+    # Always consider test ARKs valid
+    return true if ezid.start_with?(EZID_TEST_SHOULDER)
```
hectorcorrea, line 25: @jrgriffiniii - I still had to add this to the check since we have ARKs with the "ark:/99999" prefix. I also renamed the constant from `EZID_INVALID_SHOULDER` to `EZID_TEST_SHOULDER` otherwise the code will read weird if it says "return true when it has invalid shoulder"

app/controllers/users_controller.rb
```diff
@@ -6,12 +6,12 @@ class UsersController < ApplicationController
   def show
     @can_edit = can_edit?
     if current_user.id == @user.id
-      @my_datasets = Dataset.my_datasets(current_user)
-      @awaiting_datasets = Dataset.admin_awaiting_datasets(current_user)
-      @withdrawn_datasets = Dataset.admin_withdrawn_datasets(current_user)
+      @my_works = Work.my_works(current_user)
```
jrgriffiniii, line 9: 👍 polymorphism may also be used here should it ever be determined that some object less generic than `Work` need be used.

app/models/ark.rb
jrgriffiniii: I am sorry for not addressing this case.

app/models/work.rb
```diff
@@ -10,20 +13,59 @@ class Work < ApplicationRecord
     end
   end
 
-  def self.create_skeleton(title, user_id, collection_id, work_type)
+  before_update do |ds|
+    if dublin_core.present?
```
jrgriffiniii, line 17: 👍

PR #128, opened 2022-05-18
Title validation

### LINKED ISSUE #114: Submit a title
Populate the "Title" tab with a form element, allowing someone to provide a title for their data set. 
- [x] I can submit the form and the dataset has this title. 
- [ ] I can not submit the form if the title (which is a required field) is not populated
- [ ] If I submit the form with a missing title, it tells me the title is missing

hectorcorrea:
Prevents an empty title

Screen Shot 2022-05-18 at 2 59 18 PM

Closes #114

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -17,8 +17,6 @@ class WorksController < ApplicationController
 
   def show
     @work = Work.find(params[:id])
-    @datacite = Datacite::Resource.new_from_json(@work.data_cite)
-
     if @work.doi
       service = S3QueryService.new(@work.doi)
       @files = service.data_profile
@@ -27,7 +25,6 @@ class WorksController < ApplicationController
 
   def edit
     @work = Work.find(params[:id])
-    @datacite = Datacite::Resource.new_from_json(@work.data_cite)
   end
 
   def update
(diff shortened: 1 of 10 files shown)


app/controllers/works_controller.rb
```diff
@@ -17,8 +17,6 @@ def new
 
   def show
     @work = Work.find(params[:id])
-    @datacite = Datacite::Resource.new_from_json(@work.data_cite)
```
hectorcorrea, line 20: I also changed the references to the standalone `@datacite` to `@work.datacite_resource` in several files associated with the Work data entry form. Strictly speaking that is not part of the title validation but it was easy to do and makes the code less dependent on an extra variable since we are now using a property in the `@work` object.

app/views/works/edit.html.erb
```diff
@@ -1,6 +1,6 @@
 <h1>Editing Dataset</h1>
 
-<%= render 'form', dataset: @dataset %>
```
hectorcorrea, line 3: Oops...looks like I forgot to remove this in a previous PR when I ditched the "dataset" model and controller.

app/controllers/works_controller.rb
bess: 👍🏻

bess (approved):

Issue #107, opened 2022-05-09
SPIKE: Retrieve data from ORCiD


bess:

hectorcorrea:
ORCID API: https://info.orcid.org/documentation/api-tutorials/api-tutorial-read-data-on-a-record/

PR #134, opened 2022-05-19
Fetch name information from ORCID

hectorcorrea:
A prototype on how we could fetch name information from ORCID via their public API (https://pub.orcid.org/v3.0/). Notice that API does not require authentication and that the call is made straight from the browser.

When a user enters an ORCID the systems makes an API call to ORCID and fetches the given name and family name. See screenshot below:

Screen Shot 2022-05-19 at 3 42 46 PM

It works really well, but it can get confusing easily when the person in the ORCID uses a "Published name" that is different from their "name". In this case [notice how Rebecca](https://orcid.org/0000-0002-8762-8057) uses "Rebecca Sutton Koeser" as published name and "Rebecca Koeser" as name. This seems to be a feature of ORCID (read more [here](https://support.orcid.org/hc/en-us/articles/360006973853)) but I am sure it will cause confusion with our users. 

There are no tests for this feature/prototype since I have no idea how to test that an AJAX call is made when the value is entered on the textbox on the form...and honestly I am not sure it's worth adding an entire testing machinery to test something as small as this.

Related to #107


carolyncole:
@hectorcorrea This is cool!

I believe a test would be good, as we would certainly want to keep this working.  For how to test see 
https://github.com/pulibrary/orangelight/blob/5f3e184cd31a4c4dc4832feee76b2c1440c4b28d/spec/support/stub_helpers.rb#L35-L36 & 
https://github.com/pulibrary/orangelight/blob/main/spec/features/requests/request_spec.rb#L826-L827

You would set up the stub then trigger the event and then check to make sure the stub was activated...

spec/support/orcid_specs.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+require "sinatra/base"
+
+# This sets up a little local server to respond to ORCID ajax requests
```
hectorcorrea, line 4: wow!

Issue #108, opened 2022-05-09
SPIKE: DataCite DOI minting


bess:
As a data curator, I want to be able to mint a DOI quickly and easily without going to a separate website, so that I can curate data faster and with fewer mental interruptions. 

## Acceptance Criteria
- [ ] Use the DataCite API to mint a DOI
- [ ] Evaluate `datacite-ruby` to see if that makes the job easier: https://github.com/sul-dlss/datacite-ruby

If `datacite-ruby` does not work, consider evaluating Serrano: it appears to be a maintained ruby client for using crossref apis. https://github.com/sckott/serrano

PR #144, opened 2022-05-26
Adding the datacite-ruby gem and creating a test spike

carolyncole:
We really need to get a hold of an account on [the test doi system](https://doi.datacite.org/) for testing the ability to publish DOIs, but this spike does create draft DOIs in the live system if the client is switched and environment variables are set.  I have deleted the two DOIs that I created in the live system.  Draft DOIs can be deleted, but published ones can not.

The JSON response in the test was taken directly from the gem test.

refs #108

### DIFF
--- a/config/initializers/inflections.rb
+++ b/config/initializers/inflections.rb
@@ -15,3 +15,6 @@
 # ActiveSupport::Inflector.inflections(:en) do |inflect|
 #   inflect.acronym 'RESTful'
 # end
+ActiveSupport::Inflector.inflections(:en) do |inflect|
+  inflect.acronym "PUL"
+end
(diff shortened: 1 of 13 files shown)


config/initializers/inflections.rb
```diff
@@ -15,3 +15,6 @@
 # ActiveSupport::Inflector.inflections(:en) do |inflect|
 #   inflect.acronym 'RESTful'
 # end
+ActiveSupport::Inflector.inflections(:en) do |inflect|
+  inflect.acronym "PUL"
+end
```
hectorcorrea, line 20: I did not know you could do this!

hectorcorrea (approved):

config/initializers/inflections.rb
kelynch: So cool!

PR #163, opened 2022-06-15
First pass at allowing users to reorder the creators

hectorcorrea:
Allows a user to change the order of the creators. Data is saved and rendered respecting the order.

Below is an animated GIF showing it in action (it's a bit slow, but it does animate)

reorder

Note the the user cannot pick the creators table header (ORCID/Given name/Family name) and drag it, but they can put a creator row on top of the header as they drag it. It looks ugly if they do, but it causes no harm. I did not find an easy way to prevent this behavior so I am submitting this PR with this possible bug, but I think it's OK.

### DIFF
--- a/spec/controllers/works_controller_spec.rb
+++ b/spec/controllers/works_controller_spec.rb
@@ -69,6 +69,55 @@ RSpec.describe WorksController, mock_ezid_api: true do
       expect(response.location).to eq "http://test.host/works/#{work.id}"
     end
 
+    it "handles the update page" do
+      params = {
+        "title_main" => "test dataset updated",
+        "collection_id" => work.collection.id,
+        "commit" => "Update Dataset",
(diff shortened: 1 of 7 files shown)


spec/controllers/works_controller_spec.rb
```diff
@@ -69,6 +69,55 @@
...
+        "action" => "update",
+        "id" => work.id.to_s,
+        "publisher" => "Princeton University",
+        "publication_year" => "2022",
+        "given_name_1" => "Jane",
+        "family_name_1" => "Smith",
+        "sequence_1" => "1",
+        "given_name_2" => "Ada",
+        "family_name_2" => "Lovelace",
+        "sequence_2" => "2",
```
hectorcorrea, line 87: Original creators order...

PR #170, opened 2022-06-22
Preserves the ability to click on the edit textboxes

### LINKED ISSUE #169: Creators in the data entry form and reordering
With the introduction of the option to reorder creators it is now much harder to enter the data for creators. It looks like the plug-in that allows drag and drop of the creators it taking precedence when the user clicks on the textbox to enter creators. 

See if we can either localize the plug-in so the icon to reorder (rather than the entire "creator row" in the table) or another way of not messing up with the data entry.

hectorcorrea:
Turns out the jQuery plug-in has a way to account for the issue by telling it not to kick-in on certain elements. In our case I told it not to kick in on textboxes or the delete icon.

Fixes #169

### DIFF
--- a/app/views/works/_edit_javascript.html.erb
+++ b/app/views/works/_edit_javascript.html.erb
@@ -172,9 +172,15 @@
     });
 
     // Allows the creators to be reordered via drag and drop.
-    // The `cancel` property is used to prevent reordering the header (https://stackoverflow.com/a/17897706/446681)
+    // The `cancel` property "prevents sorting if you start on elements matching the selector"
+    // https://api.jqueryui.com/sortable/#method-cancel
+    //
+    //  tr:has(th)      - prevents reordering the header (https://stackoverflow.com/a/17897706/446681)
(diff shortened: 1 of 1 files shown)


app/views/works/_edit_javascript.html.erb
```diff
@@ -172,9 +172,15 @@
     });
 
     // Allows the creators to be reordered via drag and drop.
-    // The `cancel` property is used to prevent reordering the header (https://stackoverflow.com/a/17897706/446681)
+    // The `cancel` property "prevents sorting if you start on elements matching the selector"
+    // https://api.jqueryui.com/sortable/#method-cancel
+    //
+    //  tr:has(th)      - prevents reordering the header (https://stackoverflow.com/a/17897706/446681)
+    //  input           - prevents reordering on the textboxes (so they are still editable)
+    //  .delete-creator - prevents reording on the delete icon
```
jrgriffiniii, line 180: 👍 thank you for explaining these selectors within the comments!

Issue #150, opened 2022-06-02
ORCID regex


hectorcorrea:
Update the RegEx that we use to validate ORCIDs to use the one here:

https://gist.github.com/asencis/644f174855899b873131c2cabcebeb87

Notice that this one account for the fact that an ORCID can end with an X.

Our current RegEx is here: https://github.com/pulibrary/pdc_describe/blob/main/app/views/works/_edit_javascript.html.erb#L4

PR #173, opened 2022-06-23
Tuned the ORCID validation

hectorcorrea:
Several minor tweaks including allowing spaces at the end of the ORCID (this is useful when users copy and paste an ORCID) and removing these extra spaces before saving to the database.

I am also using a slightly different regex that allows an `X` as the last "digit" since this is apparently OK.

Notice that we do the ORCID validation both in JavaScript and in Ruby and therefore the regex is in two places. This is OK and no big deal.

We also have the regex twice in JavaScript and at some *future PR* I would like to change that so that the JavaScript `isOrcid()` function is defined only once.

Fixes https://github.com/pulibrary/pdc_describe/issues/150

### DIFF
--- a/app/models/orcid.rb
+++ b/app/models/orcid.rb
@@ -2,7 +2,9 @@
 require "ezid-client"
 
 class Orcid
-  ORCID_REGEX = /^\d\d\d\d-\d\d\d\d-\d\d\d\d-\d\d\d\d$/.freeze
+  # Notice that we allow for an "X" as the last digit.
+  # Source https://gist.github.com/asencis/644f174855899b873131c2cabcebeb87
+  ORCID_REGEX = /^(\d{4}-){3}\d{3}(\d|X)$/.freeze
 
   def self.valid?(orcid)
     return false if orcid.blank?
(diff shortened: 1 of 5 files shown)


app/views/works/_edit_javascript.html.erb
```diff
@@ -63,8 +64,6 @@
             $(rowToDelete).remove();
           }
         }
-      } else {
-        console.log(`${rowToDelete} does not exist - nothing to do`);
```
hectorcorrea, line 67: per your request @jrgriffiniii

app/models/orcid.rb
```diff
@@ -2,7 +2,9 @@
 require "ezid-client"
 
 class Orcid
-  ORCID_REGEX = /^\d\d\d\d-\d\d\d\d-\d\d\d\d-\d\d\d\d$/.freeze
+  # Notice that we allow for an "X" as the last digit.
+  # Source https://gist.github.com/asencis/644f174855899b873131c2cabcebeb87
```
jrgriffiniii, line 6: 👍

jrgriffiniii (approved):

PR #174, opened 2022-06-23
Added ORCID format validation for the dataset creators

### LINKED ISSUE #151: ORCID validation in edit form
Make sure we validate the ORCID entered under the Creators tab to make sure it looks like an ORCID. 

The ORCID does not have to be valid/exist, we only want to make sure it has the correct format.

See also https://github.com/pulibrary/pdc_describe/issues/150

hectorcorrea:
Added the ORCID format validation for each of the creators of a dataset. 

It is OK for a creator to not have an ORCID, but if the creator has an ORCID it must be in the correct format e.f. 0000-0000-0000-0000-0000.

Closes #151

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -106,9 +106,17 @@ class Work < ApplicationRecord
 
     if work.data_cite.present?
       work.errors.add(:base, "Must provide a title") if work.title.blank?
-      work.errors.add(:base, "Must provide at least one Creator") if work.datacite_resource.creators.count == 0
       work.errors.add(:base, "Must indicate the Publisher") if work.datacite_resource.publisher.blank?
       work.errors.add(:base, "Must indicate the Publication Year") if work.datacite_resource.publication_year.blank?
+      if work.datacite_resource.creators.count == 0
+        work.errors.add(:base, "Must provide at least one Creator")
+      else
+        work.datacite_resource.creators.each do |creator|
+          if creator.orcid.present? && Orcid.invalid?(creator.orcid)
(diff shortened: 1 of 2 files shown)


app/models/work.rb
```diff
@@ -106,9 +106,17 @@ def default_work(title, user_id, collection_id, datacite_resource, ark)
 
     if work.data_cite.present?
       work.errors.add(:base, "Must provide a title") if work.title.blank?
-      work.errors.add(:base, "Must provide at least one Creator") if work.datacite_resource.creators.count == 0
       work.errors.add(:base, "Must indicate the Publisher") if work.datacite_resource.publisher.blank?
       work.errors.add(:base, "Must indicate the Publication Year") if work.datacite_resource.publication_year.blank?
+      if work.datacite_resource.creators.count == 0
+        work.errors.add(:base, "Must provide at least one Creator")
```
jrgriffiniii, line 112: 👍

app/models/work.rb
```diff
@@ -106,9 +106,17 @@ def default_work(title, user_id, collection_id, datacite_resource, ark)
...
+      else
+        work.datacite_resource.creators.each do |creator|
+          if creator.orcid.present? && Orcid.invalid?(creator.orcid)
+            work.errors.add(:base, "ORCID for creator #{creator.value} is not in format 0000-0000-0000-0000")
```
jrgriffiniii, line 116: 👍

jrgriffiniii (approved):

PR #177, opened 2022-06-23
Prototyping a new layout based on discussions with Hannah

hectorcorrea:
Updated the data entry screen based on the feedback received from Hannah (see [notes](https://docs.google.com/document/d/1PymVnCtxrcp6D2WsLSZEKaqqR1PS1yVK6byRezqgczM/edit)). Major changes include:

* Created a "required metadata" section. 
* Moved the creators to the "required metadata" section.
* Added a (required) description field
* Updated the UI buttons to add new titles to be less intrusive
* Added option to remove titles
* Moved all other fields to the "additional metadata" section.

Screen Shot 2022-06-23 at 3 38 21 PM

More tweaks will be needed, but this is the bulk if them, and it already makes the data entry much simpler for a user: Click new dataset, enter a title, description, and creator and voilá.


app/models/pul_datacite.rb
```diff
@@ -145,7 +150,7 @@ def self.new_person(given_name, family_name, orcid_id = nil, sequence = 0)
       full_name = "#{family_name}, #{given_name}"
       creator = Creator.new(value: full_name, name_type: "Personal", given_name: given_name, family_name: family_name, sequence: sequence)
       if orcid_id.present?
-        creator.name_identifier = NameIdentifier.new_orcid(orcid_id)
+        creator.name_identifier = NameIdentifier.new_orcid(orcid_id.strip)
```
jrgriffiniii, line 153: 👍 I am sorry for missing this within a previous PR

app/views/works/_edit_javascript.html.erb
```diff
@@ -241,5 +254,8 @@
     $("#creators-table-sortable").sortable({
       cancel: "tr:has(th), input, .delete-creator"
     });
+
+    // Give the initial focus to the title.
+    $("#title_main").focus();
```
jrgriffiniii, line 259: 💯

jrgriffiniii:
Coveralls is not showing a decrease in code coverage on https://coveralls.io/builds/50336130

PR #180, opened 2022-06-27
Updating DOI code to utilize prefix environment variables

carolyncole:
Updated the code to use a variable DOI prefix based on the environment. 

Updated error handling in DOI generation and when fetching S3 files associated with a DOI that has no files in S3 yet. Also added special condition to allow the system to work in development without having to have the DataCite credentials.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,9 @@ This is convenient when you need to step through the code as `byebug` does not a
 2. Terminal two: `bin/webpack-dev-server`
 3. Access pdc_describe at [http://localhost:3000/](http://localhost:3000/)
 
+## DataCite integration
+We use DataCite to mint DOIs and in production you must to define the `DATACITE_*` environment values indicated [here](https://github.com/pulibrary/princeton_ansible/blob/main/group_vars/pdc_describe/production.yml) for the system to run. During development if you do not set these values the system will use a hard-coded DOI.
+
 ## Deploying
 pulbot: `pulbot deploy pdc_describe to [staging|production]`
 
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -122,12 +122,16 @@ class Work < ApplicationRecord
   end
 
   def draft_doi
-    self.doi ||= begin
+    self.doi ||= if Rails.env.development? && ENV["DATACITE_USER"].blank?
+                   Rails.logger.info "Using hard-coded test DOI during development."
+                   "10.34770/tbd"
+                 else
                    result = data_cite_connection.autogenerate_doi(prefix: ENV["DATACITE_PREFIX"])
-                   result.either(
-                      ->(response) { response.doi },
-                      ->(response) { raise("Something went wrong", response.status) }
-                    )
+                   if result.success?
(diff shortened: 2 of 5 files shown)


hectorcorrea:
@carolyncole Thank you so much for this. I made a few changes to the PR so that it handles errors properly (the fancy `.either` syntax was not really working) and added special clause to work in development without credentials. I deployed it to staging and it works OK.

app/services/s3_query_service.rb
```diff
@@ -45,10 +45,9 @@ def client
   # * https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html#get_object_attributes-instance_method
   # @return [<S3File>] An Array of S3File objects
   def data_profile
-    return [] if @doi == "10.34770/tbd"
     objects = []
     resp = client.list_objects_v2({ bucket: bucket_name, max_keys: 1000, prefix: prefix })
-    resp.to_h[:contents].each do |object|
+    resp.to_h[:contents]&.each do |object|
```
hectorcorrea, line 50: Prevents this code from crashing if the S3 bucket does not exist.

app/models/work.rb
```diff
@@ -122,15 +122,17 @@ def default_work(title, user_id, collection_id, datacite_resource, ark)
...
   def draft_doi
-    self.doi ||= "10.34770/tbd"
-    # TODO: Set up the doi to have  a variable prefix.  Test and production do not have the same one
-    # self.doi ||= begin
-    #                result = data_cite_connection.autogenerate_doi(prefix: "10.34770")
-    #                result.either(
-    #                   ->(response) { response.doi },
-    #                   ->(response) { raise("Something went wrong", response.status) }
-    #                 )
-    #              end
```
hectorcorrea, line 133: The `either` syntax is what the datacite gem recommends (https://github.com/sul-dlss/datacite-ruby) is more elegant but does not seem to  really catch errors so I ditched it in favor of a good and tried if/else.

app/models/work.rb
```diff
@@ -122,15 +122,17 @@ def default_work(title, user_id, collection_id, datacite_resource, ark)
...
+    self.doi ||= if Rails.env.development? && ENV["DATACITE_USER"].blank?
+                   Rails.logger.info "Using hard-coded test DOI during development."
+                   "10.34770/tbd"
```
jrgriffiniii, line 127: 👍

jrgriffiniii (approved):

jrgriffiniii:
Coveralls does not show any decrease in the test coverage (however, the API seems to be slow)

README.md
```diff
@@ -45,6 +45,9 @@ This is convenient when you need to step through the code as `byebug` does not a
 2. Terminal two: `bin/webpack-dev-server`
 3. Access pdc_describe at [http://localhost:3000/](http://localhost:3000/)
 
+## DataCite integration
```
jrgriffiniii, line 48: 💯

jrgriffiniii (approved):

PR #193, opened 2022-06-28
Started making the Depositor dashboard look more like the wireframes

hectorcorrea:
depositor-dashboard

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -31,6 +31,18 @@ class Work < ApplicationRecord
       Work.where(created_by_user_id: user)
     end
 
+    def unfinished_works(user)
+      Work.where(created_by_user_id: user, state: "AWAITING-APPROVAL")
+    end
+
+    def completed_works(user)
+      Work.where(created_by_user_id: user, state: "APPROVED")
+    end
+
+    def withdrawn_works(user)
+      Work.where(created_by_user_id: user, state: "WITHDRAWN")
+    end
+
     def admin_awaiting_works(user)
       admin_works_by_user_state(user, "AWAITING-APPROVAL")
     end
@@ -119,6 +131,13 @@ class Work < ApplicationRecord
         end
       end
     end
+  end
+
+  # For now all the administrators for the collection are considered
+  # the curators of the dataset. In the future we might assign specific
+  # curator(s) to the dataset.
+  def curators
+    collection.administrators
   end
 
   def draft_doi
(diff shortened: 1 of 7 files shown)


app/views/users/dashboard.html.erb
```diff
@@ -1,47 +1,59 @@
...
+  <%= render partial: "dataset_table", locals: {table_id: "my_withdrawn_datasets", datasets: @my_withdrawn_works} %>
+<% end %>
 
+<script type="text/javascript">
+  $(function() {
+    // https://www.datatables.net/reference/option/
+    var dataset_options = {searching: false, paging: false, info: false};
+    $('#my_unfisnihed_datasets').DataTable(dataset_options);
+    $('#my_completed_datasets').DataTable(dataset_options);
+    $('#my_withdrawn_datasets').DataTable(dataset_options);
```
hectorcorrea, line 57: Add some point we will want to tweak the options for these tables, but for now I am disabling them to keep the UI as clean as possible.

spec/system/add_orcid_spec.rb
```diff
@@ -12,31 +12,12 @@
     it "takes a user to their homepage after login", js: true do
       visit user_path(user)
       expect(page).to have_content "Welcome"
-      expect(page).to have_content "You do not have an ORCiD on file."
-      click_on "Edit"
+      expect(page).to have_content "You do not have an ORCID iD on file."
+      click_on "Add one"
       expect(page).to have_content "Editing User"
```
hectorcorrea, line 17: Notice that now we only nag the user to enter their ORCID if there is none. If they have an ORCID already on file nothing is displayed.

app/models/work.rb
```diff
@@ -31,6 +31,18 @@ def my_works(user)
       Work.where(created_by_user_id: user)
     end
 
+    def unfinished_works(user)
```
carolyncole, line 34: @hectorcorrea I might have made this a bit differently.  You can chain methods, so you could have
```
def by_user(user)
  where(created_by_user_id: user)
end

def by_state(state)
  where(state: state)
end
```
The call could then be
```
@my_unfinished_works = Work.by_user(current_user).by_state("AWAITING-APPROVAL")
```
or are you doing this to hide the states in the work class?

The only reason I am pointing this out is creating the separate methods allow seams for filtering in the future.  Aka an administrator could filter by user, or by state.

carolyncole (commented):
@hector you can tell me that the separate filter methods will not be needed.  This looks good otherwise. It seems like some tests for the work methods would be good to add though.

bess (approved):
I like @carolyncole 's suggestion but I'd also be glad to see this merged as-is and bring that in as a refactor in a separate PR

Issue #110, opened 2022-05-05
Be able to attach files < 100Mb


bess:
- [x] The form has the ability to attach up to 20 files, of < 100Mb each
- [x] Files that are submitted will be stored in the pre-curation S3 bucket for now.
- [x] The files are visible and downloadable from the show screen
- [x] There is a clear difference between pre-curated data and post-curated data 
- [x] Once pre-curated data has been submitted, it is visible in a pre-curated data section on the dataset show page
- [x] AWS S3 Buckets should use environment variables for the `staging` and `production` environments
- [x] Files are stored in S3 using the original (or very similar) filenames that the user submitted.

jrgriffiniii:
Thank you for your patience regarding my progress on this. Regarding my updates for this issue, I have the following:

> The form has the ability to attach up to 20 files, of < 100Mb each

I need to ensure that this validation for up to 20 files is in place, however https://github.com/pulibrary/pdc_describe/pull/220 ensures that multiple files can be updated and replaced

> Files that are submitted will be stored in the pre-curation S3 bucket for now.
[...]
>  AWS S3 Buckets should use environment variables for the staging and production environments

I am addressing this in the next pull request(s)

> The files are visible and downloadable from the show screen
> There is a clear difference between pre-curated data and post-curated data
> Once pre-curated data has been submitted, it is visible in a pre-curated data section on the dataset show page

The files themselves are visible, however, there needs to be clearer disambiguation between pre-curated and curated data

jrgriffiniii:
It was clarified during a meeting held on 07/15/22 that the files stored on disk *must* be human-readable, and stored within a human- navigable directory tree.

jrgriffiniii:
The following error is encountered when testing this within the `staging` environment:

https://app.honeybadger.io/projects/99214/faults/87146315

jrgriffiniii:
I have just tested this successfully from my local environment:

![image.png](https://images.zenhubusercontent.com/5b5b6aed981c3204c9ed06c8/a19dea8d-3c74-4b44-a96f-3e2ff62d3a6e)

jrgriffiniii:
The following error is being logged on the NGINX Plus nodes:

```
[...]

PR #220, opened 2022-07-15
Ensures that clients can replace individual file uploads as depositors or curators

jrgriffiniii:
Advances #110


jrgriffiniii:
This now addressing disambiguating `uncurated` from curated file uploads in the UI. Please let me know should the phrasing for this need to be updated.

bess (approved):

PR #221, opened 2022-07-15
Ensures that the S3 Bucket ActiveStorage service configurations can be set using environment variables

jrgriffiniii:
Advances #110

### DIFF
--- a/spec/models/work_spec.rb
+++ b/spec/models/work_spec.rb
@@ -38,8 +38,6 @@ RSpec.describe Work, type: :model, mock_ezid_api: true do
   let(:identifier) { @identifier }
   let(:attachment_url) { "https://example-bucket.s3.amazonaws.com/#{work.doi}/" }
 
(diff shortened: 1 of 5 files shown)


bess (changes requested):
@jrgriffiniii What is the relationship between `config/s3.yml` and `config/storage.yml`? The first one is setting the name of the post-curation S3 bucket, and the second one is also about S3 storage, but for the pre-curation bucket? I think? I'd like to get really clear on the configuration here. This code looks fine to me, but I'd like us to untangle how we want to configure storage in general before we get into a muddle.

spec/models/work_spec.rb
```diff
@@ -36,6 +36,7 @@
   # Please see spec/support/ezid_specs.rb
   let(:ezid) { @ezid }
   let(:identifier) { @identifier }
+  let(:attachment_url) { "https://example-bucket.s3.amazonaws.com/#{work.doi}/" }
```
jrgriffiniii, line 39: I am sorry, this seems to have been a rebase error, I am fixing this now.

bess (approved):

PR #241, opened 2022-07-26
Include works where the user is mentioned 

### LINKED ISSUE #223: Notifications for datasets not owned by the user
It is possible to tag a user on a comment in a dataset they don't have access to. This results in the user getting a message on their dashboard that they have an unread message (which is good) but no way of knowing which dataset the notification is for (not good).

This would be a common problem more regular users since by default they only have access to the datasets that they deposit.

We should display in the dashboard a list of datasets where the user has been mentioned regardless of access.

hectorcorrea:
Update the dashboard to include works where the user is mentioned (even if they don't own the dataset). 

This is to handle cases where a user tags a user, say `@userA`, on a dataset that they typically does not have access to (e.g. it's the PPPL collection). With this changes the dataset will still show under `@userA's` dashboard and with the proper badge indicating that there are unread notifications.

Closes #223 

Below is an example on how this looks in practice. Notice that the dashboard for Bess has a dataset with a badge with one pending notification because Bess was mentioned in a comment for this dataset.

dashboard

mention

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -12,6 +12,7 @@ AllCops:
     - "vendor/**/*"
     - "node_modules/**/*"
     - "bin/*"
+    - "storage/**/*"
 
 Layout/IndentationConsistency:
   EnforcedStyle: indented_internal_methods
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -54,7 +54,28 @@ class Work < ApplicationRecord
           works += Work.where(collection_id: collection.id, state: state)
         end
       end
+
+      # Any other works where the user is mentioned
+      works_mentioned_by_user_state(user, state).each do |work_id|
+        already_included = !works.find { |work| work[:id] == work_id }.nil?
+        works << Work.find(work_id) unless already_included
+      end
+
       works.sort_by(&:updated_at).reverse
+    end
+
+    # Returns an array of work ids where a particular user has been mentioned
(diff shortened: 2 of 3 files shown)


.rubocop.yml
```diff
@@ -12,6 +12,7 @@ AllCops:
     - "vendor/**/*"
     - "node_modules/**/*"
     - "bin/*"
+    - "storage/**/*"
```
hectorcorrea, line 15: We don't want to test our datasets :)

app/models/work.rb
```diff
@@ -46,17 +46,38 @@ def works_by_user_state(user, state)
...
+        SELECT DISTINCT works.id
+        FROM works
+        INNER JOIN work_activities ON works.id = work_activities.work_id
+        INNER JOIN work_activity_notifications ON work_activities.id = work_activity_notifications.work_activity_id
+        WHERE work_activity_notifications.user_id = #{user.id} AND works.state = '#{state}'
+      END_SQL
+      rows = ActiveRecord::Base.connection.execute(sql)
+      rows.map { |row| row["id"] }
+    end
+
```
hectorcorrea, line 80: See our ERD if this SQL is puzzling: https://docs.google.com/drawings/d/1q2sfj8rrcNVgqQPK5uT_t79A9SYqncinh3HbnCSGMyQ/edit

bess (approved):

PR #244, opened 2022-07-28
Integrating the configurations between the S3QueryService and ActiveStorage

### LINKED ISSUE #243: Evaluate, compare, and reconcile the design patterns for the pre-curation and post-curation AWS S3 storage support
As noted by @bess, the solution which I proposed in https://github.com/pulibrary/pdc_describe/pull/221 deviates significantly from the patterns found within https://github.com/pulibrary/pdc_describe/blob/main/app/services/s3_query_service.rb. These must be reconciled for a common set of interfaces offered by the `Work` models.

jrgriffiniii:
Resolves #243

### DIFF
--- a/app/services/s3_query_service.rb
+++ b/app/services/s3_query_service.rb
@@ -4,18 +4,50 @@ require "aws-sdk-s3"
 
 # A service to query an S3 bucket for information about a given data set
 class S3QueryService
+  def self.configuration
+    Rails.configuration.s3
+  end
+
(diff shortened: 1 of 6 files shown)


bess (approved):
This looks great! There is a test failing, so I won't merge it at the moment, but this config is exactly the kind of thing I was hoping for. Thank you. 🚀

bess (approved):
🚀

PR #246, opened 2022-07-29
Submit new dataset step 0

### LINKED ISSUE #199: New submission - step 0 changes
Update the "Step 0" in the new submission (see below) process to ask the user for a title and at least one creator before clicking "Create New."

step-0

This way by the time they get to the next step (where we show them the DOI) we know for sure that they have entered a title and a creator for the submission.

step-1

hectorcorrea:
Updated the Step 0 in the wizard to submit a new dataset to allow/force the user to indicate the *Title* and at least one *Creator* as requested on #199. This new workflow has the advantage that we will have at least two basic bits of information before we mint a DOI.

The following screenshots show the general workflow.

step0

step1

Closes #199

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -13,11 +13,14 @@ class WorksController < ApplicationController
   end
 
   # Renders the "step 0" information page before creating a new dataset
-  def new_submission; end
+  def new; end
 
-  def new
+  # Creates the new dataset
+  def new_submission
+    byebug
     default_collection_id = current_user.default_collection.id
-    work = Work.create_dataset("New Dataset", current_user.id, default_collection_id)
+    datacite_resource = datacite_resource_from_form
+    work = Work.create_dataset(datacite_resource.main_title, current_user.id, default_collection_id, datacite_resource)
     redirect_to edit_work_path(work, wizard: true)
   end
 
@@ -67,7 +70,7 @@ class WorksController < ApplicationController
     updates = {
       title: title_param,
       collection_id: collection_id_param,
-      data_cite: datacite_resource_from_form,
+      data_cite: datacite_resource_from_form.to_json,
       deposit_uploads: updated_deposit_uploads
     }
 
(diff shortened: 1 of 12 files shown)


app/controllers/works_controller.rb
```diff
@@ -13,11 +13,15 @@ def index
...
+  end
+
+  # Creates the new dataset
+  def new_submission
     default_collection_id = current_user.default_collection.id
-    work = Work.create_dataset("New Dataset", current_user.id, default_collection_id)
+    datacite_resource = datacite_resource_from_form
+    work = Work.create_dataset(datacite_resource.main_title, current_user.id, default_collection_id, datacite_resource)
     redirect_to edit_work_path(work, wizard: true)
   end
```
hectorcorrea, line 26: Swapped the meaning of the `new` and `new_submission` methods and their respectively routes in `config.rb` and I think this makes more sense and it's more Rails-y. `New` is now a GET operation and `new_submission` is the POST.

app/controllers/works_controller.rb
```diff
@@ -213,8 +217,8 @@ def datacite_resource_from_form
       resource = PULDatacite::Resource.new
 
       resource.description = params["description"]
-      resource.publisher = params["publisher"]
-      resource.publication_year = params["publication_year"]
+      resource.publisher = params["publisher"] if params["publisher"].present?
+      resource.publication_year = params["publication_year"] if params["publication_year"].present?
```
hectorcorrea, line 221: Don't overwrite the defaults that were set in the model.

app/views/works/_form_hidden_fields.html.erb
```diff
@@ -4,9 +4,9 @@
   initially and how many are added on the fly. This helps to process the data when the form
   is submitted.
   -->
-  <input type="text" id="existing_title_count" name="existing_title_count" value="<%= @work.datacite_resource.titles.count %>" class="hidden" />
+  <input type="text" id="existing_title_count" name="existing_title_count" value="<%= @work&.datacite_resource&.titles&.count || 0 %>" class="hidden" />
```
hectorcorrea, line 7: The extra `&.` validations are needed because this view is now rendered from another view that does not declare an `@work`.

jrgriffiniii (approved):

PR #251, opened 2022-08-02
Adding sending of XML to datacite

carolyncole:
Updated the gem to allow for xml format, which is why we are pointing at main.

Screen Shot 2022-08-02 at 11 14 15 AM

Draft PR on gem, which allowed for the above to be published via xml: https://github.com/sul-dlss/datacite-ruby/pull/18

Tested by creating a new item via `rspec spec/models/doi_spec.rb:68` and then running `rspec spec/models/doi_spec.rb:142` with the new id.  (Connected to the real server)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,7 +47,7 @@ gem "friendly_id", "~> 5.4.0"
 
 gem "faraday"
 
-gem "datacite"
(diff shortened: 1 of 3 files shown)


bess (approved):
🚀

PR #258, opened 2022-08-05
ADR for globus and s3 permissions

carolyncole:


bess (approved):

bess:
We discussed with with RDOS today, so I am merging this.

Issue #227, opened 2022-07-23
Skeleton DataCite record must be valid


bess:
The current skeleton DataCite record we produce should be valid. This will allow us to take a "lean on green" posture as we add more fields. 

**ACCEPTANCE CRITERIA**
- [x] Install the datacite-mapping gem: https://www.ruby-toolbox.com/projects/datacite-mapping
- [x] Refactor our DataCite record construction to use the methods in that gem
- [x] Produce a datacite record that is structurally valid even if not all fields are populated

PR #260, opened 2022-08-08
Build datacite in the ValidDatacite module

bess:
We'll then refactor bit by bit to use it,and then we can change the name.

Work toward #227 

@carolyncole Does this give you enough to unblock you on passing datacite during DOI creation? This is going to be a big refactor by the time it's all done and I'm trying to phase it so I'm not sitting on a feature branch for weeks.

### DIFF
--- a/app/models/valid_datacite.rb
+++ b/app/models/valid_datacite.rb
@@ -1,0 +1,137 @@
+# frozen_string_literal: true
+module ValidDatacite
+  # Placeholder while we refactor everything to re-use this class. Then we'll re-name it.
+  # Represents a PUL Datacite resource.
+  # This class should be instantiated via the `.new_from_json` method.
+  # It takes PDC Describe form submission data and turns it into valid Datacite.
+  # https://support.datacite.org/docs/datacite-metadata-schema-v44-properties-overview
+  # Note that the actual datacite mapping and XML serialization is done by the datacite-mapping gem.
+  class Resource
+    attr_accessor :mapping, :metadata_from_form, :identifier_type, :creators, :titles, :publisher, :publication_year, :resource_type, :description
+
+    ##
+    # Create a valid skeleton record. Useful for early in the workflow when we don't have much data yet,
+    # and for testing.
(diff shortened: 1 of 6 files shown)


app/models/valid_datacite.rb
```diff
@@ -0,0 +1,141 @@
...
+    def to_xml
+      datacite_mapping.write_xml
+    end
+
+    ##
+    # For each creator, make a Datacite::Mapping::Creator object
+    # TODO: This class has no field for author order. We need a place to record that.
+    def datacite_creators
+      creator_array = []
+      @metadata_from_form["creators"].each do |creator|
```
carolyncole, line 92: Could this just be a map?  Then you would not need the separate array and it would be clearer that it is a one to one relationship.
```suggestion
      @metadata_from_form["creators"].map do |creator|
```

app/models/valid_datacite.rb
```diff
@@ -0,0 +1,141 @@
...
+      end
+      creator_array
+    end
+
+    ##
+    # For each title, make a Datacite::Mapping::Title object
+    # TODO: Add qualifiers for different kinds of titles
+    def datacite_titles
+      title_array = []
+      @metadata_from_form["titles"].each do |title|
```
carolyncole, line 108: I would prefer map here to
```suggestion
      @metadata_from_form["titles"].map do |title|
```

db/migrate/20220808200759_add_datacite_xml_to_work.rb
```diff
@@ -0,0 +1,5 @@
+class AddDataciteXmlToWork < ActiveRecord::Migration[6.1]
+  def change
+    add_column :works, :datacite_xml, :text
```
carolyncole, line 3: 👍 Assuming we will delete the other column after the full refactor

carolyncole (changes requested):
@bess looking like a greta first step.  Just a couple small tweaks for readability.

db/migrate/20220808200759_add_datacite_xml_to_work.rb
bess: Yes, that was my plan.

carolyncole (approved):

PR #262, opened 2022-08-09
Upgrade rails for security fix

bess:


carolyncole (approved):
Tested on staging.  Seemed to work.

Issue #261, opened 2022-08-09
Move to storing XML for the datastore


bess:
Now that we are producing valid Datacite XML records, these should be our storage medium for the data. 

**ACCEPTANCE CRITERIA**
- [ ] Refactor `Work` object to store XML in the database instead of JSON
- [ ] Ensure updates to the record persist to the stored XML
- [ ] Re-name `ValidDatacite` to `PULDatacite` after everything has been refactored to use the new datacite structure
- [ ] Remove the stored JSON field from the database once we aren't using it any longer

bess:
Closing this issue in favor of new tickets that will come out of a refactoring plan.

PR #265, opened 2022-08-09
System spec to test form submission

bess:
Work toward #261 

Compile assets when running tests in browser.

```
RUN_IN_BROWSER=true be rspec spec/system/form_submission_spec.rb
```

Before: 
Screen Shot 2022-08-09 at 3 17 44 PM

After: 
Screen Shot 2022-08-09 at 3 18 52 PM


carolyncole (approved):
Thank you!

Issue #207, opened 2022-07-06
Back the system with a state machine


carolyncole:
I believe that adding a state machine gem like [AASM](https://github.com/aasm/aasm) will give us greater flexibility as the system gets more complex.

We utilized this gem in Approvals  to allow for complex state transitions with notifications (https://github.com/pulibrary/approvals/blob/16fec05410196bc3be689784c4a9e919a110102a/app/models/concerns/aasm_config.rb)

Of note to one of the current tickets is the logging at every state change: https://github.com/pulibrary/approvals/blob/16fec05410196bc3be689784c4a9e919a110102a/app/models/concerns/aasm_config.rb#L12

[Here](https://github.com/pulibrary/pdc_describe/blob/main/docs/work_states.md) is a document that describes all the desired work states.

hectorcorrea:
As you work on this ticket, keep this other ticket in mind: https://github.com/pulibrary/pdc_describe/issues/237

carolyncole:
Need to handle state transition errors:
![Screen Shot 2022-08-16 at 7 39 51 AM](https://user-images.githubusercontent.com/1599081/184870932-0eeb9b3e-aea5-4845-99a3-a64501b59308.png)

PR #281, opened 2022-08-12
Adding in the AASM gem and making the work have states

carolyncole:
refs #207

I do not think this completes the transition to states, as there is still a lot of information in before hooks around save, but it is a first step and everything is still working according to the tests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@
 source "https://rubygems.org"
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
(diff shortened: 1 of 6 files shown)


carolyncole:
I believe this document reflects the wished state of the system: #282.  It is fairly close to the current state of the system after this PR, but I think for the moment everything starts in Awaiting Approval and the Draft state is really not utilized.

Twade968 (approved):
Reviewed by Bess and Tyler together.

PR #285, opened 2022-08-12
Fixes bug in Edit form

### LINKED ISSUE #284: Edit error due to incorrect helper name
The Edit form throws an error when editing a work that has files attached to it. The issue is that the `_deposit_uploads_form.html.erb` is referencing a non-existing helper. It references `deposit_uploads` instead of `deposit_uploads_file_name. 

We fixed this the new submission form in PR https://github.com/pulibrary/pdc_describe/pull/283 but evidently we missed it for the Edit form.

hectorcorrea:
Fixes a bug in the form rendering when there are indeed files attached to a work. This similar to the fix that we did in PR https://github.com/pulibrary/pdc_describe/pull/283

Fixes #284 

I added a separate issue to update the tests (https://github.com/pulibrary/pdc_describe/issues/286) since that will take a bit more work and I did not want to keep production broken.

### DIFF
--- a/app/views/works/_deposit_uploads_form.html.erb
+++ b/app/views/works/_deposit_uploads_form.html.erb
@@ -21,7 +21,7 @@
                   <td>
                     <span>
                       <i class="bi bi-file-arrow-down"></i>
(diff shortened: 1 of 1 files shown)


hectorcorrea:
Here is an example (in staging) of a Work that already has files and that now it can be edited with this PR: https://pdc-describe-staging.princeton.edu/describe/works/66

bess (approved):

PR #301, opened 2022-08-16
[WIP] Restructuring the S3QueryService Class to query AWS Bucket resources using ActiveStorage

### LINKED ISSUE #288: s3_query_service.rb and work.deposit_uploads
Currently the code in `s3_query_service.rb` fetches data from S3 for a given work (by its DOI), however, for works where the user uploaded the files from PDC Describe (instead of directly to S3) the metadata for theses files lives in Rails ActiveStorage tables (accessible via `work.deposit_uploads`) and the information could be duplicated.

This double source of data should be handled in the code "somehow" as it currently causes issues since sometimes we get data via `work.deposit_uploads` (with uses ActiveStorage tables) but other times we get the data via `service = S3QueryService.new(@work.doi)` (which goes straight to S3)

jrgriffiniii:
refs #288 

This PR allows you to see the files added to S3 outside of the system upload. Those files can not be downloaded yet, which is why this PR only refs #288.


app/models/work.rb
```diff
@@ -188,9 +207,15 @@ def valid_to_submit
     errors.clear
     validate_ark
     validate_metadata
-    if deposit_uploads.length > 20
-      errors.add(:base, "Only 20 files may be uploaded by a user to a given Work. #{deposit_uploads.length} files were uploaded for the Work: #{ark}")
+
+    if pre_curation_uploads.length > 20
```
carolyncole, line 211: Do we need to check the total of post and pre curated so we do not allow them to add more than 20 and then can not transfer them when they get approved?

app/models/work.rb
```diff
@@ -149,7 +151,24 @@ def default_work(user_id, collection_id, resource, ark)
 
     work.metadata = work.resource.to_json
 
-    new_attachments = work.deposit_uploads.reject(&:persisted?)
+    new_attachments = work.pre_curation_uploads.reject(&:persisted?)
+    new_attachments.each do |attachment|
```
carolyncole, line 155: Could this block be extracted to a method, so that it can be utilized for both post and pre curated attachments?

app/services/s3_query_service.rb
```diff
@@ -95,14 +60,17 @@ def client
   #   ok is false if there is an error connecting to S3. Otherwise true.
   def data_profile
     objects = []
-    resp = client.list_objects_v2({ bucket: bucket_name, max_keys: 1000, prefix: prefix })
-    resp.to_h[:contents]&.each do |object|
-      s3_file = S3File.new(filename: object[:key], last_modified: object[:last_modified], size: object[:size])
+    return objects if model.nil?
```
carolyncole, line 63: I thought this code was to get the files that were put into S3, but are not represented yet in the database. Now it looks like this is not the case.  Is that a next step, or do I have the wrong idea about what this code does?

carolyncole (changes requested):
@jrgriffiniii I really like the separation of pre-curation and post-curation files.  I added a couple comments and a question.

carolyncole (approved):

carolyncole (changes requested):
@jrgriffiniii The code look great!
I'm unsure what is going on, but when I deploy the branch to staging this branch causes an error.  Sorry!  I would love to get this in.  It is possible the configurations need to change?

app/services/s3_query_service.rb
```diff
@@ -85,6 +85,40 @@ def client
     @client ||= Aws::S3::Client.new(region: region, credentials: credentials)
   end
 
+  def model
+    @model ||= Work.find_by(doi: @doi)
+  end
+
+  def model_s3_files
```
hectorcorrea, line 92: @jrgriffiniii -- If I understand this correctly `model_s3_files` includes files the user added via PDC Describe upload feature and that PDC Described pushed to S3, whereas `client_s3_files` includes files all files in the S3 folder (regardless of how they got there, e.g. uploaded via PDC Describe or manually uploaded to S3). 

`s3_files` returns the combination of these two methods and handles duplicates gracefully. Is that understanding correct? If so, could we add a comment to these methods to clarify this?

carolyncole:
@jrgriffiniii  I believe changing https://github.com/pulibrary/pdc_describe/blob/main/config/environments/staging.rb#L44 to `:amazon_pre_curation` and production also would solve the deployment issue.  It works on my local system

carolyncole (approved):

hectorcorrea (approved):

PR #313, opened 2022-08-18
Remove code that handles Dublin Core data

### LINKED ISSUE #280: Remove dublin core from the work
We have settled on one internal format (Work.metadata) and the use of Serializers to convert that format to other schemas.  We should not store dublin_core in the database.

- [ ] Remove references dublin_core from the Work
- [ ] Remove [app/services/dspace_import_service.rb] (https://github.com/pulibrary/pdc_describe/blob/ad0574ea3d90d4090beda1d1eb4c3188fb6c8e42/app/services/dspace_import_service.rb) This is a manual process because Dspace metadata is very dirty.
- [ ] Remove any dublin_core tickets

hectorcorrea:
Closes #280


hectorcorrea:
I think the decrease in test coverage that Coveralls is flagging is because I deleted a bunch of tests since the code/classes that they were testing is now gone.

bess (approved):

PR #316, opened 2022-08-19
DOI and ARK

### LINKED ISSUE #299: We do not have the ability to store multiple identifiers in our internal JSON representation
DOI and Ark are two identifiers that should be stored in our internal representation instead of at the class level.

Unfortunately storing them in out internal representation is only possible if the data has ARK OR a DOI, but not both.  We need to update our internal representation to allow both and move the work level information down into the metadata.

This code limits us to one: https://github.com/pulibrary/pdc_describe/blob/main/app/models/pul_datacite.rb#L9-L10

Datacite only allows for one identifier.  We will pass the DOI to Datacite (https://support.datacite.org/docs/datacite-metadata-schema-v44-mandatory-properties#1-identifier)

We will always mint a DOI for new works.  The ARK will only be utilized for records with existing ARKs

This will be done when:
- [x] Both ARK and DOI should be able to be stored in the PULDataCite::Resource
- [x] These can be named values, which may or may not have an associated type
- [x] extension for other ids has been kept in mind

hectorcorrea:
Moved DOI from a field in Work table `Work.field` to a property of the resource `Work.resource.doi`. 

Moved ARK to work the same way, i.e. `Work.resource.ark`. Stopped minting ARKs for new Works automatically and instead allow the user to enter an ARK for a given work, this will be useful for data migrated from DataSpace which has ARKs.

Fixes #299

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -19,7 +19,7 @@ class Work < ApplicationRecord
       transitions from: :none, to: :draft, guard: :valid_to_draft
     end
 
-    event :ready_for_review do
+    event :complete_submission do
       transitions from: :draft, to: :awaiting_approval, guard: :valid_to_submit
     end
 
@@ -291,6 +291,10 @@ class Work < ApplicationRecord
         Ark.update(ark, url)
       end
     end
+  end 
+  
+  def current_transition
+    aasm.current_event.to_s.humanize.delete("!")
   end
 
   private
@@ -315,7 +319,6 @@ class Work < ApplicationRecord
     end
 
     def track_state_change(user, state = aasm.to_state)
-      save!
       uw = UserWork.new(user_id: user.id, work_id: id, state: state)
       uw.save!
       WorkActivity.add_system_activity(id, "marked as #{state}", user.id)
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -291,8 +291,8 @@ class Work < ApplicationRecord
         Ark.update(ark, url)
       end
     end
-  end 
-  
+  end
+
   def current_transition
     aasm.current_event.to_s.humanize.delete("!")
   end
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -63,18 +63,8 @@ class Work < ApplicationRecord
       work
     end
 
-    # Convenience method to create Datasets with the DataCite profile
-    def create_dataset(user_id, collection_id, resource, ark = nil)
-      work = default_work(user_id, collection_id, resource, ark)
-      work.draft_doi
-
-      # We skip the validation since we don't have all the required fields yet
-      work.save!(validate: false)
-      work
-    end
-
     def unfinished_works(user)
-      works_by_user_state(user, "awaiting_approval")
+      works_by_user_state(user, ["none", "draft", "awaiting_approval"])
     end
 
     def completed_works(user)
@@ -101,9 +91,9 @@ class Work < ApplicationRecord
         end
 
         # Any other works where the user is mentioned
-        works_mentioned_by_user_state(user, state).each do |work_id|
-          already_included = !works.find { |work| work[:id] == work_id }.nil?
-          works << Work.find(work_id) unless already_included
+        works_mentioned_by_user_state(user, state).each do |work|
+          already_included = !works.find { |existing_work| existing_work[:id] == work.id }.nil?
+          works << work unless already_included
         end
 
         works.sort_by(&:updated_at).reverse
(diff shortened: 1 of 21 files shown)


app/models/work.rb
```diff
@@ -113,10 +113,6 @@ def works_mentioned_by_user_state(user, state)
 
   before_save do |work|
     # Ensure that the metadata JSON is persisted properly
-    if work.profile == "DATACITE" && work.ark.blank?
-      work.ark = Ark.mint
-    end
-
```
hectorcorrea, line 119: We don't mint ARKs for new works anymore...

app/views/works/_additional_form.html.erb
```diff
@@ -19,10 +19,9 @@
...
-  <!-- ARK field  -->
-  <% if @work.ark.present? %>
-    <div class="field">
-      ARK: <%= @work.ark %>
-    </div>
-  <% end %>
+  <!-- ARK field (mostly for legacy DSpace works) -->
+  <div class="field">
+    ARK: <br>
+    <input type="text" id="ark" name="ark" class="input-text-long" value="<%= @work.resource.ark %>" />
```
hectorcorrea, line 25: Notice that the ARK a field is now a user entered field (in the Additional Fields tab). This will allow us to indicate the ARK value for records that we migrate from DataSpace.

app/models/work.rb
```diff
@@ -161,6 +147,9 @@ def valid_to_draft
   def valid_to_submit
     valid_to_draft
     validate_metadata
+    if state == "approved"
```
carolyncole, line 150: I believe that update_ark_information and publish_doi should likely be in the same after: method.

The way this is written it will happen during validation even if the item is not valid.  [Maybe](https://github.com/pulibrary/pdc_describe/blob/8ea4e5d67be70c2bc4b966d934821a447b147eb4/app/models/work.rb#L31) becomes 
```
transitions from: :awaiting_approval, to: :approved, guard: :valid_to_submit, after: publish
...
def publish
   publish_doi
   update_ark_information
end
```

app/views/works/_required_form.html.erb
```diff
@@ -2,14 +2,14 @@
...
-    <input class="text-not-editable" type="text" id="doi" name="doi" value="<%= @work.doi %>" readonly/>
+    <input class="text-not-editable" type="text" id="doi" name="doi" value="<%= @work.resource.doi %>" readonly/>
   </div>
   <div>&nbsp;</div>
 <% else %>
   <div class="field">
     <p class="doi-text">By initiating this new submission, we have reserved a draft DOI for your use.<br/>
       This unique identifier will become fully minted and searchable at the completion of this process.</p>
-    <p>DOI: <%= @work.doi %></p>
+    <p>DOI: <%= @work.resource.doi %></p>
```
carolyncole, line 12: I wonder with all the places that we are reaching into the resource for doi if we should have a helper method like title, so we can still call work.doi.

lib/tasks/works.rake
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+namespace :works do
+  desc "Transfers DOI value from Work.doi to Work.resource.doi"
+  task transfer_doi_fix: :environment do
```
carolyncole, line 4: 🎉 Great to have a migration!

spec/models/work_spec.rb
```diff
@@ -162,29 +189,11 @@
...
-    before { allow(Ark).to receive(:update) }
-    let(:ezid) { "ark:/99999/dsp01qb98mj541" }
-
-    around do |example|
-      Rails.configuration.update_ark_url = true
-      example.run
-      Rails.configuration.update_ark_url = false
-    end
-
-    it "updates the ARK metadata" do
```
carolyncole, line 175: I would expect this test would still be valid.  You would just have to approve the item in the test.

carolyncole (changes requested):
Looks great!  I asked for a couple updates.

spec/models/work_spec.rb
hectorcorrea: It is still valid. The way Git aligns the difference looks like I deleted the test but in fact I just moved it to a different section (see lines 123-129) and added another test related to ARK (lines 131-137)

app/models/work.rb
hectorcorrea: Done. Added the `after` trigger and move the code to a new `publish` method

app/views/works/_required_form.html.erb
hectorcorrea: I'll ticket this as a separate issue. For now I want to have a way to still fetch the `work.doi` from the ActiveRecord so that I can migrate it to `work.resource.doi`

hectorcorrea: https://github.com/pulibrary/pdc_describe/issues/319

carolyncole (approved):

PR #320, opened 2022-08-22
Publishing DOI after approval

### LINKED ISSUE #196: When a dataset is approved the DOI is marked as final/discoverable in Datacite
We currently have the process to generate a draft DOI. We should also figure out how to mark the draft DOI as final DOI and submit the proper metadata for it to DataCite.

This will be needed once we implement the Approve button for a dataset. Right now the button just flags the dataset as APPROVED but eventually it should update DataCite as well.

If an error occurs during the DataCite publishing we will add a notification tagging the curator telling them that there was an issue publishing to DataCite (perhaps with the details) (as @hectorcorrea suggested in slack)

carolyncole:
Sends the update to DataCite
Handles any errors by adding WorkActivities

fixes #196

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -279,27 +279,27 @@ class Work < ApplicationRecord
     end
   end
 
-  def publish
-    # call to publish_doi will go here
-    update_ark_information
-  end
-
-  # Update EZID (our provider of ARKs) with the new information for this work.
-  def update_ark_information
-    # We only want to update the ark url under certain conditions.
-    # Set this value in config/update_ark_url.yml
-    if Rails.configuration.update_ark_url
-      if ark.present?
-        Ark.update(ark, url)
-      end
-    end
-  end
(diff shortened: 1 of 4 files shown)


app/models/work.rb
```diff
@@ -358,5 +358,34 @@ def validate_creators
...
+      else
+        result = data_cite_connection.update(id: doi, attributes: doi_attributes)
+        if result.failure?
+          message = "@#{curator_or_current_uid(user)} Error publishing DOI. #{result.failure.status} / #{result.failure.reason_phrase}"
+          WorkActivity.add_system_activity(id, message, user.id, activity_type: "DATACITE_ERROR")
+        end
+      end
+    end
+
+    def curator_or_current_uid(user)
```
hectorcorrea, line 374: great idea!

hectorcorrea (approved):

PR #322, opened 2022-08-22
Makes sure the DOI minted is saved

### LINKED ISSUE #321: DOI is not being saved
When creating a new work we are minting a new DOI but the value is not being saved

The controller makes the correct call to `draft!`:

```
  def new_submission
    default_collection_id = current_user.default_collection.id
    resource = resource_from_form
    work = Work.new(created_by_user_id: current_user.id, collection_id: default_collection_id, metadata: resource.to_json)
    work.draft!(current_user)
    redirect_to edit_work_path(work, wizard: true)
  end
```

and the draft transition (in models/work.rb) makes the call to draft the DOI but this value is not saved.

```
    event :draft, after: :draft_doi do
      transitions from: :none, to: :draft, guard: :valid_to_draft
    end
```

The `!` in the call to `draft!` forces a save but we are minting the value in the `after` clause and that is not forcing a save.

hectorcorrea:
Fixes #321

### DIFF
--- a/spec/models/work_spec.rb
+++ b/spec/models/work_spec.rb
@@ -330,6 +330,7 @@ RSpec.describe Work, type: :model, mock_ezid_api: true do
     let(:draft_work) do
       work = Work.new(collection: collection, metadata: FactoryBot.build(:resource).to_json)
       work.draft!(user)
+      work = Work.find(work.id)
       work
     end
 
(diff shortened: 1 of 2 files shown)


spec/models/work_spec.rb
```diff
@@ -330,16 +330,18 @@
     let(:draft_work) do
       work = Work.new(collection: collection, metadata: FactoryBot.build(:resource).to_json)
       work.draft!(user)
+      work = Work.find(work.id)
```
hectorcorrea, line 333: Forces to make sure we use the value in the DB. I thought work.reload will do this but that didn't work.

carolyncole (approved):

PR #326, opened 2022-08-23
Drop the work.doi and work.ark database columns

hectorcorrea:
Drop the work.doi and work.ark database columns since now we store the data as part of the JSON `resource` field. Aliased their getters in the work model (but not the assignments) and updated the tests

### DIFF
--- a/spec/models/work_spec.rb
+++ b/spec/models/work_spec.rb
@@ -6,8 +6,7 @@ RSpec.describe Work, type: :model, mock_ezid_api: true do
   let(:collection) { Collection.research_data }
   let(:user_other) { FactoryBot.create :user }
   let(:superadmin_user) { User.from_cas(OmniAuth::AuthHash.new(provider: "cas", uid: "fake1", extra: { mail: "fake@princeton.edu" })) }
-  let(:doi) { "https://doi.org/10.34770/0q6b-cj27" }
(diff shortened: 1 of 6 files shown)


spec/models/work_spec.rb
```diff
@@ -6,8 +6,7 @@
   let(:collection) { Collection.research_data }
   let(:user_other) { FactoryBot.create :user }
   let(:superadmin_user) { User.from_cas(OmniAuth::AuthHash.new(provider: "cas", uid: "fake1", extra: { mail: "fake@princeton.edu" })) }
-  let(:doi) { "https://doi.org/10.34770/0q6b-cj27" }
-  let(:work) { FactoryBot.create(:draft_work, doi: doi) }
+  let(:work) { FactoryBot.create(:draft_work) }
```
carolyncole, line 9: You can add a transient attribute like https://github.com/pulibrary/pdc_describe/blob/main/spec/factories/resource.rb#L5-L7 To allow for setting the doi on the factory.  Not needed since the tests still pass...

carolyncole (approved):
Added a comment on what could be done to continue setting the DOI on the factory, but it is not required.

spec/models/work_spec.rb
hectorcorrea: Good to know, thanks!

PR #338, opened 2022-08-26
Moved Datacite serialization to its own module, refactored to use the…

### LINKED ISSUE #276: Move ValidDatacite class to a module PDCSerialization::Datacite
- [x] Create a PDCSerialization directory
- [x] Move ValidDatacite to Datacite under that module
- [x] Change the interface to accept a work to serialize (leave skeleton)
- [x] Add class documentation for how we will utilize this class
        ```
        PDCSerialization::Datacite.from_work(work).to_xml
        ```

hectorcorrea:
Moved Datacite serialization to its own module, refactored to use the Resource class (rather than accessing the hash directly).

Closes #276


app/models/pdc_serialization/datacite.rb
```diff
@@ -0,0 +1,137 @@
...
+        resource_type: datacite_resource_type(resource_type)
+      )
+      mapping.write_xml
+    end
+
+    ##
+    # Creates a PDCSerialization::Datacite object from a Work
+    def self.new_from_work(work)
+      new_from_work_resource(work.resource)
+    end
```
hectorcorrea, line 43: I named this `new_from_work` instead of `from_work` (as requested on the original issue) because we use `new_from_x` in `PDCMetadata::Resource`

app/models/pdc_serialization/datacite.rb
```diff
@@ -0,0 +1,155 @@
...
+        creators: creators_from_work_resource(resource.creators),
+        titles: titles_from_work_resource(resource.titles),
+        publisher: ::Datacite::Mapping::Publisher.new(value: resource.publisher),
+        publication_year: resource.publication_year,
+        resource_type: datacite_resource_type(resource.resource_type)
+      )
+      Datacite.new(mapping)
+    end
+
+    def self.creators_from_work_resource(creators)
```
carolyncole, line 77: @hectorcorrea are these part of the public API? creators_from_work, name_identifier_from_identifier, titles_from_work_resource?  If not they should be private.  Utilize the `class << self` to allow for private class methods (https://github.com/pulibrary/pdc_describe/blob/71b93988be3f82e003c80493bd8a1b0dda126f7b/app/models/pdc_metadata/creator.rb#L11)

carolyncole (changes requested):
@hectorcorrea looks great.  I think you have some private methods public.  If they are public then they should be tested.

app/models/pdc_serialization/datacite.rb
hectorcorrea: Good idea. I've made the change.

app/models/pdc_serialization/datacite.rb
```diff
@@ -0,0 +1,160 @@
...
+        titles: titles_from_work_resource(resource.titles),
+        publisher: ::Datacite::Mapping::Publisher.new(value: resource.publisher),
+        publication_year: resource.publication_year,
+        resource_type: datacite_resource_type(resource.resource_type)
+      )
+      Datacite.new(mapping)
+    end
+
+    class << self
+      def creators_from_work_resource(creators)
```
carolyncole, line 80: @hectorcorrea This will need the private keyword to make them private.

hectorcorrea: Are you sure? If I run this on the terminal the method is not available:

```
[1] pry(main)> datacite = PDCSerialization::Datacite.new({})
=> #<PDCSerialization::Datacite:0x00007fa16fcb35c8 @mapping={}>

[2] pry(main)> datacite.datacite_resource_type("x")
NoMethodError: undefined method `datacite_resource_type' for #<PDCSerialization::Datacite:0x00007fa16fcb35c8 @mapping={}>
from (pry):2:in `<main>'


```

hectorcorrea: nevermind...my bad

hectorcorrea: @carolyncole thanks for the catch. I've marked those methods as private. Now they are more private that [Magnum PI](https://en.wikipedia.org/wiki/Magnum,_P.I.) :)

carolyncole (approved):

PR #344, opened 2022-08-29
Use our new XML serialization

### LINKED ISSUE #328: Rework or remove to_xml in PDCMetadata::Resource
The code that exists now in to_xml should either allow and utilize for xml serializations, or be removed in favor of the new PDCSerialization::Datacite.

hectorcorrea:
`PDCMetadata::Resource.to_xml` had a custom XML serialization that we added initially to show how we can output Datacite XML out of our data. This PR drops the custom implementation in favor of the one that we now provide in `PDCSerialization::Datacite.to_xml`. 

Closes #328

### DIFF
--- a/spec/fixtures/files/datacite_basic.xml
+++ b/spec/fixtures/files/datacite_basic.xml
@@ -1,21 +1,23 @@
 <?xml version="1.0"?>
-<resource xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://datacite.org/schema/kernel-4" xsi:schemaLocation="http://datacite.org/schema/kernel-4 https://schema.datacite.org/meta/kernel-4.4/metadata.xsd">
-  <identifier identifierType="DOI">10.5072/example-full</identifier>
-  <titles>
-    <title>hello world</title>
-  </titles>
-  <description descriptionType="Other">this is an example description</description>
+<resource xsi:schemaLocation='http://datacite.org/schema/kernel-4 http://schema.datacite.org/meta/kernel-4/metadata.xsd' xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance' xmlns='http://datacite.org/schema/kernel-4'>
+  <identifier identifierType='DOI'>10.5072/example-full</identifier>
   <creators>
-    <creator nameType="Personal">
+    <creator>
       <creatorName>Miller, Elizabeth</creatorName>
       <givenName>Elizabeth</givenName>
       <familyName>Miller</familyName>
-      <nameIdentifier schemeURI="https://orcid.org" nameIdentifierScheme="ORCID">1234-5678-9012-1234</nameIdentifier>
+      <nameIdentifier nameIdentifierScheme='ORCID' schemeURI='https://orcid.org'>1234-5678-9012-1234</nameIdentifier>
     </creator>
-    <creator nameType="Organization">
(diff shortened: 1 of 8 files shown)


spec/system/work_spec.rb
```diff
@@ -68,10 +67,21 @@
...
+      sign_in user
+      allow_any_instance_of(PDCMetadata::Resource).to receive(:to_xml).and_return(invalid_xml)
+    end
 
     it "Validates the record and prints any errors", js: true do
       visit datacite_validate_work_path(work)
-      expect(page).to have_content "The value has a length of '0'"
+      expect(page).to have_content "This element is not expected"
     end
   end
```
hectorcorrea, line 86: Passing an invalid XML to serialize is now more difficult since we always use our internal `@work.resource` object and that object is serialized to XML via the PDCSerialization::Datacite. The new serialization rejects on creation any object that does not meet the required fields and therefore I had to change the test to force it to return an invalid XML so the code in the controller that does the validation is executed and tested.

spec/models/pdc_metadata/resource_spec.rb
```diff
@@ -24,8 +22,6 @@
     expect(ds.main_title).to eq "hello world"
     expect(ds.resource_type).to eq "Dataset"
     expect(ds.creators.count).to be 2
-    expect(ds.creators.first.affiliations.count).to be 0
-    expect(ds.creators.second.affiliations.count).to be 1
```
hectorcorrea, line 28: We currently do not support affiliations and therefore I removed these two tests. These tests tested functionality that we do not provide but they were here since serialization was free-standing before (i.e. not attached to the PDCSerialization::Datacite class). 

When/if we add affiliations to the workflow (e.g. allow a creator to be affiliated with an organization) then we can add test for it.

spec/fixtures/files/datacite_basic.xml
```diff
@@ -1,21 +1,23 @@
...
+    <creator>
       <creatorName>Miller, Elizabeth</creatorName>
       <givenName>Elizabeth</givenName>
       <familyName>Miller</familyName>
-      <nameIdentifier schemeURI="https://orcid.org" nameIdentifierScheme="ORCID">1234-5678-9012-1234</nameIdentifier>
+      <nameIdentifier nameIdentifierScheme='ORCID' schemeURI='https://orcid.org'>1234-5678-9012-1234</nameIdentifier>
     </creator>
-    <creator nameType="Organization">
-      <creatorName>Princeton University</creatorName>
+    <creator>
```
carolyncole, line 11: @hectorcorrea do we no longer allow for organizations?

hectorcorrea: We never really did. Datacite does allow for organizations but we don't allow users to enter them and therefore the code never really produces organizations. This test used to test that how this _could_ work using our custom XML serialization.

hectorcorrea:
The Coveralls issue makes me think that we could drop the `PDCMetadata::Creator.to_xml` method since we don't need it anymore. This was used when we were using our custom XML serialization.

carolyncole (approved):

hectorcorrea:
@carolyncole I dropped another unused method (and now Coveralls is all happy again). Could you take another look and hopefully re-approve ?

carolyncole (approved):
🥂 Glad to be rid of all that hand created xml!

PR #353, opened 2022-08-30
Restructuring and styling the ActiveStorage and S3 upload components

jrgriffiniii:
Refs #288 (please see https://github.com/pulibrary/pdc_describe/issues/288#issuecomment-1231663795)


app/controllers/works_controller.rb
```diff
@@ -40,8 +44,11 @@ def show
       end
       service = S3QueryService.new(@work, false)
```
carolyncole, line 45: @jrgriffiniii this will get the wrong objects for procreation, since it is set to false.  I was thinking we would want to retrieve both, but that might be something for discussion.

app/views/works/uploads/_post_curation_s3_resources.html.erb
```diff
@@ -0,0 +1,46 @@
...
+                <th scope="col"><span><%= t('works.form.post_curation_uploads.created_at') %></span></th>
+              </tr>
+            </thead>
+            <tbody>
+              <% @post_curation_s3_resources.each_with_index do |file, ix| %>
+                <tr class="files">
+                  <td>
+                    <span>
+                      <i class="bi bi-file-arrow-down"></i>
+                      <a href="<%= file.filename %>" class="documents-file-link" target="_blank" title="<%= file.filename %>"><%= truncate(file.filename.to_s, length: 80) %></a>
```
carolyncole, line 24: Instead of a truncate should we be getting the filename out of the path?

carolyncole (changes requested):
@jrgriffiniii The work that I did earlier (while you were away) creates attachments out of the pre-curated s3 files. https://github.com/pulibrary/pdc_describe/blob/main/app/controllers/works_controller.rb#L38-L40 So for pre-curated files there should always be a blob.  I was unsure if the same approach was desired for post curated, so I just left them as non active storage objects.  That said I think we need a holistic approach so we are not dealing with so many cases.  I wonder if a discussion around this topic is merited.  Aka how do the ActiveStorage items get cleaned up when the work is approved and the files are moved in S3.  Should those pointers just be moved instead of querying S3.  In theory there should not be new files just showing up in the post curation s3 bucket.

app/controllers/works_controller.rb
jrgriffiniii: I am sorry, I see this now, I can fix these errors immediately.

app/views/works/show.html.erb
```diff
@@ -110,14 +110,27 @@
...
 
-<% if @work.pre_curation_uploads.present? %>
-  <%= render(partial: 'works/uploads') %>
-<% end %>
-
-<%= render(partial: 'works/files') %>
+<section class="uploads">
+  <% if @work.approved? %>
+    <h2 class="h2"><%= t('works.uploads.post_curation.heading') %></h2>
+    <% if @work.post_curation_uploads.present? %>
```
jrgriffiniii, line 118: This (ActiveStorage) should be removed for post-curation resources

app/views/works/show.html.erb
```diff
@@ -110,14 +110,27 @@
...
+    <% end %>
+    <% if @post_curation_s3_resources.present? %>
+      <%= render(partial: 'works/uploads/post_curation_s3_resources') %>
+    <% end %>
+  <% else %>
+    <h2 class="h2"><%= t('works.uploads.pre_curation.heading') %></h2>
+    <% if @work.pre_curation_uploads.present? %>
+      <%= render(partial: 'works/uploads/pre_curation_attachments') %>
+    <% end %>
+    <% if @pre_curation_s3_resources.present? %>
```
jrgriffiniii, line 129: This should be removed for pre-curation resources (as these are persisted as ActiveStorage::Blob objects)

carolyncole (changes requested):
Sorry!  A couple more small tweaks.  Overall this looks fantastic!

carolyncole (approved):
🎉

Issue #332, opened 2022-08-24
A submitter should not have the ability to approve a Work


carolyncole:
Carolyn (cac9) does not exists in the super admin list, but she can approve a work on staging. See https://github.com/pulibrary/pdc_describe/blob/main/config/superadmins.yml
![Screen Shot 2022-08-24 at 2.20.04 PM.png](https://images.zenhubusercontent.com/5c925e2d9d1bd30b8d310436/7521405b-935b-4a26-8be0-b32dcc3816ea)

We are assuming we might want to utilize CanCanCan and Rolify Gems for this implementation.

This will be complete when:

- [ ] A regular submitter does not see the approve button in the UI
- [ ] The work will not allow a regular submitter to run the approve action
- [ ] Create a automated test that submitter can not approve the work
- [ ] Create a automated test that Super Admins can still approve the work
- [ ] Create a automated test that a administrator of the collection can still approve the work
- [ ] Create a automated test that an administrator of another collection can not approve the work

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-Workcycles-61a4f1a12a399b001730f65a/issues/pulibrary/pdc_describe/332?planning-poker) with Zenhub @bess @jrgriffiniii @kelynch

PR #355, opened 2022-08-31
Utilizing rolify for collection admin and submitters

carolyncole:
Also added migration for UserCollection to roles, so existing collections can have the same access

The plan would be to merge (or deploy to staging) migrate and make sure folks still have access.  After that there would be a follow up PR to delete UserCollection from the system.


app/models/user_collection.rb
```diff
@@ -20,6 +20,15 @@ def self.add_admin(user_id, collection_id)
     end
   end
 
+  def migrate
+    rolify_fole = if role == "ADMIN"
```
bess, line 24: Should this be `rolify_role`?

app/models/user_collection.rb
```diff
@@ -20,6 +20,15 @@ def self.add_admin(user_id, collection_id)
...
+                    :collection_admin
+                  else
+                    :submitter
+                  end
+    user.add_role rolify_fole, collection
```
bess, line 29: and here

bess (dismissed):
This looks great! I saw one place that I think might be a typo? But I could be wrong. Happy to merge.

app/models/user_collection.rb
carolyncole: Thanks @bess!  That is a typo.  We are not keeping track of Foles 🐴 in the system

bess (approved):

carolyncole:
refs #332

PR #364, opened 2022-09-01
Removing UserCollections from the system

carolyncole:
Now that we are utilizing rolify, we no longer need UserCollection records.
Refs #332


spec/controllers/collections_controller_spec.rb
```diff
@@ -129,6 +129,12 @@
       expect(response.status).to eq 400
     end
 
+    it "fails for an unknow user" do
```
bess: * unknown

bess (approved):

PR #377, opened 2022-09-02
Allow access to default collection

carolyncole:
Even if another role has already been assigned.  Each user should get access to thier default collection

Also removes reference to setup_user_default_collections as it no longer exists.  Utilizing assign_default_role instead

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -203,7 +203,7 @@ class User < ApplicationRecord
   end
 
   def assign_default_role
-    add_role(:submitter, default_collection) if roles.blank?
+    add_role(:submitter, default_collection) unless has_role?(:submitter, default_collection)
   end
 end
 # rubocop:enable Metrics/ClassLength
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -203,7 +203,7 @@ def pending_notifications_count
   end
 
   def assign_default_role
-    add_role(:submitter, default_collection) if roles.blank?
+    add_role(:submitter, default_collection) unless has_role?(:submitter, default_collection)
```
jrgriffiniii, line 206: 👍

lib/tasks/users.rake
```diff
@@ -52,14 +52,14 @@ namespace :users do
       user.default_collection_id = Collection.research_data.id
       if fixit
         user.save!
-        user.setup_user_default_collections
+        user.assign_default_role
```
jrgriffiniii, line 55: 👍

jrgriffiniii (approved):

Issue #307, opened 2022-08-17
A collection admin can edit any object in their collection


carolyncole:
As a collection admin, curator, I CAN edit a record created by another user in the collection I administer

- [x] Create a system spec to test if curator users can edit a work created by another user in a collection they curate
- [x] Create a system spec to test if curator users can edit a collection to add curators and submitters
- [x] Create a system spec to test if curator users can NOT edit a collection to add curators and submitters for collections they do NOT curate
- [x] Create a system spec to test if curator users can NOT edit a work created by another user in a collection they do NOT curate
- [x] If the test does not pass create a draft PR and ticket the need

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-Workcycles-61a4f1a12a399b001730f65a/issues/pulibrary/pdc_describe/307?planning-poker) with ZenHub @bess @hectorcorrea @jrgriffiniii

PR #382, opened 2022-09-06
Test that a collection admin can edit any item in their collection

bess:
Work toward #307

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -28,6 +28,12 @@ class Collection < ApplicationRecord
     User.with_role(:super_admin)
   end
 
+  ##
+  # The current user adds a new admin user to a collection.
+  # For use in the UI - we need to check whether the current_user
+  # has the right permissions to add someone as an admin_user.
+  # @param [User] current_user - the currently logged in user
+  # @param [User] admin_user - the user who will be granted admin rights on this collection
   def add_administrator(current_user, admin_user)
     if current_user.has_role?(:super_admin) || current_user.has_role?(:collection_admin, self)
       if admin_user.has_role? :collection_admin, self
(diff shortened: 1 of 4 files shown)


app/models/collection.rb
```diff
@@ -28,6 +28,12 @@ def super_administrators
     User.with_role(:super_admin)
   end
 
+  ##
```
carolyncole, line 31: ❤️

spec/system/authz_admin_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+
+      it "can add submitters to the collection" do
+        login_as research_data_moderator
+        expect(research_data_moderator.can_admin?(Collection.research_data)).to eq true
+        expect(new_submitter.can_submit?(Collection.research_data)).to eq false
+        visit collection_path(Collection.research_data)
+        fill_in "submitter-uid-to-add", with: new_submitter.uid
+        click_on "Add Submitter"
+        # This does not work. Bug ticketed here: https://github.com/pulibrary/pdc_describe/issues/372
```
carolyncole, line 43: @bess I would love if these would be pending tests that would then fail if the bug is fixed vs passing commented out tests.  It would make them less likely to get ignored when fixing the failure.  At the moment, your PR and My PR could both be merged and I would never have to fix this test.

carolyncole (changes requested):
@bess Just one small change if you do not mind.  I just want to make it harder for these nice tests to get left in a bad state...

spec/system/authz_admin_spec.rb
bess: Great point. Fixing now.

carolyncole (approved):

PR #397, opened 2022-09-08
Moving from webpacker to vite

carolyncole:
Utilized https://vite-ruby.netlify.app/guide/migration.html#webpacker-%F0%9F%93%A6 and referenced https://github.com/pulibrary/orangelight/pull/3164

This was complicated by the fact that the servers utilize the describe path in their base url.


bess (approved):
🚀 LGTM! I tried this out in staging and it seems to work great.

bess (approved):

PR #403, opened 2022-09-12
Allow files to be displayed after edit

### LINKED ISSUE #376: Edit with uploads causes an exception
for example: when you remove the description and the click save on this work:
https://pdc-describe-staging.princeton.edu/describe/works/75/edit

I am unsure why the initial works, but I am guessing the data is just not set up the same way on an error.

```
Completed 500 Internal Server Error in 2521ms (ActiveRecord: 115.7ms | Allocations: 54342)


  
ActionView::Template::Error (undefined method `each_with_index' for nil:NilClass):
    25:                   </td>
    26:                 </tr>
    27:               <% else %>
    28:                 <% @uploads.each_with_index do |file, ix| %>
    29:                   <tr class="files">
    30:                     <td>
    31:                       <span>
  
app/views/works/_uploads_form.html.erb:28
app/views/works/_form.html.erb:35
app/views/works/_form.html.erb:15
```

### LINKED ISSUE #400: When I edit and save a work the files should not be deleted
When I create a work and upload files via the UI
Then I go to the show page the files are present
When I edit the work and change the description
Then the show page the files are no longer present

This is complete when:
- [ ] I can edit a work that has files and the files remain
- [ ] There is an automated test that proves that editing the description of a work does not remove the files in a work.

carolyncole:
Also makes sure the files still exists both before and after an edit. 
Also makes sure the files still display both before and after an update error.

fixes #376 and fixes #400

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -12,7 +12,6 @@ Rails.application.configure do
   # Compile and load CSS when running tests if we're running in browser
   if ENV["RUN_IN_BROWSER"]
     config.assets.compile = true
-    config.asset_host = "http://localhost:3000"
   end
 
   config.cache_classes = false
--- a/spec/factories/work.rb
+++ b/spec/factories/work.rb
@@ -4,7 +4,7 @@ FactoryBot.define do
   factory :work do
     factory :draft_work do
       transient do
-        doi { "https://doi.org/10.34770/123-abc" }
+        doi { "10.34770/123-abc" }
       end
       collection { Collection.research_data }
       state { "draft" }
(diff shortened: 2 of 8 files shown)


config/environments/test.rb
```diff
@@ -12,7 +12,6 @@
   # Compile and load CSS when running tests if we're running in browser
   if ENV["RUN_IN_BROWSER"]
     config.assets.compile = true
-    config.asset_host = "http://localhost:3000"
```
carolyncole, line 15: Vite does not like assets to be served from development.

spec/factories/work.rb
```diff
@@ -4,7 +4,7 @@
   factory :work do
     factory :draft_work do
       transient do
-        doi { "https://doi.org/10.34770/123-abc" }
+        doi { "10.34770/123-abc" }
```
carolyncole, line 7: DOIs do not contain a URL.

bess (approved):
🚀 🎸 🤩

PR #413, opened 2022-09-15
Required field validation 

hectorcorrea:
Sends the user back to the Wizard while the dataset is in `draft` and prevents a dataset from moving from `draft` to `awaiting-approval` until all required fields have been entered.

edit-validate


app/controllers/works_controller.rb
```diff
@@ -8,7 +8,7 @@
 # rubocop:disable Metrics/MethodLength
 # rubocop:disable Style/For
 class WorksController < ApplicationController
-  around_action :rescue_aasm_error, only: [:approve, :withdraw, :resubmit, :completed, :create]
+  around_action :rescue_aasm_error, only: [:approve, :withdraw, :resubmit, :create]
```
hectorcorrea, line 11: We handle completed (now validate) with its own code.

app/controllers/works_controller.rb
```diff
@@ -147,11 +147,18 @@ def review
...
-    redirect_to user_url(current_user)
+    @uploads = @work.uploads
+    @wizard_mode = true
+    begin
+      @work.complete_submission!(current_user)
+      redirect_to user_url(current_user)
+    rescue AASM::InvalidTransition
+      # Work couldn't transition, send the user back to the edit page
+      render "edit", status: :unprocessable_entity
+    end
```
hectorcorrea, line 161: Notice the specific handle for validate renders the Edit page (so the user can fix the errors)

config/routes.rb
```diff
@@ -23,7 +23,7 @@
   get "works/:id/file-other", to: "works#file_other", as: :work_file_other
   get "works/:id/review", to: "works#review", as: :work_review
   post "works/:id/review", to: "works#review"
-  post "works/:id/completed", to: "works#completed", as: :work_completed
+  post "works/:id/validate", to: "works#validate", as: :work_validate
```
hectorcorrea, line 26: I changed the URL so that if the user looks at the URL in the browser (e.g. when there is an error during validation) they don't get the wrong impression that they have "completed" but see that we are still in the validation part of the process.

app/views/works/show.html.erb
```diff
@@ -40,7 +40,7 @@
 </style>
 
 <div>
-  <%= link_to("Edit", edit_work_path(@work), class: "btn btn-primary") %>
+  <%= link_to("Edit", edit_work_path(@work, wizard: @work.draft? ? true : nil), class: "btn btn-primary") %>
```
carolyncole, line 43: 🎉 Seems good for most users to be utilizing the wizard.

app/controllers/works_controller.rb
carolyncole: Note, you would not need to repeat this code here.  In a PR earlier this week I broke out a different location for errors during the creation: https://github.com/pulibrary/pdc_describe/blob/main/app/controllers/works_controller.rb#L284-L290

spec/controllers/works_controller_spec.rb
```diff
@@ -647,10 +647,10 @@
           work.resource.description = nil
           work.save
           sign_in user
-          post :completed, params: { id: work.id }
+          post :validate, params: { id: work.id }
+          expect(response).to render_template("edit")
           expect(response.status).to be 422
           expect(work.reload).to be_draft
```
carolyncole, line 653: It would still be good to validate the errors here

spec/system/work_spec.rb
```diff
@@ -103,6 +103,24 @@
     end
   end
 
+  context "when editing a work" do
+    let(:draft_work) { FactoryBot.create(:draft_work) }
+    let(:completed_work) { FactoryBot.create(:completed_work) }
+    let(:user) { draft_work.created_by_user }
+
+    it "uses the wizard if the work is in draft" do
```
carolyncole, line 111: ❤️

carolyncole (dismissed):
Could you please change error_action to include validate, that way we a using the same pattern, or if you prefer your pattern then change create to match yours and remove the edit_action pattern I added.
```
    def error_action
      if action_name == "create"
        :new
      elsif action_name == "validate"
        :edit
      else
        :show
      end
    end
```

spec/controllers/works_controller_spec.rb
hectorcorrea: Good idea. Done.

app/controllers/works_controller.rb
hectorcorrea: Good idea, done.

hectorcorrea:
Made the changes suggested.

jrgriffiniii (approved):

jrgriffiniii:
As the requested changes were addressed (please see https://github.com/pulibrary/pdc_describe/pull/413#issuecomment-1249788893), I am going to proceed with merging this. Thank you very much @hectorcorrea!

Issue #388, opened 2022-09-07
When a work has an existing DOI, redirect it


bess:
The works we're migrating from PPPL will have existing DOIs that are not currently part of the DataSpace record. When a work with an existing DOI is approved, we need to redirect that DOI to the PDC Discovery URL for the work. 

ACCEPTANCE CRITERIA
- [x] I can input a DOI on the initial screen instead of having the system register one for me
- [x] Validate the DOI when it has been entered
- [ ] That DOI will, at the "Publish" step in the workflow, be re-directed to the PDC Discovery URL for this work
- [ ] Do something with DOIs we did not create ?? Still figuring out the something ??
- [x] There is an automated test for allowing the user to enter a DOI on the first step of the wizard for an existing DOI, which proves that no DOI is minted in this case.
- [ ] There is an automated test for allowing a already published DOI to be updated with a new URL

astrochun:
@bess and @kelynch , I have the following JSON between DOE DOIs and the ARKs. This was outputted by `dspace-osti`. Hopefully this helps.
 
[redirects.txt](https://github.com/pulibrary/pdc_describe/files/9506342/redirects.txt)

carolyncole:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-Workcycles-61a4f1a12a399b001730f65a/issues/pulibrary/pdc_describe/388?planning-poker) with Zenhub @bess

carolyncole:
I did some testing of updating with items in our DOI prefix and items not in our DOI prefix (most if not all of the PPPL DOIs).  We do not have permission to update the DOIs with a new URL in a prefix that we do not own.

I saw responses like 
```
You are not authorized to access this resource
```

astrochun:
> I did some testing of updating with items in our DOI prefix and items not in our DOI prefix (most if not all of the PPPL DOIs). We do not have permission to update the DOIs with a new URL in a prefix that we do not own.
> 
> I saw responses like
> 
> ```
> You are not authorized to access this resource
> ```

I'm not surprised. It may be best to use the DOI prefix (`10.?????/`) in your logical statement before doing so.

carolyncole:
I noticed that OSTI is also utilizing DataCite.  I went ahead and contacted DataCite to see if they had recommendations for how to request a URL changed on another prefix.

kelynch:
We may be able to just use the EZID API to update the ark URL redirects, which are what powers the OSTI DOI redirects. 
[...]

PR #414, opened 2022-09-16
Validating DOI based on format and then checking if it exists

carolyncole:
Converting the ENV references in the code to a config file to make it more consistent with the rest of the configuration Added a url for DOI checking

refs #388


bess (approved):
⭐

Issue #17, opened 2022-03-22
DataCite - Using an existing ARK updates the ezid metadata


bess:
When we're migrating an object, it will already have an ARK. When that's the case, do not register a new ARK. Instead, update the metadata at ezid to point to the new location.

ACCEPTANCE CRITIERIA
- [x] The DataCite input form has a place for an existing ARK
- [x] When that field is filled out, the form validates that the ARK is legit before allowing submission
- [x] When the existing ARK field is filled out with a legit ARK, no new ARK is registered
- [x] When the existing ARK field is filled out with a legit ARK, the metadata at ezid is updated to point to the new location of the object

carolyncole:
@bess It was not clear, but there was already ARK validation in the system: https://github.com/pulibrary/pdc_describe/pull/281/files#diff-ebe838f9d71d26b6227d6d37c96a7e2d93fb68a80d55075c7f2374bfcb4688abR339-R343

carolyncole:
No new ark is registered after this commit: aa9c96ee97e2c3d8e205e1bd556a9d86c35fcef2
Should we really test a negative?

bess:
@carolyncole Not totally sure what to do here, since we're not totally clear on whether we're registering new arks on the regular. As long as we're pretty sure we're not proliferating new arks I'm fine with whatever you deem appropriate testing.

carolyncole:
The final item can be checked off by this existing test: https://github.com/pulibrary/pdc_describe/blob/main/spec/system/migrate_submission_spec.rb#L103-L145

PR #416, opened 2022-09-16
@carolyncole Adding code to only update DOIs that are part of our prefix

### LINKED ISSUE #388: When a work has an existing DOI, redirect it

carolyncole:
Also added test to make sure Arks are updated and Honeybadger is notified if the DOI can not be updated and there is no ARK

fixes #388 
refs #17 Adds as part of the tests to make sure the Ark is updated

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -519,12 +519,14 @@ class Work < ApplicationRecord
     def publish_doi(user)
       if Rails.env.development? && Rails.configuration.datacite.user.blank?
         Rails.logger.info "Publishing hard-coded test DOI during development."
-      else
+      elsif doi.starts_with?(Rails.configuration.datacite.prefix)
         result = data_cite_connection.update(id: doi, attributes: doi_attributes)
         if result.failure?
           message = "@#{curator_or_current_uid(user)} Error publishing DOI. #{result.failure.status} / #{result.failure.reason_phrase}"
           WorkActivity.add_system_activity(id, message, user.id, activity_type: "DATACITE_ERROR")
         end
+      elsif ark.blank? # we can not update the url anywhere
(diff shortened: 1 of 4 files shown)


app/models/work.rb
```diff
@@ -519,12 +519,14 @@ def validate_creators
...
         Rails.logger.info "Publishing hard-coded test DOI during development."
-      else
+      elsif doi.starts_with?(Rails.configuration.datacite.prefix)
         result = data_cite_connection.update(id: doi, attributes: doi_attributes)
         if result.failure?
           message = "@#{curator_or_current_uid(user)} Error publishing DOI. #{result.failure.status} / #{result.failure.reason_phrase}"
           WorkActivity.add_system_activity(id, message, user.id, activity_type: "DATACITE_ERROR")
         end
+      elsif ark.blank? # we can not update the url anywhere
+        Honeybadger.notify("Publishing for a DOI we do not own and no ARK is present: #{doi}")
```
bess, line 529: 👍🏻

bess (approved):

PR #418, opened 2022-09-19
Add an RSS feed for works, to be used by discovery for indexing

bess:
Work toward https://github.com/pulibrary/pdc_discovery/issues/322

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -12,6 +12,10 @@ class WorksController < ApplicationController
 
   def index
     @works = Work.all
+    respond_to do |format|
+      format.html
+      format.rss { render layout: false }
+    end
   end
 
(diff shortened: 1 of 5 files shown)


escowles:
I'm curious: why use RSS here instead of OAI-PMH?

bess:
@escowles My thought was that RSS is easy to implement and the main thing I'm trying to accomplish here is a list of content to index for discovery. We know we eventually want an OAI-PMH feed on Discovery (or do we? I've also heard we want ResourceSync instead?) I thought I'd be opening a can of worms by implementing OAI and that it would be scope creep past what I need to deliver for the ticket I'm working on. Glad to revisit this decision if you have more information than I do or would prefer a different approach.

jrgriffiniii (approved):

PR #434, opened 2022-09-23
Adds nil check

### LINKED ISSUE #433: Error on save
The Work controller has the following code in the `update()` method (https://github.com/pulibrary/pdc_describe/blob/de60613bce87bc0074a679784c5388e20b237602/app/controllers/works_controller.rb)

```
    if @work.approved?
      upload_keys = work_params[:deleted_uploads]
      deleted_uploads = WorkUploadsEditService.find_post_curation_uploads(work: @work, upload_keys: upload_keys)
```

This code throws an error when `work_params[:deleted_uploads]` is `nil`. 

There is a validation in `WorkUploadsEditService.find_post_curation_uploads` that defaults the value to `[]` (see https://github.com/pulibrary/pdc_describe/blob/de60613bce87bc0074a679784c5388e20b237602/app/services/work_uploads_edit_service.rb#17) but that validation does not account for when the actual parameter passed is `nil`

hectorcorrea:
Fixes #433

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -109,7 +109,7 @@ class WorksController < ApplicationController
     }
 
     if @work.approved?
-      upload_keys = work_params[:deleted_uploads]
+      upload_keys = work_params[:deleted_uploads] || []
       deleted_uploads = WorkUploadsEditService.find_post_curation_uploads(work: @work, upload_keys: upload_keys)
 
       return head(:forbidden) unless deleted_uploads.empty?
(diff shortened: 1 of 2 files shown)


hectorcorrea:
@jrgriffiniii This is the PR with the test that we worked out. Feel free to update it if you think in a better approach. Thanks for you help on this.

jrgriffiniii:
I am sorry for the delay, this is the approach which I was trying to implement: https://github.com/pulibrary/pdc_describe/pull/436

jrgriffiniii (approved):

hectorcorrea:
Merging to fix the save process for existing works. A better approach is coming in #436

PR #439, opened 2022-09-26
Reduces the default retry_interval

### LINKED ISSUE #432: ARK validation is too slow
The process to validate an ARK is too slow __when the ARK is invalid__, this makes the UI feel like the system is stuck.

Currently when the user enters an invalid ARK in the Wizard we validate it as soon as they click "Save Work" to move to the next step. When the work has been approved and the user is not using the Wizard the validation happens when the click "Save Work" to complete their edit.

Review if there is a way to handle this better so the user does not think the system is frozen/unresponsive.

See also #426.

hectorcorrea:
The culprit on the slow validation of the ARK is that the gem that we are using does an automatic retry when there is an error with the id provided. By default the gem waits 15 seconds in between retries and it retries twice. This means that if the user provides an invalid ARK the validation takes between 33-35 seconds. Of those, 30 seconds are due to the retry logic.

This is the retry logic on the gem: https://github.com/duke-libraries/ezid-client/blob/dfcf7f49995560ed48df407560c4fe3fb6dbfa7b/lib/ezid/requests/request.rb#L49-L64

This PR changes the default retry value to `1` so that retries happen (almost) immediately and a user waits between 3-5 seconds rather than 33-35 seconds.

This is the code on the gem that shows how the retry value is initialized: https://github.com/duke-libraries/ezid-client/blob/dfcf7f49995560ed48df407560c4fe3fb6dbfa7b/lib/ezid/configuration.rb#L44-L52

Closes #432

### DIFF
--- a/config/initializers/ezid_client.rb
+++ b/config/initializers/ezid_client.rb
@@ -4,4 +4,5 @@ Ezid::Client.configure do |config|
   config.default_shoulder = "ark:/99999/fk4" || ENV["EZID_DEFAULT_SHOULDER"]
   config.user = "apitest" || ENV["EZID_USER"]
   config.password = "apitest" || ENV["EZID_PASSWORD"]
+  config.retry_interval = 1 || ENV["EZID_RETRY_INTERVAL"]
 end


bess (approved):
Great sleuthing! ⭐

PR #449, opened 2022-09-28
Renders collection names as links

### LINKED ISSUE #334: Curator dashboard should link to the collections they curate
On the Curator's dashboard we show which collection the user curates, but do not give them a way to access that collection.
Screen Shot 2022-08-24 at 2 49 01 PM

This will be complete when
- [ ] The collection list is a collection of links instead of just plain text

hectorcorrea:
collection_name_as_link

Closes #334


app/views/users/show.html.erb
```diff
@@ -22,7 +22,10 @@
 <% if @user.super_admin? %>
   <!-- not sure we want to display the collections here, it's pointless -->
 <% elsif @user.curator? %>
-  <p>Curator for: <%= @user.admin_collections.map(&:title).to_sentence %></p>
+  <p>Curator for:
+    <%= @user.admin_collections.map do |collection|
+      link_to(collection.title, collection_url(collection))
```
carolyncole, line 27: @hectorcorrea is there a reason not to utilize relative paths?
```suggestion
      link_to(collection.title, collection_path(collection))
```

spec/spec_helper.rb
```diff
@@ -111,3 +111,7 @@
   #   # as the one that triggered the failure.
   #   Kernel.srand config.seed
 end
+
+def capybara_root_url
```
carolyncole, line 115: I believe if we utilize relative urls this would not be needed

spec/system/user_show_spec.rb
```diff
@@ -12,5 +13,13 @@
       visit user_path(user_admin)
       expect(page.html.include?(sort_value)).to be true
     end
+
+    it "renders the collection names as links", js: true do
+      sign_in moderator_user
+      visit user_path(moderator_user)
+      pppl_url = capybara_root_url + collection_path(Collection.plasma_laboratory)
```
carolyncole, line 20:
```suggestion
      pppl_url = collection_path(Collection.plasma_laboratory)
```

carolyncole (changes requested):
I'm thinking utilizing relative path would make this feature simpler in the long run.

hectorcorrea:
@carolyncole good idea, I switched to use relative paths.

carolyncole (approved):

Issue #421, opened 2022-09-20
Describe Form: metadata entry tabs


hhkitsune:
## Description

Users will have three named tabs to help navigate the metadata entry page. 

*This feature ticket focuses on the tabs themselves. Other tickets will focus on the content within each tab.*

### The tabs will organize metadata properties into three areas:
**1. Required Metadata**
- The depositor and curator can input to all fields except the Datacite DOI
- These are properties that are required by Datacite to mint the DOI and those which we require locally  

**2. Additional Metadata** 
- The depositor and curator can input to all fields

**3. Curator Controlled** 
- These include automated and manual fields 
- Only the curator can change these fields
- The depositor may only see what is inputted and may request the curator to make any changes

## Image shows the three tabs (at left)

![Screen Shot 2022-09-20 at 4 34 51 PM](https://user-images.githubusercontent.com/50670293/191359308-f6f580dd-f6cf-42ec-ab89-b2e1ff6ec907.jpeg)

## Acceptance Criteria

- [ ] There are Three tabs (Required Metadata, Additional Metadata, Curator Controlled)described above are created
- [ ] The Curator Controlled is only editable by Collection Admins and Super Admins
- [ ] There are system test(s) to prove that a submitter can **not** edit the fields on the Curator Controlled tab
- [ ] There are system test(s) to prove that a collection Admin can edit the fields on the Curator Controlled tab
- [ ] Tabs will highlight to show users which tab is currently active (already does)

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-Workcycles-61a4f1a12a399b001730f65a/issues/pulibrary/pdc_describe/421?planning-poker) with Zenhub @bess @hectorcorrea @jrgriffiniii

PR #452, opened 2022-09-29
Renaming Identifiers tab to Curator Controlled

carolyncole:
Also only allowing collection admins and super admins access to the edit form

refs #421

Will still need to control access to the fields at the controller level, but this seemed like a good first step.


hectorcorrea (dismissed):

hectorcorrea:
The idea looks good except for the failing tests :(

carolyncole:
@hectorcorrea thanks!  I updated the tests.  I forgot there were a couple tests that needed to utilize a collection admin user to have access to editing doi, and also my rebase with main caused s3 to need stubbing.

hectorcorrea (approved):

PR #464, opened 2022-10-04
Migrate data using system specs

bess:
Work toward https://github.com/pulibrary/pdc_discovery/issues/340

Here is one of our sample objects on the pdc-describe staging server:

Screen Shot 2022-10-04 at 3 18 23 PM


carolyncole (changes requested):
Looks great @bess, but it looks like you have the ARK commit in this PR too.

PR #466, opened 2022-10-06
Explicitly deleting files as they change

### LINKED ISSUE #394: In the Edit Form allow files to be uploaded outside the wizard
Add a tab under additional metadata to allow the user to upload files.
This should also show the existing files attached to a work instead of them being below the metadata fields.

This is complete when:

- [ ] The edit form, not in wizard mode, allows the user to enter files when none have been previously uploaded.
- [ ] The existing files uploaded would display on the files tab, not below the metadata
- [ ] There is an Automated test that allows a user to exit wizard mode and still be able to upload files

carolyncole:
We no longer rely on ActiveStorage to do the Right thing, by setting the attribute pre_curation_uploads. 
Instead we explicitly call purge on items that are deleted.  ActiveStorage will utilize a purge_later which introduces a timing issue when deleting items and then reloading from S3 for the show page, immediately after the deletes.

fixes #394

### DIFF
--- a/app/services/work_uploads_edit_service.rb
+++ b/app/services/work_uploads_edit_service.rb
@@ -1,15 +1,21 @@
 # frozen_string_literal: true
 class WorkUploadsEditService
   class << self
-    def precurated_file_list(work, work_params)
-      if work_params.key?(:pre_curation_uploads)
-        work_params[:pre_curation_uploads]
-      elsif work_params.key?(:replaced_uploads)
-        replace_uploads(work.pre_curation_uploads, work_params[:replaced_uploads])
-      elsif work_params.key?(:deleted_uploads)
-        remaining_pre_curation_uploads(work.pre_curation_uploads, work_params[:deleted_uploads])
-      else
-        work.pre_curation_uploads.map(&:blob)
+    def update_precurated_file_list(work, work_params)
+      if work_params.key?(:deleted_uploads) || work_params.key?(:pre_curation_uploads) || work_params.key?(:replaced_uploads)
+        if work_params.key?(:deleted_uploads)
+          delete_pre_curation_uploads(work.pre_curation_uploads, work_params[:deleted_uploads])
+        elsif work_params.key?(:pre_curation_uploads)
+          work.pre_curation_uploads.each(&:purge)
+          work.reload # reload the work to pick up the changes in the attachments
+          Array(work_params[:pre_curation_uploads]).each { |new_upload| work.pre_curation_uploads.attach(new_upload) }
+        elsif work_params.key?(:replaced_uploads)
+          replace_uploads(work, work_params[:replaced_uploads])
+        end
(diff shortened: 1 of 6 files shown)


app/services/work_uploads_edit_service.rb
```diff
@@ -1,15 +1,21 @@
...
-    def precurated_file_list(work, work_params)
-      if work_params.key?(:pre_curation_uploads)
-        work_params[:pre_curation_uploads]
-      elsif work_params.key?(:replaced_uploads)
-        replace_uploads(work.pre_curation_uploads, work_params[:replaced_uploads])
-      elsif work_params.key?(:deleted_uploads)
-        remaining_pre_curation_uploads(work.pre_curation_uploads, work_params[:deleted_uploads])
-      else
-        work.pre_curation_uploads.map(&:blob)
+    def update_precurated_file_list(work, work_params)
```
jrgriffiniii, line 4: 👍

app/services/work_uploads_edit_service.rb
```diff
@@ -1,15 +1,21 @@
...
+      if work_params.key?(:deleted_uploads) || work_params.key?(:pre_curation_uploads) || work_params.key?(:replaced_uploads)
+        if work_params.key?(:deleted_uploads)
+          delete_pre_curation_uploads(work.pre_curation_uploads, work_params[:deleted_uploads])
+        elsif work_params.key?(:pre_curation_uploads)
+          work.pre_curation_uploads.each(&:purge)
+          work.reload # reload the work to pick up the changes in the attachments
```
jrgriffiniii, line 10: 👍 this was affecting my earlier feature implementations

app/services/work_uploads_edit_service.rb
```diff
@@ -21,32 +27,25 @@ def find_post_curation_uploads(work:, upload_keys: [])
...
-            if replaced_uploads_params.key?(key)
-              replaced = replaced_uploads_params[key]
-              updated_uploads << replaced
-            else
-              updated_uploads << existing.blob
-            end
+            next unless replaced_uploads_params.key?(key)
+            new_uploads << replaced_uploads_params[key]
+            existing.blob.purge
+            existing.purge
```
jrgriffiniii, line 37: This is not a criticism, but I believe that this might be redundant (please see https://github.com/rails/rails/blob/main/activestorage/app/models/active_storage/attachment.rb#L32). I suspect that `existing.purge` should also invoke `purge` for the associated `blob`.

jrgriffiniii (approved):

app/services/work_uploads_edit_service.rb
carolyncole: @jrgriffiniii yes, I think you are correct.  The reload was needed and this was put in to try and figure out why the Purge Job was still getting called before I put in the reload.   I will take it out.

jrgriffiniii (approved):

bess (approved):

PR #482, opened 2022-10-12
Adds support for Work contributors

### LINKED ISSUE #184: Contributors
Add support for contributors to a Dataset. Contributors are similar to creators but they have a specified role (https://support.datacite.org/docs/datacite-metadata-schema-v44-recommended-and-optional-properties#7-contributor).

* Add a separate tab for Contributors to the data entry form
* Allow a user to add one or more contributors to the dataset (similar UI as we do for creators)
* Allow user to indicate a role (`contributorType`) for each of the contributors -- notice that this is a controlled list.
* Display contributors in the Show page.
* Render contributors in the DataCite XML of the dataset.

hectorcorrea:
Added Contributors to the Additional Metadata tab. They work like creators, except they are optional and they have an extra attribute ("role")

Screen Shot 2022-10-12 at 2 30 18 PM

Closes #184


app/views/works/_contributors_table.html.erb
```diff
@@ -0,0 +1,42 @@
...
+      <th>Given name</th>
+      <th>Family name</th>
+      <th>Role</th>
+      <th>&nbsp;</th> <!-- move icon-->
+      <th>&nbsp;</th> <!-- delete icon-->
+    </tr>
+  </tbody>
+</table>
+
+<% if @work %>
```
carolyncole, line 25: @hectorcorrea This seems odd...  I would be surprised that this form would get displayed without the work being set at least to a Work.new.  I can not imagine you adding this unless it was needed, but we may need to ticket figuring out why...

spec/system/authz_submitter_spec.rb
```diff
@@ -15,6 +15,8 @@
       Collection.create_defaults
       stub_s3
       stub_datacite(host: "api.datacite.org", body: datacite_register_body(prefix: "10.34770"))
+      # Make the screen larger so the save button is alway on screen. This avoids random `Element is not clickable` errors
```
carolyncole, line 18: @hectorcorrea another thing we may want to ticket.  Should we just make the screen larger to begin with?  I was wondering this myself when I copied this the other week.

carolyncole (changes requested):
I made a comment on a couple if things we may want to ticket (They are not just part of this PR).  Could you please add a system test to edit the contributors (or did I miss it)?  Otherwise this looks fantastic.

app/views/works/_contributors_table.html.erb
hectorcorrea: Good catch. I removed it.

spec/system/work_spec.rb
```diff
@@ -159,6 +172,42 @@
...
+      expect(contributor_text).to eq("Robert Smith  Simon Gallup")
+
+      # drag the first contributor to the second contributor
+      source = page.all(".bi-arrow-down-up")[0].native
+      target = page.all(".bi-arrow-down-up")[1].native
+      builder = page.driver.browser.action
+      builder.drag_and_drop(source, target).perform
+      contributor_text_after = page.all("tr")[1..2].map { |each| each.all("input").map(&:value) }.flatten.join(" ").strip
+
+      # This is really strange, but my local machine likes to drag from bottom to top and CircleCI likes to drag
```
carolyncole, line 198: @hectorcorrea did you see this same behavior, or did you just copy paste? I'm happy for the comment, and just curious...

spec/system/authz_submitter_spec.rb
hectorcorrea: Ticketed it: https://github.com/pulibrary/pdc_describe/issues/492

Issue #488, opened 2022-10-13
Produce migration spec for CKavity Library: Next-Generation Sequencing


Twade968:
Following the example of [our bitKlavier rspec test ](https://github.com/pulibrary/pdc_describe/blob/main/spec/system/bitklavier_form_submission_spec.rb), produce system specs to migrate this object:

 [A library of novel genes with combinatorially diverse cavities, built on a stably folded structural template](https://dataspace.princeton.edu/handle/88435/dsp015999n626m)

**Acceptance Criteria**

- [ ] The Rspec test re-describes this work according to RDOS guidelines for DataCite description

PR #495, opened 2022-10-13
 CKavity Library

Twade968:
Fixes for ticket #488


Twade968:
redescribed from this data set: https://dataspace.princeton.edu/handle/88435/dsp015999n626m

```
<?xml version="1.0"?>
<resource xsi:schemaLocation='http://datacite.org/schema/kernel-4 http://schema.datacite.org/meta/kernel-4/metadata.xsd' xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance' xmlns='http://datacite.org/schema/kernel-4'>
  <identifier identifierType='DOI'>10.34770/gg40-tc15</identifier>
  <creators>
    <creator>
      <creatorName>Karas, Christina</creatorName>
      <givenName>Christina</givenName>
      <familyName>Karas</familyName>
    </creator>
    <creator>
      <creatorName>, National Science Foundation</creatorName>
      <givenName>National Science Foundation</givenName>
      <familyName></familyName>
    </creator>
  </creators>
  <titles>
    <title>
      CKavity Library: Next-Generation Sequencing / A library of novel genes
      with combinatorially diverse cavities, built on a stably folded structural
      template
    </title>
  </titles>
  <publisher>Princeton University Lewis-Sigler Institute</publisher>
  <resourceType resourceTypeGeneral='Dataset'/>
  <publicationYear>2020</publicationYear>
  <relatedIdentifiers>
    <relatedIdentifier relationType='IsIdenticalTo' relatedIdentifierType='ARK'>ark:/88435/dsp0159999n626m</relatedIdentifier>
  </relatedIdentifiers>
  <version>1</version>
  <rightsList>
    <rights rightsURI='https://creativecommons.org/licenses/by/4.0/' rightsIdentifier='CC BY'>Creative Commons Attribution 4.0 International</rights>
  </rightsList>
</resource>
```

bess (approved):

matthewjchandler:
@Twade968 there are several issues here. Let me preface everything by saying we (PRDS) did not curate this one, and I'm not sure which submission form was used for it.

I'm looking at the full item record to understand it better: https://dataspace.princeton.edu/handle/88435/dsp015999n626m?mode=full

The dc.contributor.author field was not used. Instead, there are three dc.creator fields in the original record that should be translated to the DataCite Creator field. One of the names listed as a creator was repeated as a contributor, and we do not want to replicate that error. In addition, we typically do not input funding agencies as contributors, so the Creator field you have for the National Science Foundation should be omitted.

The original record has a title as well as an alternative title, and the title field already includes a subtitle, so I do not think that the two title fields should be merged in DataCite. Instead, the alternative title from the original record can be translated as a second title with the type "AlternativeTitle" (see https://support.datacite.org/docs/datacite-metadata-schema-v44-mandatory-properties#3-title). 

For some reason, this record does not have an issue date (which is generally required). However, I can see that the dc.date.accessioned is in 2019, so the DataCite record should have the Publication Year field set to 2019 instead of 2020.

While the original record does have the publisher set to "Princeton University Lewis-Sigler Institute", our current practice is to correct such entries to "Princeton University".

The original record has two description fields filled: abstract and table of contents. My understanding is that for the purposes of migration, we are copying over abstracts as Description Type "Other" and omitting other description fields for now.

PR #504, opened 2022-10-17
Log changes

### LINKED ISSUE #198: Describe Form: Initial Messaging Feature & Provenance Log (concept to discuss)
## Description

(Note: from a discussion with @hectorcorrea) 

We may need to implement a simple messaging feature for the MVP while building towards the more involved feature we had discussed as a group. A more involved feature would allow for messaging to specific metadata fields or other comment areas of the form. It may also allow curator suggestions that a depositor can accept, reject, or further discuss.

Hector and I discussed having something similar to Github messaging (see image 1) that will allow the conversation between curator and depositor. Both curator and depositor should see the messaging. On the administrative side, we also want to see a provenance log for actions taken within the record. Github shows this with the conversation, but we might show it separately. (See image 2.) 

### Image 1 shows Github messaging and provenance for actions: 

Screen Shot 2022-06-30 at 5 10 11 PM

### Image 2 shows provenance for ScholarSphere (note: we imagine something similar for Describe admin; not public)

Screen Shot 2022-06-30 at 5 33 47 PM

## Acceptance Criteria
- [ ] Discuss concept with RDSS
- [ ] (if needed) Follow up with feedback from RDOS
- [ ] HH to make into actionable ticket(s)

hectorcorrea:
Keeps tracks of changes to the metadata (provenance log) in works. 

history

Handles the Provenance Log part of #198 for **metadata**. See issue #517 for file provenance log.

Closes #198

### DIFF
--- a/app/models/work_activity.rb
+++ b/app/models/work_activity.rb
@@ -47,52 +47,70 @@ class WorkActivity < ApplicationRecord
     end
   end
 
-  def comment_html
-    # convert user references to user links
-    text = message.gsub(USER_REFERENCE) do |at_uid|
-      uid = at_uid[1..-1]
-      user = User.where(uid: uid).first
-      user_info = user&.display_name_safe || uid
-      "<a class='comment-user-link' title='#{user_info}' href='{USER-PATH-PLACEHOLDER}/#{uid}'>#{at_uid}</a>"
-    end
-    # allow ``` for code blocks (Kramdown only supports ~~~)
-    text = text.gsub("```", "~~~")
-    Kramdown::Document.new(text).to_html
-  end
+  private
 
-  def changes_html
-    text = ""
-    json = JSON.parse(message)
-    json.keys.each do |key|
-      if json[key].is_a?(Array)
-        text += "<p><b>#{key}</b>:"
-        json[key].each do |value|
-          if value["action"] == "added"
-            text += <<-HTML
-              <i>(added)</i> <span>#{value["value"]}</span>
-            HTML
-          elsif value["action"] == "removed"
-            text += <<-HTML
-              <i>(removed)</i> <span>#{value["value"]}</span>
-            HTML
-          else
-            text += <<-HTML
-              <span style="text-decoration: line-through;">#{value["from"]}</span>
-              <span>#{value["to"]}</span>
-            HTML
+    def comment_html
+      # convert user references to user links
+      text = message.gsub(USER_REFERENCE) do |at_uid|
+        uid = at_uid[1..-1]
+        user = User.where(uid: uid).first
+        user_info = user&.display_name_safe || uid
+        "<a class='comment-user-link' title='#{user_info}' href='{USER-PATH-PLACEHOLDER}/#{uid}'>#{at_uid}</a>"
+      end
+      # allow ``` for code blocks (Kramdown only supports ~~~)
+      text = text.gsub("```", "~~~")
+      Kramdown::Document.new(text).to_html
+    end
+
+    # Returns the message formatted to display changes that were logged
+    def changes_html
+      text = ""
+      json = JSON.parse(message)
+      json.keys.each do |key|
+        if json[key].is_a?(Array)
+          # Multi-value change logged, process each individual entry
+          text += "<p><b>#{key}</b>:"
+          json[key].each do |value|
+            text += change_value_html(value)
           end
+          text += "</p>"
+        else
+          # Single-value change logged
+          value = json[key]
+          text += change_value_html(value)
         end
-        text += "</p>"
+      end
+      text
+    end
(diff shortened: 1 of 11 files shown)


app/models/work_activity.rb
```diff
@@ -40,15 +40,78 @@ def created_by_user
...
+
+    # Returns the message formatted to display changes that were logged as an activity
+    def changes_html
+      text = ""
+      changes = JSON.parse(message)
+      changes.keys.each do |field|
+        if changes[field].is_a?(Array)
+          # Multi-value change, process each individual entry
+          values = changes[field].map { |value| change_value_html(value) }
+          text += "<p><b>#{field}</b>:#{values.join}</p>"
```
carolyncole, line 73: @hectorcorrea was there intention to make the formatting different for a multi-value?  It might be nice to move the formatting out of the if, but adding a space would match it up.
```suggestion
          text += "<p><b>#{field}</b>: #{values.join}</p>"
```

app/models/work_activity.rb
```diff
@@ -40,15 +40,78 @@ def created_by_user
...
+      HTML
+    end
+
+    def change_set_html(from, to)
+      if from.blank? && to.present?
+        # value was set
+        "<span>#{to}</span><br/>"
+      elsif from.present? && to.present?
+        # value was changed
+        "<span>#{to}</span><br/>"
```
carolyncole, line 111: Should this be different than line 108?

app/services/form_to_resource_service.rb
```diff
@@ -14,7 +14,9 @@ def convert(params, work, current_user)
       resource.publisher = params["publisher"] if params["publisher"].present?
       resource.publication_year = params["publication_year"] if params["publication_year"].present?
       resource.rights = PDCMetadata::Rights.find(params["rights_identifier"])
-      resource.keywords = (params["keywords"] || "").split(",")
+      resource.keywords = (params["keywords"] || "").split(",").map(&:strip)
+      resource.resource_type = params["resource_type"]
```
carolyncole, line 18: This is in process_curator_controlled now, https://github.com/pulibrary/pdc_describe/blob/53e9f0d41bbb95cb6707737a20a7d75d5b6c6fe1/app/services/form_to_resource_service.rb#L42-L43

app/services/resource_compare_service.rb
```diff
@@ -0,0 +1,218 @@
+# frozen_string_literal: true
+
+# Compares two PDCMetadata::Resource objects and provides a hash with the `differences`
+#
+# If there are no differences between the two objects `identical?` returns `true` and `differences == {}`
+# If there are differences the `differences` hash has the following structure:
+#
+# ```
+#    :field_name = {action:, from:, to:, value: }     for single-value fields (e.g. description)
```
carolyncole, line 9: I was wondering if it might make life simpler to just return an array no matter what. This is just a pondering based on having to do a if above.

app/services/resource_compare_service.rb
```diff
@@ -0,0 +1,218 @@
...
+    compare_resources
+  end
+
+  def identical?
+    @differences == {}
+  end
+
+  private
+
+    def compare_resources
```
carolyncole, line 32: I really liked how you had compare_value in creator.  I wonder if you added a compare_value to the Resource types if a large quantity of this comparison could be simplified.  I expected you to diff the compare_value in this routine when I saw it.

carolyncole (changes requested):
@hectorcorrea only one definite change needed.  The others are more ideas...

app/services/form_to_resource_service.rb
hectorcorrea: Good catch, thanks!

app/models/work_activity.rb
hectorcorrea: I wanted to highlight that these are different situations (value set for the first time vs value change) in case we want to display that kind of change different to the user. At this point we display it the same, but this makes it easier for somebody to change it if we want something different once users see the log of changes with real data.

app/services/resource_compare_service.rb
hectorcorrea: This was a great insight...I ended changing the code to always return an array and it looks much better and simpler.

app/models/work_activity.rb
hectorcorrea: This has now been simplified now that we always return an array. Again, thanks for the hint.

app/services/resource_compare_service.rb
hectorcorrea: I am still on the fence on what will be best for this in the long term. But certainly a good idea to keep in mind as we add more (complex) properties to resource.

carolyncole (approved):

Issue #505, opened 2022-10-17
Error while editing an approved work


carolyncole:
I came across this error when adding some tests for the edit of  an approved work.  I will be putting a pending test into a PR soon...

[pdc_describe/staging] NoMethodError: undefined method `id' for #<S3File:0x00007fe4e8789a18 @filename="10.80021/sxes-6k67/84/2022-09-05_09-31-44_834.heic", @last_modified=2022-09-29 18:28:26 UTC, @size=2746794, @checksum="59c4a9c70640148e7c8273140f5e6bac", @query_service=#<S3QueryService:0x00007fe4e8740048 @model=#<Work id: 84, work_type: "DATASET", state: "approved", collection_id: 1, created_by_user_id: 14, created_at: "2022-09-09 17:13:10.759490000 +0000", updated_at: "2022-09-29 18:28:25.950793000 +0...

## Backtrace

    line 243 of [PROJECT_ROOT]/app/models/work.rb: block in uploads_attributes
    line 241 of [PROJECT_ROOT]/app/models/work.rb: map
    line 241 of [PROJECT_ROOT]/app/models/work.rb: uploads_attributes

## Root Cause
The root cause of this issue is that we are showing uploads [here](https://github.com/pulibrary/pdc_describe/blob/main/app/views/works/_form.html.erb#L44-L49).  Not pre-curation or post-curation, but just uploads.  Post-curation files are different from pre-curation files now and can (and should not) be displayed in the same way. 

We could easily solve the bug by not displaying the files on the edit form for approved works.  What should we really do though?  I feel like displaying the files (non editable) for curators, and maybe linking to Globus if they want to modify them seems like a reasonable first step.

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/99214/faults/89273390)

Issue #255, opened 2022-08-03
When the Dataset is approved make the dataset metadata readonly


carolyncole:
The metadata can not be changed by the depositor.  The curators will still have access to modify the metadata.

Acceptance:

- [ ] Metadata can not be change by the depositor after the dataset is approved
- [ ] Metadata can be changed by the depositor before the dataset is approved
- [ ] Metadata can be changed by the curator before and after the dataset is approved

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-Workcycles-61a4f1a12a399b001730f65a/issues/pulibrary/pdc_describe/255?planning-poker) with ZenHub @bess @jrgriffiniii

PR #506, opened 2022-10-17
Removing Edit button from show of approved work

carolyncole:
Only curators and super admins should have edit capabilities on approved works. There is a pending test, because there is a bug on the edit screen for approved works (#505).

refs #255

### DIFF
--- a/app/views/works/show.html.erb
+++ b/app/views/works/show.html.erb
@@ -40,7 +40,9 @@
 </style>
 
 <div>
-  <%= link_to("Edit", edit_work_path(@work, wizard: @work.draft? ? true : nil), class: "btn btn-primary") %>
+  <% if @work.draft?  || @work.awaiting_approval? || current_user.has_role?(:collection_admin, @work.collection) %>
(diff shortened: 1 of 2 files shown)


app/views/works/show.html.erb
```diff
@@ -40,7 +40,9 @@
 </style>
 
 <div>
-  <%= link_to("Edit", edit_work_path(@work, wizard: @work.draft? ? true : nil), class: "btn btn-primary") %>
+  <% if @work.draft?  || @work.awaiting_approval? || current_user.has_role?(:collection_admin, @work.collection) %>
+    <%= link_to("Edit", edit_work_path(@work, wizard: @work.draft? ? true : nil), class: "btn btn-primary") %>
```
mccalluc, line 44: Not a change in this PR, but would plain `@work.draft?` be sufficient here? or is it important to return `nil` instead of `false`? Just trying to get familiar with this codebase.
```suggestion
    <%= link_to("Edit", edit_work_path(@work, wizard: @work.draft?), class: "btn btn-primary") %>
```

carolyncole: @mccalluc interesting question.  Since I just moved the code I do not know.  That said I could try it and see if all the tests still pass, but I would not want to make it part of this PR.

PR #518, opened 2022-10-18
Allow resource type to change on assignment

### LINKED ISSUE #502: Edit of resource_type and resourceTypeGeneral
The value entered by the user for `resource_type` and `resourceTypeGeneral` is not being saved to the database nor is it honored when rendering the edit form.

carolyncole:
Allow the resource type to change when the entire resource is assigned. fixes #502

The resource type and resource type general were not connected back through to the model at any point in the edit cycle.


hectorcorrea (approved):
Thank you!

PR #523, opened 2022-10-19
Allow user to toggle display of change history

hectorcorrea:
Tiny enhancement to the change log display

# Default display (comments + change history)
show

# Without change history
show-me-not


hectorcorrea:
@carolyncole I had the same thought, 6 of one 1/2 dozen of another :) -- but as you said, this will be an easy change later if that's what users prefer.

Issue #494, opened 2022-10-13
Describe: Related Identifier Property Set


hhkitsune:
# Description

This ticket shows the [12 RelatedIdentifer](https://support.datacite.org/docs/datacite-metadata-schema-v44-recommended-and-optional-properties#12-relatedidentifier) property set that we would like to start with for the MVP. Note: There are other properties in this set, but this is what we have decided to prioritize for the MVP.

## Datacite properties to include in this set:

- Related Identifier (Top, use different label; see slides)
- Related Identifier Type (subproperty; mandatory if RelatedIdentifier is used)
- Relation Type (subproperty; mandatory if RelatedIdentifier is used)

### Image shows wireframe for this property set

![Screen Shot 2022-10-11 at 11 49 24 AM](https://user-images.githubusercontent.com/50670293/195684368-0ce77542-e9aa-44a1-9866-cca8e3ccd844.jpeg)

# Acceptance Criteria

- [x] Set name text displays above the set: Related object (i.e., published dependencies)
- [x] Set follows the order described in [RDOS_publishing ticket 7](https://github.com/pulibrary/RDOS_Publishing/issues/7)
- [x] Set includes three properties: Related Identifier, Related Identifier Type, Relation Type that correspond to Datacite 4.4 (12) RelatedIdentifier, (12.a) relatedIdentifierType, and (12.b) relationType.
- [ ] (12.a) relatedIdentifierType and (12.b) relationType are mandatory when (12) RelatedIdentifier is used. However, it is not mandatory to use (12) RelatedIdentifier.
- [x] The DataCite controlled vocabulary is used to populate values for relatedIdentifierType and relationType
- [x] The DataCite record produced needs to include the human-friendly version of any controlled vocabularies. E.g., use "arXiv", not "ARXIV"

## Reference
- [(12) RelatedIdentifier](https://support.datacite.org/docs/datacite-metadata-schema-v44-recommended-and-optional-properties#12-relatedidentifier)
- [(12.a) relatedIdentifierType](https://support.datacite.org/docs/datacite-metadata-schema-v44-recommended-and-optional-properties#12a-relatedidentifiertype)
- [(12.b) relationType](https://support.datacite.org/docs/datacite-metadata-schema-v44-recommended-and-optional-properties#12b-relationtype)

bess:
Added last acceptance criteria, plus one more, to a different ticket: https://github.com/pulibrary/pdc_describe/issues/591

PR #528, opened 2022-10-19
Implement select menus for Related Objects

bess:
Work toward #494 

Screen Shot 2022-10-19 at 4 42 21 PM
Screen Shot 2022-10-19 at 4 42 03 PM

### DIFF
--- a/app/models/pdc_metadata/related_object.rb
+++ b/app/models/pdc_metadata/related_object.rb
@@ -20,5 +20,21 @@ module PDCMetadata
     def self.new_related_object(related_identifier, related_identifier_type, relation_type)
       RelatedObject.new(related_identifier: related_identifier, related_identifier_type: related_identifier_type, relation_type: relation_type)
     end
+
+    ##
+    # Generate a list of valid options for the related_identifier_type field
+    def self.related_identifier_type_options
+      pairs = Datacite::Mapping::RelatedIdentifierType.to_a.map { |value| [value.key, value.value] }
+      built = Hash[pairs]
+      built.with_indifferent_access
+    end
+
+    ##
+    # Generate a list of valid options for the relation_type field
+    def self.relation_type_options
+      pairs = Datacite::Mapping::RelationType.to_a.map { |value| [value.key, value.value] }
+      built = Hash[pairs]
+      built.with_indifferent_access
+    end
   end
 end
--- a/app/views/works/_edit_javascript.html.erb
+++ b/app/views/works/_edit_javascript.html.erb
@@ -107,7 +107,7 @@
     }
 
     var relatedIdentififerTypeSelectHtml = function(relatedIdentifierTypeId, related_identifier_type) {
-      var i, html, related_identifier_type_options, key, value;
+      var i, html, key, value;
       types = relatedIdentifierTypes();
       html = `<select id="${relatedIdentifierTypeId}" name="${relatedIdentifierTypeId}">`;
       for(i = 0; i < types.length; i++) {
@@ -121,7 +121,7 @@
     }
 
     var relationTypeSelectHtml = function(relationTypeId, relation_type) {
-      var i, html, related_identifier_type_options, key, value;
+      var i, html, key, value;
       types = relationTypes();
       html = `<select id="${relationTypeId}" name="${relationTypeId}">`;
       for(i = 0; i < types.length; i++) {
--- a/app/views/works/_edit_javascript.html.erb
+++ b/app/views/works/_edit_javascript.html.erb
@@ -477,7 +477,7 @@
       }
     }
 
-    // Load any existing related objects into the edit form. 
+    // Load any existing related objects into the edit form.
     // If there are any related objects they should appear in hidden <span> tags.
     if ($(".related-object-data").length == 0) {
       // Add an empty related object for the user to fill it out
--- a/app/views/works/_edit_javascript.html.erb
+++ b/app/views/works/_edit_javascript.html.erb
@@ -81,6 +81,7 @@
       return html;
     }
 
+    // ************************************************ //
     // Related Objects
     // related_identifier:, related_identifier_type:, relation_type
     var addRelatedObjectHtml = function(num, related_identifier, related_identifier_type, relation_type) {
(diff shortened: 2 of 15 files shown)


app/models/pdc_metadata/related_object.rb
```diff
@@ -20,5 +20,21 @@ def value
     def self.new_related_object(related_identifier, related_identifier_type, relation_type)
       RelatedObject.new(related_identifier: related_identifier, related_identifier_type: related_identifier_type, relation_type: relation_type)
     end
+
+    ##
```
mccalluc, line 24: Is there a style guide that suggests the `##`? A one-line comment is probably enough.

app/views/works/_edit_javascript.html.erb
```diff
@@ -81,30 +81,122 @@
...
         <td>
-          <input type="text" id="${relationTypeId}" name="${relationTypeId}" value="${relation_type}" />
+          ${relationTypeHtml}
         </td>
       </tr>`;
       $("#related-objects-table").append(rowHtml);
     }
 
+    var relatedIdentififerTypeSelectHtml = function(relatedIdentifierTypeId, related_identifier_type) {
+      var i, html, related_identifier_type_options, key, value;
```
mccalluc, line 110: Unless we're expecting to mutate these, using `const` can make sure we don't do it by mistake.

app/views/works/_edit_javascript.html.erb
```diff
@@ -81,30 +81,122 @@
...
+      types = relatedIdentifierTypes();
+      html = `<select id="${relatedIdentifierTypeId}" name="${relatedIdentifierTypeId}">`;
+      for(i = 0; i < types.length; i++) {
+        key = types[i].key;
+        value = types[i].value;
+        selected = related_identifier_type == key ? "selected" : "";
+        html += `<option value="${key}" ${selected}>${value}</option>`;
+      }
```
mccalluc, line 118:
```suggestion
      html += types.map((({key, value}) => {
        const selected = related_identifier_type == key ? "selected" : "";
        return `<option value="${key}" ${selected}>${value}</option>`;
      }).join('')
```

Using `map` may be more concise than a `for` loop.

app/views/works/_edit_javascript.html.erb
mccalluc: is `related_identifier_type_options` used?

app/views/works/_edit_javascript.html.erb
```diff
@@ -81,30 +81,122 @@
...
-          <input type="text" id="${relationTypeId}" name="${relationTypeId}" value="${relation_type}" />
+          ${relationTypeHtml}
         </td>
       </tr>`;
       $("#related-objects-table").append(rowHtml);
     }
 
+    var relatedIdentififerTypeSelectHtml = function(relatedIdentifierTypeId, related_identifier_type) {
+      var i, html, related_identifier_type_options, key, value;
+      types = relatedIdentifierTypes();
```
mccalluc, line 111:
```suggestion
      const types = relatedIdentifierTypes();
```

app/views/works/_edit_javascript.html.erb
```diff
@@ -81,30 +81,122 @@
...
+        selected = related_identifier_type == key ? "selected" : "";
+        html += `<option value="${key}" ${selected}>${value}</option>`;
+      }
+      html += `</select>`;
+      return html;
+    }
+
+    var relationTypeSelectHtml = function(relationTypeId, relation_type) {
+      var i, html, related_identifier_type_options, key, value;
+      types = relationTypes();
```
mccalluc, line 125: Same comments as above. This is almost the same as the function above: Maybe pass `types` as a third parameter?

app/views/works/_edit_javascript.html.erb
```diff
@@ -81,30 +81,122 @@
...
+      types.push({key: "IS_REVIEWED_BY", value: "IsReviewedBy"})
+      types.push({key: "REVIEWS", value: "Reviews"})
+      types.push({key: "IS_DERIVED_FROM", value: "IsDerivedFrom"})
+      types.push({key: "IS_SOURCE_OF", value: "IsSourceOf"})
+      types.push({key: "IS_OBSOLETED_BY", value: "IsObsoletedBy"})
+      types.push({key: "OBSOLETES", value: "Obsoletes"})
+      types.push({key: "IS_DESCRIBED_BY", value: "IsDescribedBy"})
+      types.push({key: "DESCRIBES", value: "Describes"})
+      types.push({key: "IS_REQUIRED_BY", value: "IsRequiredBy"})
+      types.push({key: "REQUIRES", value: "Requires"})
```
mccalluc, line 171: Why does this need to be hardcoded? Can't `Datacite::Mapping::RelationType` be used here, too?

app/views/works/_edit_javascript.html.erb
```diff
@@ -81,30 +81,122 @@
...
+      types.push({key: "ISTC", value: "ISTC"})
+      types.push({key: "LISSN", value: "LISSN"})
+      types.push({key: "LSID", value: "LSID"})
+      types.push({key: "PMID", value: "PMID"})
+      types.push({key: "PURL", value: "PURL"})
+      types.push({key: "UPC", value: "UPC"})
+      types.push({key: "URL", value: "URL"})
+      types.push({key: "URN", value: "URN"})
+      types.push({key: "IGSN", value: "IGSN"})
+      types.push({key: "W3ID", value: "w3id"})
```
mccalluc, line 195: ditto?

spec/fixtures/cytoskeletal_metadata.json
```diff
@@ -150,8 +150,8 @@
     "related_objects": [
         {
             "related_identifier": "https://www.biorxiv.org/content/10.1101/545517v1",
-            "related_identifier_type": "arXiv",
-            "relation_type": "isCitedBy"
+            "related_identifier_type": "ARXIV",
+            "relation_type": "IS_CITED_BY"
```
mccalluc, line 154: This surprises me, a little: I would think the all-caps versions are just internal keys generated for the ruby library. I may be thinking about it wrong.

spec/system/work_edit_spec.rb
```diff
@@ -93,8 +93,8 @@
       visit edit_work_path(work)
       click_on "Additional Metadata"
       expect(page.find("#related_identifier_1").value).to eq "https://www.biorxiv.org/content/10.1101/545517v1"
-      expect(page.find("#related_identifier_type_1").value).to have_content "arXiv"
-      expect(page.find("#relation_type_1").value).to have_content "isCitedBy"
+      expect(page.find("#related_identifier_type_1").value).to eq "ARXIV"
```
mccalluc, line 96: yeah, if this is coming through in the html, I'd think we'd prefer "arXiv" to "ARXIV"?

mccalluc (changes requested):
- Some stylistic suggestions in the JS: If we'd like to stay with the status-quo style, let me know and I'll stop mentioning these.
- Question about comment conventions.
- unused JS variable. Are you using ESLint?
- Perhaps combine functions?
- Not sure why lists are hard-coded.
- Capitalization?

mccalluc (approved):
We identified refactorings that can happen down the road:
- #531
- #533
- #534

PR #530, opened 2022-10-19
Add net-smtp

### LINKED ISSUE #525: Add net-smtp to Gemfile
I got this warning on startup:
> You don't have net-smtp installed in your application. Please add it to your Gemfile and run bundle install

Carolyn thinks other folks haven't hit it because
> I bet no one noticed it missing in pdc_describe because we mostly installed pdc_discovery first https://github.com/pulibrary/pdc_discovery#dependencies

I should be able to do this.

mccalluc:
- Fix #525

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,6 +52,8 @@ gem "datacite", github: "sul-dlss/datacite-ruby", branch: "main"
 
 gem "kramdown"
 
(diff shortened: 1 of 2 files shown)


carolyncole (approved):
Makes @mccalluc machine work, and causes no harm on staging

PR #544, opened 2022-10-21
fix js build for pulldown

### LINKED ISSUE #540: "Uncaught SyntaxError: import declarations may only appear at top level of a module"
I've just deployed v0.32.0 to staging. I logged in, and I'm trying to click on the button with my NetID, but no pulldown appears: Instead in the console I see this error:
> Uncaught SyntaxError: import declarations may only appear at top level of a module

(This is not a recipe for a reproducer... I'm not sure exactly what I did before I got here.)

The link to the bad JS is https://vpn.princeton.edu/https/pdc-describe-staging.princeton.edu/describe/vite/assets/application.4e0043f6.js

```
pan_eval((function(){
import"./jquery-ui-triggeredAutocomplete.1ed59131.js";....................
```
(following code truncated)

It seems that the asset bundler isn't handling ES Modules? I'm not sure if this code changed recently.

mccalluc:
- Eventually, fix #540

but right now this doesn't work. (Errors with `Uncaught ReferenceError: jquery is not defined`.) Filing as draft, just so the branch doesn't get lost.

The most useful thing would probably be figuring out how to replicate the production build locally.
I've put more info on the history of the bug in the issue.


mccalluc:
Still not working: 
> Uncaught Error: rails-ujs has already been loaded!

mccalluc:
Closing. We realized that this error was caused by using the VPN with a prefixed URL. The issue should remain open, but it's less pressing.

PR #546, opened 2022-10-24
Updating the edit form to only show upload edits before approval

### LINKED ISSUE #505: Error while editing an approved work

carolyncole:
Only show the ability to edit the uploads before the Work is approved, since after it has been approved we drive everything off of S3 

Also changed test for edit of an approved work to administered by so that as an administrator I could edit my own work.

fixes #505

### DIFF
--- a/app/views/works/_form.html.erb
+++ b/app/views/works/_form.html.erb
@@ -44,7 +44,14 @@
       <% if !@wizard_mode %>
         <hr />
         <div class="deposit-uploads">
-          <%= render(partial: 'works/uploads_form', locals: { form: form }) %>
+          <% if @work.approved? %>
+            <section class="uploads">
+              <h2 class="h2"><%= t('works.uploads.post_curation.heading') %></h2>
+              <%= render(partial: 'works/uploads/post_curation_s3_resources') %>
+            </section>
+         <% else %>
+            <%= render(partial: 'works/uploads_form', locals: { form: form }) %>
+          <% end %>
         </div>
       <% end %>
 
--- a/spec/system/view_data_in_s3_spec.rb
+++ b/spec/system/view_data_in_s3_spec.rb
@@ -73,7 +73,7 @@ RSpec.describe "View status of data in S3", mock_ezid_api: true do
 
       context "when user is a curator" do
         let(:user) { FactoryBot.create(:research_data_moderator) }
-        pending "shows data from S3" do
+        it "shows data from S3" do
           stub_s3(data: s3_data)
           visit work_path(work)
           expect(page).to have_content file1.filename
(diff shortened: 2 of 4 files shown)


app/views/works/_form.html.erb
```diff
@@ -44,7 +44,14 @@
       <% if !@wizard_mode %>
         <hr />
         <div class="deposit-uploads">
-          <%= render(partial: 'works/uploads_form', locals: { form: form }) %>
+          <% if @work.approved? %>
+            <section class="uploads">
+              <h2 class="h2"><%= t('works.uploads.post_curation.heading') %></h2>
+              <%= render(partial: 'works/uploads/post_curation_s3_resources') %>
+            </section>
+         <% else %>
```
mccalluc, line 52:
```suggestion
          <% else %>
```

app/views/works/_form.html.erb
```diff
@@ -44,7 +44,14 @@
       <% if !@wizard_mode %>
         <hr />
         <div class="deposit-uploads">
-          <%= render(partial: 'works/uploads_form', locals: { form: form }) %>
+          <% if @work.approved? %>
+            <section class="uploads">
+              <h2 class="h2"><%= t('works.uploads.post_curation.heading') %></h2>
+              <%= render(partial: 'works/uploads/post_curation_s3_resources') %>
```
mccalluc, line 50: I think this is the only place this partial is used: Maybe move the h2 into it for consistency?

spec/system/view_data_in_s3_spec.rb
```diff
@@ -83,6 +83,20 @@
           expect(page).to have_content file1.filename
           expect(page).to have_content file2.filename
         end
+
+        context "when user is a super super_admin_user" do
```
mccalluc, line 87:
```suggestion
        context "when user is a super_admin_user" do
```

Probably accidental repetition? Though if there are super-super users, that would be good to know.

mccalluc (approved):

app/views/works/_form.html.erb
carolyncole: This view is utilized a couple places.  I grabbed it from the show.

PR #547, opened 2022-10-24
Use raw() to inject JSON

### LINKED ISSUE #533: Refactor the way we handle controlled vocabularies in js
Currently, we re-create some of the datacite controlled vocabularies in our js files. It would be better to rely on the vocabulary lists produced by the `Datacite::Mapping` gem. 

Can we replace structures like this:
```
    var relatedIdentifierTypes = function() {
      var types = [];
      types.push({key: "ARK", value: "ARK"})
      types.push({key: "ARXIV", value: "arXiv"})
      types.push({key: "BIBCODE", value: "bibcode"})
      types.push({key: "DOI", value: "DOI"})
      types.push({key: "EAN13", value: "EAN13"})
      types.push({key: "EISSN", value: "EISSN"})
      types.push({key: "HANDLE", value: "Handle"})
      types.push({key: "ISBN", value: "ISBN"})
      types.push({key: "ISSN", value: "ISSN"})
      types.push({key: "ISTC", value: "ISTC"})
      types.push({key: "LISSN", value: "LISSN"})
      types.push({key: "LSID", value: "LSID"})
      types.push({key: "PMID", value: "PMID"})
      types.push({key: "PURL", value: "PURL"})
      types.push({key: "UPC", value: "UPC"})
      types.push({key: "URL", value: "URL"})
      types.push({key: "URN", value: "URN"})
      types.push({key: "IGSN", value: "IGSN"})
      types.push({key: "W3ID", value: "w3id"})
      return types;
    }
```
With a ruby function that creates the needed JSON and includes it where needed? 

### Acceptance criteria
- [ ] More than one team member should pair on this

mccalluc:
- Fix #533

On the issue this was identified as something to pair up for... so I shouldn't have rushed ahead with this. Filing as draft, until there's a good time for a screen share. There are other ways this could be done... for instance, passing in the data structure, and then iterating over it in ERB to build the JSON.

### DIFF
--- a/app/views/works/_edit_javascript.html.erb
+++ b/app/views/works/_edit_javascript.html.erb
@@ -135,65 +135,11 @@
     }
 
     var relationTypes = function() {
-      var types = [];
-      types.push({key: "IS_CITED_BY", value: "IsCitedBy"})
-      types.push({key: "CITES", value: "Cites"})
-      types.push({key: "IS_SUPPLEMENT_TO", value: "IsSupplementTo"})
-      types.push({key: "IS_SUPPLEMENTED_BY", value: "IsSupplementedBy"})
-      types.push({key: "IS_CONTINUED_BY", value: "IsContinuedBy"})
-      types.push({key: "CONTINUES", value: "Continues"})
-      types.push({key: "HAS_METADATA", value: "HasMetadata"})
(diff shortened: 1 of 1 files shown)


mccalluc:
oh ... in this same file there is also a list of contributor roles: That can be cleaned up when we pair.

hectorcorrea (approved):

PR #548, opened 2022-10-24
Simplify select menu JS

### LINKED ISSUE #531: Refactor the way we do controlled vocabulary drop down menus
Currently, we are using js methods like this to generate drop down menus. We're using this pattern in several places, and it would be more concise and maintainable to refactor these into a single method that gets a vocabulary object passed to it. 

```
    var relatedIdentififerTypeSelectHtml = function(relatedIdentifierTypeId, related_identifier_type) {
      var i, html, key, value;
      const types = relatedIdentifierTypes();
      html = `<select id="${relatedIdentifierTypeId}" name="${relatedIdentifierTypeId}">`;
      for(i = 0; i < types.length; i++) {
        key = types[i].key;
        value = types[i].value;
        selected = related_identifier_type == key ? "selected" : "";
        html += `<option value="${key}" ${selected}>${value}</option>`;
      }
      html += `</select>`;
      return html;
    }

    var relationTypeSelectHtml = function(relationTypeId, relation_type) {
      var i, html, key, value;
      const types = relationTypes();
      html = `<select id="${relationTypeId}" name="${relationTypeId}">`;
      for(i = 0; i < types.length; i++) {
        key = types[i].key;
        value = types[i].value;
        selected = relation_type == key ? "selected" : "";
        html += `<option value="${key}" ${selected}>${value}</option>`;
      }
      html += `</select>`;
      return html;
    }
```

### Acceptance criteria
- [ ] Two or more members of the team should pair on this

mccalluc:
- Fix #531

### DIFF
--- a/app/views/works/_edit_javascript.html.erb
+++ b/app/views/works/_edit_javascript.html.erb
@@ -68,16 +68,12 @@
     }
 
     function makeSelectHtml(selectId, currentKey, kvList) {
-      var i, html, roles, key, value;
       html = `<select id="${selectId}" name="${selectId}">`;
-      for(i = 0; i < kvList.length; i++) {
-        key = kvList[i].key;
-        value = kvList[i].value;
-        selected = currentKey == key ? "selected" : "";
-        html += `<option value="${key}" ${selected}>${value}</option>`;
-      }
-      html += `</select>`;
-      return html;
+      const options = kvList.map(
+        ({key, value}) => 
+        `<option value="${key}" ${currentKey == key ? "selected" : ""}>${value}</option>`
+      );
+      return `<select id="${selectId}" name="${selectId}">${options}</select>`
     }
 
     // ************************************************ //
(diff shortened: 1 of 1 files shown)


app/views/works/_edit_javascript.html.erb
```diff
@@ -67,18 +67,13 @@
...
-      for(i = 0; i < roles.length; i++) {
-        key = roles[i].key;
-        value = roles[i].value;
-        selected = activeRole == key ? "selected" : "";
-        html += `<option value="${key}" ${selected}>${value}</option>`;
-      }
-      html += `</select>`;
-      return html;
+    function makeSelectHtml(selectId, currentKey, kvList) {
+      html = `<select id="${selectId}" name="${selectId}">`;
```
hectorcorrea, line 71: This line can be removed, right?

hectorcorrea:
Love this, great job simplifying the code!

hectorcorrea (dismissed):

hectorcorrea (approved):

Issue #517, opened 2022-10-18
Provenance Log for files in Work


hectorcorrea:
In addition to handle Provenance Log for metadata (issue #198) we should also log somehow files added and removed (and perhaps updated?). 

I am spinning this as a separate issue from #198 since files can be added/removed/updated outside of PDC Describe (e.g. a user going directly to Globus/AWS) and therefore we'll need a different mechanism to detect the changes if we want to implement this feature. 

We could still use the same mechanism to log the changes as we do for metadata (i.e save the changes in the `WorkActivity` table) and render them on the Show page as we do for metadata provenance.

PR #549, opened 2022-10-24
Log file changes

hectorcorrea:
Logs file changes (add/delete) when done via the Work Edit page.

Screen Shot 2022-10-24 at 2 17 51 PM

Works towards #517

The bulk of the changes in this PR are on the `WorkUploadsEditService` class. I made the methods in this class instance methods (rather than class methods) so that we can keep track of the changes. Notice the new calls to `track_change(...)` and `work.log_file_changes(@changes, @current_user)`. The rest of the logic stayed the same.

### DIFF
--- a/app/models/work_activity.rb
+++ b/app/models/work_activity.rb
@@ -41,7 +41,7 @@ class WorkActivity < ApplicationRecord
 
   def message_html
     if activity_type == "CHANGES"
-      changes_html
+      metadata_changes_html
     elsif activity_type == "FILE-CHANGES"
       file_changes_html
     else
@@ -74,7 +74,7 @@ class WorkActivity < ApplicationRecord
     end
 
     # Returns the message formatted to display _metadata_ changes that were logged as an activity
-    def changes_html
+    def metadata_changes_html
       text = ""
       changes = JSON.parse(message)
       changes.keys.each do |field|
--- a/app/models/work_activity.rb
+++ b/app/models/work_activity.rb
@@ -42,6 +42,8 @@ class WorkActivity < ApplicationRecord
   def message_html
     if activity_type == "CHANGES"
       changes_html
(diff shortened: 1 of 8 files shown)


app/models/work_activity.rb
```diff
@@ -62,8 +64,17 @@ def comment_html
...
 
-    # Returns the message formatted to display changes that were logged as an activity
-    def changes_html
+    # Returns the message formatted to display _file_ changes that were logged as an activity
+    def file_changes_html
+      changes = JSON.parse(message)
+      changes_html = changes.map do |change|
+        "<i>#{change['action']}</i> <span>#{change['filename']}<span><br/>"
+      end
+      "<p><b>Files</b>: #{changes_html.join}</p>"
```
carolyncole, line 73: I wonder if making the changes something like a data table would be better for screen readers.  I know paragraphs of text are not the greatest.  This should not block this PR, but something to think about.

carolyncole (changes requested):
@hectorcorrea looks great, but  I believe that you can fix the code up instead of just ignoring the rubocop error.  That method has gotten large and complex.  It would be nice to see each part of the if then else block be named and put into their own method.  That will make our lives easier when we go to understand and maintain the code.  This is a small enough PR that making some code improvements will not make it unreadable.

app/models/work_activity.rb
hectorcorrea: Good idea, I added a separate ticket for that https://github.com/pulibrary/pdc_describe/issues/553

hectorcorrea:
@carolyncole I refactored the code into its own method so that we reduce the complexity of the original and Rubocop is happy: https://github.com/pulibrary/pdc_describe/pull/549/commits/6f7a2aade39b17553aefc71d35a6369ca0319b8e

carolyncole (approved):

PR #550, opened 2022-10-25
Add notification to curators when an item is marked complete

### LINKED ISSUE #505: Error while editing an approved work

carolyncole:
fixes #505


app/models/work.rb
```diff
@@ -723,5 +723,11 @@ def publish_precurated_files
 
       delete_pre_curation_s3_dir
     end
+
+    def notify_collection_curators(current_user)
+      curators = collection.administrators.map { |admin| "@#{admin.uid}" }.join(", ")
+      notification = "#{curators} The [work](#{work_path(self)}) is ready for review."
```
mccalluc, line 729: I'm curious whether this is a full url or just a path... but that's something for me to track down.

spec/models/work_spec.rb
```diff
@@ -548,6 +548,15 @@
...
       expect { awaiting_approval_work.draft!(user) }.to raise_error AASM::InvalidTransition
     end
+
+    it "notifies the curators it is ready for review" do
+      curator = FactoryBot.create(:research_data_moderator)
+      expect { awaiting_approval_work }.to change { WorkActivity.where(activity_type: "SYSTEM").count }.by(2)
+      expect(WorkActivity.where(activity_type: "SYSTEM").last.message).to eq("marked as awaiting_approval")
+      curator_notification = WorkActivity.where(activity_type: "SYSTEM").first.message
+      expect(curator_notification).to include("@#{curator.uid}")
+      expect(curator_notification). to include(Rails.application.routes.url_helpers.work_path(awaiting_approval_work))
```
mccalluc, line 558:
```suggestion
      expect(curator_notification).to include(Rails.application.routes.url_helpers.work_path(awaiting_approval_work))
```

mccalluc (approved):

mccalluc (approved):

app/models/work.rb
carolyncole: Just a relative path. I see your point, a path might not be helpful in an email.

app/models/work.rb
```diff
@@ -728,5 +728,11 @@ def publish_precurated_files
 
       delete_pre_curation_s3_dir
     end
+
+    def notify_collection_curators(current_user)
+      curators = collection.administrators.map { |admin| "@#{admin.uid}" }.join(", ")
+      notification = "#{curators} The [work](#{work_url(self)}) is ready for review."
```
carolyncole, line 734: @mccalluc change this to the url so it will be more helpful in an email.

mccalluc (approved):

PR #554, opened 2022-10-25
Remove ERB wrapper around JS

mccalluc:
- establish a `pdc` global in JS, and keep the datacite enumerations there.
- remove the ERB templates from the JS, and move the file over with the rest of the JS.
- remove references to the partial, since it doesn't exist.
- Fix one undeclared variable. (Vite imports this as a ES module now, so the syntax is a little fussier.)
- Add a `vite_javascript_tag`... but I think Vite should just be scanning the directory for files, and this shouldn't be needed... I may be confused, or something may be misconfigured.

I've confirmed that the edit page loads and the pull downs are there... but this JS is now included on every page in the application, and it's possible there are unintended changes in behavior. I don't really know all the functionality of the application. My sense is that just having one site-wide bundle for the JS is more standard and will be easier to maintain... but it's possible this is not the time to do it. Thoughts?


mccalluc:
... and I should fix the test failures in any case.

mccalluc:
Talked with Hector: Making this application-wide is risky, so where there were partial includes formerly, I'm going to import the script instead... maybe using Vite.

hectorcorrea (approved):

Issue #309, opened 2022-08-17
remove ETDs, Library Resources from the codebase


carolyncole:
These items were defined as out of scope for PDC describe.
Library Resources will be in Figgy
ETDs will not be in PDC Describe (unsure where they will live)

- [x] Search for Library Resources and ETD to find the locations
- [x] Remove them from the staging and production servers
- [x] Remove any associated tests

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-Workcycles-61a4f1a12a399b001730f65a/issues/pulibrary/pdc_describe/309?planning-poker) with ZenHub @bess @hectorcorrea @jrgriffiniii

carolyncole:
Some of this was completed by #352

carolyncole:
Staging & Production already only had the two expected collections

bess:
Closing!

PR #555, opened 2022-10-25
Removing the library_resource collection

carolyncole:
refs #309

The collection was not utilized, otherwise the test would have failed.


hectorcorrea (approved):

hectorcorrea:
@carolyncole we could probably remove `Collection.library_resource` from the `Collection` class itself too since we are not going to store Library resources in PDC Describe (at one point we thought we would.)

carolyncole:
@hectorcorrea where is that defined?  I was not finding it.  When I look at collection it is not defined there.

hectorcorrea (approved):

PR #558, opened 2022-10-25
Track files added directly to AWS S3

hectorcorrea:
Logs files added directly to AWS S3. Notice that we do not link the new files to the current user since we really don't know who added them to S3. The display for now does not display a name, we could update it to display something, but for now in this PR I am only focusing on showing that a file was added to the work.

Screen Shot 2022-10-25 at 5 46 01 PM

Works toward #517

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -366,7 +366,7 @@ class Work < ApplicationRecord
 
   def log_file_changes(changes, current_user)
     return if changes.count == 0
-    WorkActivity.add_system_activity(id, changes.to_json, current_user.id, activity_type: "FILE-CHANGES")
+    WorkActivity.add_system_activity(id, changes.to_json, current_user&.id, activity_type: "FILE-CHANGES")
   end
 
   def activities
@@ -489,10 +489,16 @@ class Work < ApplicationRecord
   # However, a consequence of this is that #after_save is invoked whenever a new attached Blob or Attachment object is persisted.
   def attach_s3_resources
     return if approved?
+    changes = []
     # This retrieves and adds S3 uploads if they do not exist
     pre_curation_s3_resources.each do |s3_file|
+      changes << { action: :added, filename: s3_file.filename }
       add_pre_curation_s3_object(s3_file)
     end
+
+    # Log the new files, but don't link the change to the current_user since we do not know
+    # who added the files directly to AWS S3.
+    log_file_changes(changes, nil) if changes.count > 0
   end
 
   delegate :resource_type, :resource_type=, :resource_type_general, :resource_type_general=, to: :resource


app/models/work.rb
```diff
@@ -366,7 +366,7 @@ def log_changes(resource_compare, current_user)
 
   def log_file_changes(changes, current_user)
     return if changes.count == 0
-    WorkActivity.add_system_activity(id, changes.to_json, current_user.id, activity_type: "FILE-CHANGES")
+    WorkActivity.add_system_activity(id, changes.to_json, current_user&.id, activity_type: "FILE-CHANGES")
```
carolyncole, line 369: @hectorcorrea I wonder if the parameter should really be `current_user_id` and the two places calling log_file_changes should pass the id.  The impression that a current_user will be sent in and then not could get confusing when maintaining the code.

app/models/work_activity.rb
```diff
@@ -68,9 +68,14 @@ def comment_html
...
-        "<i>#{change['action']}</i> <span>#{change['filename']}<span><br/>"
+        icon = if change["action"] == "deleted"
+                 '<i class="bi bi-file-earmark-minus-fill file-deleted-icon"></i>'
+               else
+                 '<i class="bi bi-file-earmark-plus-fill file-added-icon"></i>'
+               end
+        "<tr><td>#{icon}</td><td>#{change['action']}</td> <td>#{change['filename']}</td>"
       end
-      "<p><b>Files</b>: #{changes_html.join}</p>"
+      "<p><b>Files updated:</b></p><table>#{changes_html.join}</table>"
```
carolyncole, line 78: I like the table!

app/views/works/_work_activity.html.erb
```diff
@@ -0,0 +1,36 @@
...
+      <% if activity.activity_type == 'CHANGES' %>
+        <div class="activity-history-item activity-history-log-item">
+          <div class="activity-history-log-title">
+            <b><%= activity.created_by_user&.display_name_safe %></b>
+            <span title="<%= activity.created_at.localtime %>"><%= distance_of_time_in_words_to_now(activity.created_at) %></span>
+          </div>
+          <span class="comment-html"><%= activity.message_html.gsub('{USER-PATH-PLACEHOLDER}', users_path).html_safe %></span>
+        </div>
+      <% elsif activity.activity_type == 'FILE-CHANGES' %>
+        <div class="activity-history-item activity-history-log-item">
```
carolyncole, line 18: I would love to see this as a partial and the parts that are changing based on the type passed in as locals
`render partail: 'activity_log', locals: { user_name: activity.created_by_user&.display_name_safe, time_title: activity.created_at.localtime, time: distance_of_time_in_words_to_now(activity.created_at), comment: activity.message_html.gsub('{USER-PATH-PLACEHOLDER}', users_path).html_safe }`

``` In the partial
<div class="activity-history-item activity-history-comment-item">
      <div class="activity-history-comment-title">
        <b><%= user_name %></b>
        <span title="<%= time_title %>"><%= time %></span>
      </div>
      <span class="comment-html"><%= comment %></span>
    </div>
```

carolyncole (changes requested):
@hectorcorrea a couple of changes, possibly...

carolyncole:
Also I tested this on staging with two S3 files and a globus file.  All three showed up nicely.

app/views/works/_work_activity.html.erb
hectorcorrea: I created a separate issue for this: https://github.com/pulibrary/pdc_describe/issues/564

app/models/work.rb
hectorcorrea: Good idea. I've made the change.

app/models/work.rb
```diff
@@ -489,10 +489,17 @@ def delete_pre_curation_s3_dir
...
     pre_curation_s3_resources.each do |s3_file|
-      add_pre_curation_s3_object(s3_file)
+      if add_pre_curation_s3_object(s3_file)
+        changes << { action: :added, filename: s3_file.filename }
+      end
     end
+
+    # Log the new files, but don't link the change to the current_user since we really don't know
+    # who added the files directly to AWS S3.
+    log_file_changes(changes, nil)
```
jrgriffiniii, line 502: 👍

jrgriffiniii (approved):

app/models/work_activity.rb
```diff
@@ -68,9 +68,14 @@ def comment_html
     def file_changes_html
       changes = JSON.parse(message)
       changes_html = changes.map do |change|
-        "<i>#{change['action']}</i> <span>#{change['filename']}<span><br/>"
+        icon = if change["action"] == "deleted"
+                 '<i class="bi bi-file-earmark-minus-fill file-deleted-icon"></i>'
+               else
+                 '<i class="bi bi-file-earmark-plus-fill file-added-icon"></i>'
+               end
+        "<tr><td>#{icon}</td><td>#{change['action']}</td> <td>#{change['filename']}</td>"
```
mccalluc, line 76: In ERB, by default it HTML-encodes templated values... but that's not happening here. We wouldn't expect to have any special characters in `change['action']` or `change['filename']`... but if there were user supplied data, we'd have a XSS vulnerability ... and it's hard to evaluate whether each case is safe or not.

One possibility would be to import https://api.rubyonrails.org/classes/ERB/Util.html#method-c-h and have a practice of using it any time we're interpolating plain strings into html

```suggestion
        "<tr><td>#{icon}</td><td>#{h(change['action'])}</td> <td>#{h(change['filename'])}</td>"
```

A complementary practice might be to mark html strings as such, so every interpolation is either `html_...` or `h(...)`

```suggestion
        "<tr><td>#{html_icon}</td><td>#{change['action']}</td> <td>#{change['filename']}</td>"
```

Or a different approach would be to use `html_safe`... but I think that could get confusing outside of ERB.

---------

But it probably makes sense to decide on a policy and apply it across the board, rather than worry about it here. But wanted to bring it up.

mccalluc (approved):
Some concerns about XSS vulnerabilities, but I'll file a new issue for that and we can decide on a path and apply it across the board.

app/models/work_activity.rb
hectorcorrea: These are all good observations. In this case the `change['action']` values are not user entered so we are OK. This is an example of how these values are added (see line 496): https://github.com/pulibrary/pdc_describe/blob/main/app/models/work.rb#L488-L498

PR #567, opened 2022-10-26
Apply some manual changes for eslint

mccalluc:
This starts to get into stylistic preferences... I don't care strongly about these, but I do think it makes sense to just pick a style guide and stick with it. There are other options, if this doesn't feel like a style you like.


mccalluc:
Hmm... Circle CI seems to be running on the backend, but display of results fails. I'm going to undo the last couple commits on this branch and leave it at the last green checkmark.

jrgriffiniii (approved):

Issue #563, opened 2022-10-26
accepted is not the approved state and the system should error


carolyncole:
We should cause an error when the work is set to a state that is not one of our valid states.

It is possible this is a configuration flag in the Act As State Machine (AASM) gem (https://github.com/aasm/aasm).

In the tests we are setting the state to "accepted", but we should be utilizing "approved" for an approved work.

This seems to be the only location this is still an issue:
https://github.com/pulibrary/pdc_describe/blob/1938727491805a46119ffab5b1c0fc8f35b5a24c/spec/system/work_upload_s3_objects_spec.rb#L128

- [ ] Remove accepted from the test and determine what should really be tested.  It does not make sense to check approved in "Uploading S3 Bucket Objects for new Work"
- [ ] Set up a failsafe to only allow defined states in the work model
- [ ] Add an automated test to be certain the code catches when a bad state is set

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-Workcycles-61a4f1a12a399b001730f65a/issues/pulibrary/pdc_describe/563?planning-poker) with Zenhub @bess @hectorcorrea @mccalluc

carolyncole:
https://github.com/aasm/aasm/blob/252ade3a20dd86abed3ec76b965f85da530e2611/README_FROM_VERSION_3_TO_4.md#new-configuration-option-no_direct_assignment May be the AASM way to not allow invalid states.  We may just want a validation on the model that the state is one of `work.aasm.states.map(&:to_s).include?(work.state)` (assuming work is set)

Issue #519, opened 2022-10-19
Download link for approved content should link to Globus download


carolyncole:
Currently we are linking directly to amazon, which gives an access error.
We would like this link to utilize Globus, where we have public access set up. 

The Globus url should be configurable on a per environment basis:
  Staging: https://g-5beea4.90d4e.bd7c.data.globus.org/pdc-describe-staging-postcuration
  Production: https://g-ef94ef.f0ad1.36fe.data.globus.org

In addition files names with spaces should create the appropriate url.

I had a file that included a space in the name.  This caused the show page to fail after it had been approved.
https://pdc-describe-staging.princeton.edu/describe/works/92

## Error
[pdc_describe/staging] URI::InvalidURIError: bad URI(is not URI?): "https://pdc-describe-staging-postcuration.s3.amazonaws.com/10.80021/ge21-2181/92/10.34770-tbd-1-Screen Shot 2022-09-09 at 9.44.24 PM.png"

## Backtrace

    line 35 of [PROJECT_ROOT]/app/models/s3_file.rb: uri
    line 39 of [PROJECT_ROOT]/app/models/s3_file.rb: url
    line 25 of [PROJECT_ROOT]/app/views/works/uploads/_post_curation_s3_resources.html.erb: block in _app_views_works_uploads__post_curation_s__resources_html_erb__1251326391469060684_106300

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/99214/faults/89325049)

## Acceptance criteria
- [ ] I can load https://pdc-describe-staging.princeton.edu/describe/works/92
- [ ] An automated test to show spaces in file names do not cause issues after being approved
- [ ] Download links start with \Globus urls instead of amazon s3 urls

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-Workcycles-61a4f1a12a399b001730f65a/issues/pulibrary/pdc_describe/519?planning-poker) with Zenhub @bess @hectorcorrea @jrgriffiniii

PR #571, opened 2022-10-27
Fix test typos

mccalluc:
- Partial solution for #563

While trying to address 
- #519

we got stuck because there were typos in the tests... and fixing the typos caused the tests to fail. There were two typos:
- `filename2` was assigned a value based on `file1`.
- The state was set to `accepted` rather than `approved`.

In this PR I'm trying to make a minimal set of changes that fix the typos, and get the tests to pass.... but I don't really understand the desired behavior in detail, and I'm not confident that there aren't other problems in the supporting code in the tests.


mccalluc:
huh: timeout error on that last run:
```
  1) Form submission for migrating attention migrate record from dataspace produces and saves a valid datacite record
     Failure/Error: fill_in "family_name_4", with: "Guterstam"
     
     Net::OpenTimeout:
       execution expired
```

hectorcorrea (approved):

PR #572, opened 2022-10-27
Validate work state

mccalluc:
- Fix: #563
- Blocked by: #571

The tests currently set an invalid state, so those tests will error out in this branch. I made an attempt to fix that in #571... but I'm not super confident about that work, so I didn't base this branch on that one. Where ever the fix comes from, that PR or a different one, once it's merged to main, merge main to this branch, and it should be green.

Overriding the setter and catching the error at the earliest possible point feels good, but there may be other approaches. I looked at `no_direct_assignment`, but that causes tons of FactoryBot errors.

### DIFF
--- a/spec/models/work_spec.rb
+++ b/spec/models/work_spec.rb
@@ -55,8 +55,8 @@ RSpec.describe Work, type: :model do
   end
 
   it "prevents invalid state assignment" do
-    work = Work.new()
-    expect { work.state = 'sorry' }.to raise_error(StandardError, /Invalid state 'sorry'/)
+    work = Work.new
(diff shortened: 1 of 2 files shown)


spec/models/work_spec.rb
```diff
@@ -54,6 +54,11 @@
     expect { work.save! }.to raise_error ActiveRecord::RecordInvalid
   end
 
+  it "prevents invalid state assignment" do
+    work = Work.new
+    expect { work.state = "sorry" }.to raise_error(StandardError, /Invalid state 'sorry'/)
```
hectorcorrea, line 59: 😄

hectorcorrea (approved):

PR #573, opened 2022-10-27
Fixes error saving resource_type_general

### LINKED ISSUE #562: resource_type_general bug
The field `resource_type_general` is being reported as changed with every save, even when we don't change it. I am not sure if the problem is in the field itself (the way we save and fetch it) or with the code to detect changes to it.

hectorcorrea:
The core of this PR is a little change to handle the fact that were using two different kind of values for resource_type_general. Sometimes we were using a symbol (e.g. `:DATASET`) and other times we were using a string (e.g. `"DATASET"`). 

That fix was easy, I just made sure we convert it to symbol on save :
```
        resource.resource_type_general = params["resource_type_general"].to_sym if params["resource_type_general"]
```

In troubleshooting this I noticed that we do not allow the user to enter this value (and others) unless the have the right privileges, which is great, except that we were losing the **existing values** when a user without the right privilege edited other values in the work. I changed the code to always show the values as HTML INPUT elements but make them readonly so that the user cannot change them but they are rendered and posted back on save.

Screen Shot 2022-10-27 at 5 06 08 PM

Closes #562


app/services/form_to_resource_service.rb
```diff
@@ -31,16 +31,14 @@ def convert(params, work, current_user)
...
-          resource.collection_tags = params["collection_tags"].split(",").map(&:strip) if params["collection_tags"]
-          resource.resource_type = params["resource_type"] if params["resource_type"]
-          resource.resource_type_general = params["resource_type_general"] if params["resource_type_general"]
-        end
+        resource.doi = params["doi"] if params["doi"].present?
+        resource.ark = params["ark"] if params["ark"].present?
+        resource.version_number = params["version_number"] if params["version_number"].present?
+        resource.collection_tags = params["collection_tags"].split(",").map(&:strip) if params["collection_tags"]
+        resource.resource_type = params["resource_type"] if params["resource_type"]
+        resource.resource_type_general = params["resource_type_general"].to_sym if params["resource_type_general"]
```
hectorcorrea, line 41: The call to `to_sym` here is what was missing to make sure we always use a symbol and not a string.

app/services/form_to_resource_service.rb
```diff
@@ -8,8 +8,8 @@ class << self
     #  @param [User] current_user user currently authorized with the system.  Utilizes the current_user to validate access.
```
mccalluc, line 8:
```suggestion
    #  @param [User] _current_user user currently authorized with the system.  Utilizes the _current_user to validate access.
```

mccalluc: ... but it's not actually used, so why not drop it entirely?

app/services/form_to_resource_service.rb
mccalluc, line 41:
```suggestion
        resource.resource_type_general = params["resource_type_general"]&.to_sym
```
Should be equivalent?

app/views/works/_curator_controlled_display.html.erb
```diff
@@ -1,35 +1,37 @@
 <div class="section-title">
-  <!-- DOI field (mostly for legacy DSpace works) -->
+  <!-- DOI field -->
   <div class="field">
-    DOI <br>
-    <%= @work.doi %>
+    DOI:<br>
+    <input class="text-not-editable" type="text" id="doi" name="doi" value="<%= @work.resource.doi %>" readonly />
```
mccalluc, line 5:
```suggestion
    <input id="doi" name="doi" value="<%= @work.resource.doi %>" readonly />
```

This class is being consistently applied with `readonly`, so I think just `readonly` would suffice? There's one CSS rule to update, but I think that's it. (See earlier comment.)

`text` is the default, so we could leave it off, I think?

(Same comment for all the inputs below.)

app/assets/stylesheets/application.scss
```diff
@@ -123,3 +123,8 @@
   color:#27c0fb;
   font-size: 18px;
 }
+
+input[readonly] {
+  border-width: 1px;
+  border-style: solid;
```
mccalluc, line 129:
```suggestion
  border-style: solid;
  background: #f4f4f4;
```

and get rid of the redundant `text-not-editable`

mccalluc (changes requested):
I think there are some redundant bits that could be be left out... but if I'm missing the bigger picture, never mind.

hectorcorrea:
@mccalluc thanks for the suggestions. I've applied them.

mccalluc (approved):

PR #574, opened 2022-10-28
Implementing support for the enabling and disabling of e-mail message notifications from members of Collections to Users

### LINKED ISSUE #360: Email Notifications can be configured in addition to in Application Notifications
This Gem is a possible solution: [mailboxer](https://rubygems.org/gems/mailboxer/versions/0.15.1 ) allows configuration for both in app notifications and also email notifications.
This is utilized in [hyrax](https://github.com/samvera/hyrax/blob/main/hyrax.gemspec#L65)
Possible alternatives to mailboxer: https://www.ruby-toolbox.com/categories/e_mail.html

The system user should be able to decide if they want email or just in app notifications.

This is complete when

- [x] The moderator has the ability to configure notifications
- [x] The current set of notifications also trigger emails if desired ( This includes when a person is tagged )
- [x] There is a record of each notification that has been sent
- [x] There are automated test that show a user can configure emails and does receive them if configure

jrgriffiniii:
Resolves #360


mccalluc (commented):
Maybe it's too late on a Friday, but the tests are confusing me. I might be missing something obvious.

spec/models/collection_spec.rb
```diff
@@ -54,4 +54,112 @@
...
+
+        updated_state = collection.messages_enabled_for?(user: user)
+        expect(updated_state).to be true
+
+        collection.disable_messages_for(user: user)
+        collection.save!
+        collection.reload
+
+        updated_state = collection.messages_enabled_for?(user: user)
+        expect(updated_state).to be true
```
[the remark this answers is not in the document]
escowles: I'll defer to the team if there's a preference, but I haven't heard a philosophy articulated in this area. Personally, I feel like keeping the tests maintainable and readable would be more important to me than making them literal/uncomplicated.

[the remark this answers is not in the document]
jrgriffiniii: I am very sorry for these errors. Please know that this was related to my erroneous usage of `ActiveRecord::Associations::CollectionProxy#delete_by` (this should be https://api.rubyonrails.org/v6.1.6.1/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-destroy). I have also added comments to the test suite clarifying the context for these.

app/controllers/users_controller.rb
```diff
@@ -54,11 +60,32 @@ def set_user
 
     # Only allow a list of trusted parameters through.
     def user_params
-      params.require(:user).permit([:display_name, :full_name, :family_name, :orcid, :email_messages_enabled])
+      @user_params ||= params.require(:user).permit([:display_name, :full_name, :family_name, :orcid, :email_messages_enabled, { collections_with_messaging: {} }])
```
mccalluc, line 63: I'm not sure what role the `@user_params` plays?

app/controllers/users_controller.rb
```diff
@@ -54,11 +60,32 @@ def set_user
...
+      form_value.to_s == COLLECTION_MESSAGING_ENABLED
+    end
+
+    def update_collections_with_messaging
+      if user_params.key?(:collections_with_messaging)
+        extracted = user_params.extract!(:collections_with_messaging)
+        collections_with_messaging = extracted[:collections_with_messaging]
+
+        collections_with_messaging.each_pair do |collection_id, param|
+          selected_collection = Collection.find_by(id: collection_id)
```
mccalluc, line 81: Is each `find_by` a separate DB hit? If we know is a small number, then probably fine, but if it could be large, consider finding them all with one query. It looks like there used to be a [build-in](https://stackoverflow.com/questions/9650205/rails-find-all-with-conditions) syntax for this, but we'd now use `where()`.

app/models/collection.rb
```diff
@@ -74,7 +81,6 @@ def delete_permission(current_user, removed_user)
         errors.add(:delete_permission, "Cannot remove yourself from a collection. Contact a super-admin for help.")
       else
         errors.delete(:delete_permission)
-        removed_user.remove_role :collection_admin, self
```
mccalluc, line 77: Maybe retitle `delete_permission`? I'm not sure which permission I should expect to see removed here.

app/models/user.rb
```diff
@@ -214,5 +222,32 @@ def assign_default_role
   def email_messages_enabled?
     email_messages_enabled
   end
+
+  def collections_with_messaging_dis
```
mccalluc, line 226: I'm not sure what `dis` means in this context. I'm also not sure that this method is used? It might be implicit.

app/controllers/users_controller.rb
```diff
@@ -54,11 +60,32 @@ def set_user
...
     def can_edit?
       return true if current_user.id == @user.id
       current_user.super_admin?
     end
+
+    def parameter_enables_messaging?(form_value)
+      form_value.to_s == COLLECTION_MESSAGING_ENABLED
+    end
+
+    def update_collections_with_messaging
```
carolyncole, line 75: Is there a way to break this method up? possibly move the contents of the loop into a method
```
      def configure_collection_notification(collection_id, param)
          selected_collection = Collection.find_by(id: collection_id)

          if parameter_enables_messaging?(param)
            @user.enable_messages_from(collection: selected_collection)
          else
            @user.disable_messages_from(collection: selected_collection)
          end
      end
```

.rubocop.yml
```diff
@@ -37,11 +37,13 @@ Metrics/ClassLength:
   Exclude:
     - "app/services/s3_query_service.rb"
     - "app/models/pdc_metadata/resource.rb"
+    - "app/models/collection.rb"
 
 Metrics/MethodLength:
   Exclude:
     - "app/models/work.rb"
     - "spec/support/work_s3_requests_specs.rb"
+    - "app/controllers/users_controller.rb"
```
carolyncole, line 46: I would love to not exclude all methods from size checking.

app/models/collection.rb
```diff
@@ -3,6 +3,13 @@
 class Collection < ApplicationRecord
   resourcify
 
+  # CollectionOptions model extensible options set for Collections and Users
+  has_many :collection_options, dependent: :destroy
+  has_many :collection_messaging_options, -> { where(option_type: CollectionOption::EMAIL_MESSAGES) }, class_name: "CollectionOption", dependent: :destroy
```
carolyncole, line 8: cool!

app/models/collection.rb
carolyncole, line 77: Why did this get changed in this PR?

app/models/collection.rb
```diff
@@ -86,6 +92,29 @@ def submitters
     User.with_role(:submitter, self)
   end
 
+  # Permit a User to receive notification messages for members of this Collection
+  # @param user [User]
+  def enable_messages_for(user:)
+    raise(ArgumentError, "User #{user.uid} is not an administrator for this collection #{title}") unless user.super_admin? || user.can_admin?(self)
```
carolyncole, line 98: I believe this will also work, because [super admins have administrative](https://github.com/pulibrary/pdc_describe/blob/a46d7752190be72e777d0616e97b836240aae7e3/app/models/user.rb#L72) access to any collection.
```suggestion
    raise(ArgumentError, "User #{user.uid} is not an administrator for this collection #{title}") unless user.can_admin?(self)
```

app/models/collection.rb
```diff
@@ -86,6 +92,29 @@ def submitters
...
+    collection_messaging_options << CollectionOption.new(option_type: CollectionOption::EMAIL_MESSAGES, collection: self, user: user)
+  end
+
+  # Disable a User from receiving notification messages for members of this Collection
+  # @param user [User]
+  def disable_messages_for(user:)
+    raise(ArgumentError, "User #{user.uid} is not an administrator for this collection #{title}") unless user.super_admin? || user.can_admin?(self)
```
carolyncole, line 105: same here:
```suggestion
    raise(ArgumentError, "User #{user.uid} is not an administrator for this collection #{title}") unless user.can_admin?(self)
```

spec/models/user_spec.rb
```diff
@@ -238,4 +238,112 @@
...
+        expect(enabled_state).to be true
+
+        user.disable_messages_from(collection: collection)
+        user.save!
+        user.reload
+
+        disabled_state = user.messages_enabled_from?(collection: collection)
+        expect(disabled_state).to be false
+      end
+    end
```
mccalluc, line 340: All your `disable` tests also cover `enable`: If they were really orthogonal, separate tests would be great... but since a failure in `enable` would also break `disable` tests, I think we could get by with just the `disable` tests.

app/models/collection.rb
```diff
@@ -86,6 +92,29 @@ def submitters
...
+    users_with_messaging.destroy(user)
+  end
+
+  # Returns true if a given user has notification e-mails enabled for this collection
+  # @param user [User]
+  # @return [Boolean]
+  def messages_enabled_for?(user:)
+    found = users_with_messaging.find_by(id: user.id)
+
+    !found.nil?
```
carolyncole, line 115: I find positive checks easier to read in the long run.
```suggestion
    found.present?
```

carolyncole (changes requested):
Looking really nice.  Just a couple of adjustments

mccalluc (approved):
More changes since I started review, so this might be moot. Some questions, but I think it looks good. Where I'm confused, it might be worth while to check with another dev and see what they think: It may just be me.

app/models/collection.rb
jrgriffiniii: I am sorry, this must have been a rebase error. This has been fixed.

app/controllers/users_controller.rb
carolyncole: @mccalluc I'm guessing this is just a [memoizing ](https://www.justinweiss.com/articles/4-simple-memoization-patterns-in-ruby-and-one-gem/)

app/controllers/users_controller.rb
```diff
@@ -27,9 +31,12 @@ def edit
   end
 
   # PATCH/PUT /users/1 or /users/1.json
+  # rubocop:disable Metrics/MethodLength
```
carolyncole, line 34: I think this is no longer needed.  Thanks for breaking up the routine...  Sorry if you already planned to remove it.

app/controllers/users_controller.rb
jrgriffiniii: No, I apologize, `collections_with_messaging = extracted[:collections_with_messaging]` yields a `Hash`, which itself is parsed from the HTTP request `params`

app/helpers/application_helper.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 module ApplicationHelper
+  # include PdcDescribeHelper
```
jrgriffiniii, line 3: This was an experiment I attempted, but I can remove this for now.

app/controllers/users_controller.rb
jrgriffiniii: Yes, sorry, the memoization is needed given that https://github.com/pulibrary/pdc_describe/pull/574/files#diff-cfdccd0a9d5df5a43aaad2a35d36ebbe187c52ad5fdc9846fa189d04537adb6eR77 invokes `#extract!`, updating the state of the HTTP request `params`. If this is not explicitly memoized, then `#extract!` does not mutate the `params` (please see https://api.rubyonrails.org/v6.1.6.1/classes/ActionController/Parameters.html#method-i-extract-21 for reference)

app/controllers/users_controller.rb
jrgriffiniii: I am sorry, https://github.com/pulibrary/pdc_describe/pull/574/commits/a542b7eba7d814f2d3d6855c88ad4238c6058718 still raises a failure (please see https://app.circleci.com/pipelines/github/pulibrary/pdc_describe/1738/workflows/7eeffb68-411d-457c-8717-6d95d478a97f/jobs/5069). I am going to restore this.

app/models/collection.rb
jrgriffiniii: I am sorry, this was removed in error during a rebase.

carolyncole (approved):

PR #575, opened 2022-10-31
Adding an automated test for curator notification

### LINKED ISSUE #359: Notify the collection moderators when a item in the collection is marked complete
This can be accomplished via the after on the AASM,  like [publish](https://github.com/pulibrary/pdc_describe/blob/main/app/models/work.rb#L35)

This is complete when

- [x] The moderator(s) of a collection receives an in app notification when a dataset is marked as complete
- [ ] There is an automated test that checks that these notifications get sent

carolyncole:
fixes #359

### DIFF
--- a/spec/system/form_submission_spec.rb
+++ b/spec/system/form_submission_spec.rb
@@ -3,6 +3,7 @@ require "rails_helper"
 
 RSpec.describe "Form submission for a legacy dataset", type: :system do
   let(:user) { FactoryBot.create(:princeton_submitter) }
+  let!(:curator) { FactoryBot.create(:user, collections_to_admin: [Collection.first]) }
   let(:title) { "Sowing the Seeds for More Usable Web Archives: A Usability Study of Archive-It" }
   let(:contributors) do
     [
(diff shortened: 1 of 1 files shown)


spec/system/form_submission_spec.rb
```diff
@@ -75,6 +77,15 @@
...
       expect(page).to have_content "awaiting_approval"
+
+      # Now sign is as the collection curator and see the notification on your dashboard
+      sign_out user
+      sign_in curator
+      visit(user_path(curator))
+      expect(page).to have_content curator.display_name
+      within("#unfinished_datasets span.badge.rounded-pill.bg-primary") do
+        expect(page).to have_content "1"
+      end
```
mccalluc, line 88:
```suggestion
      within("#unfinished_datasets") do
        expect(page).to have_content "Sowing the Seeds for More Usable Web Archives: A Usability Study of Archive-It"
        expect(page).to have_content "awaiting_approval"
      end
```
The exact element path doesn't seem like a requirement, but seeing the title and the status does.

mccalluc (approved):
Looks good: A couple suggestions.

spec/system/form_submission_spec.rb
carolyncole: @mccalluc I'm looking for the badge with the 1 here.  I'll add a comment

spec/system/form_submission_spec.rb
```diff
@@ -39,6 +40,7 @@
   end
   context "happy path" do
     it "produces and saves a valid datacite record", js: true do
+      curator # make sure the curator exists prior to starting this process so they can be notified
```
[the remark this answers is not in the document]
carolyncole: This was just to construct the item, I could have done a let! instead.  Maybe that would be better

carolyncole:
@mccalluc do these changes make more sense to you?

mccalluc (approved):

PR #576, opened 2022-10-31
Updating replacement key to be upload key

### LINKED ISSUE #557: Work edit - replace single file does not work
Attempting to replace a single file via the Work + Edit screen does not work. Notice how I selected file "small.txt" to replace "tiny.txt" in the following screenshot

Screen Shot 2022-10-25 at 3 53 50 PM

but then the code in [WorkUploadsEditService](https://github.com/pulibrary/pdc_describe/blob/main/app/services/work_uploads_edit_service.rb#L34-L48) fails to replace the file. This is because the comparison is comparing and index value to a key value and they do not match:

Screen Shot 2022-10-25 at 3 55 55 PM

carolyncole:
This will allow the replacement to be more accurate and replace the expected upload.

fixes #557

### DIFF
--- a/app/views/works/_uploads_form.html.erb
+++ b/app/views/works/_uploads_form.html.erb
@@ -33,7 +33,7 @@
                     </td>
                     <td><span><%= upload.created_at %></span></td>
                     <td>
-                      <input id="work-deposit-uploads" name="work[replaced_uploads][<%= upload.id %>]" type="file" />
+                      <input id="work-deposit-uploads-<%= upload.id %>" name="work[replaced_uploads][<%= upload.key %>]" type="file" />
                     </td>
                     <td>
                       <input id="work-uploads-<%= upload.id %>-delete" name="work[deleted_uploads][<%= upload.key %>]" type="checkbox" value="1"/>
--- a/spec/controllers/works_controller_spec.rb
+++ b/spec/controllers/works_controller_spec.rb
@@ -299,21 +299,21 @@ RSpec.describe WorksController do
         file.close
         file
       end
-      let(:uploaced_temp_file1) { Rack::Test::UploadedFile.new(temp_file1.path, "text/plain") }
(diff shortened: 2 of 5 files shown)


app/views/works/_uploads_form.html.erb
```diff
@@ -33,7 +33,7 @@
                     </td>
                     <td><span><%= upload.created_at %></span></td>
                     <td>
-                      <input id="work-deposit-uploads" name="work[replaced_uploads][<%= upload.id %>]" type="file" />
+                      <input id="work-deposit-uploads-<%= upload.id %>" name="work[replaced_uploads][<%= upload.key %>]" type="file" />
```
mccalluc, line 36: Just looking at this, I don't have a clean sense of what `id` is for, and what `key` is for.

spec/controllers/works_controller_spec.rb
```diff
@@ -327,8 +327,8 @@
 
       let(:uploaded_files) do
         {
-          "0" => uploaded_file1,
-          "2" => uploaded_file2
+          work.pre_curation_uploads[0].key => uploaded_file1,
+          work.pre_curation_uploads[2].key => uploaded_file2
```
mccalluc, line 331: It surprises me that the numbers aren't in sync... Maybe a comment to explain why?

mccalluc (approved):
A couple questions, but looks good.

app/views/works/_uploads_form.html.erb
carolyncole: @mccalluc I'm matching delete for consistency. Key in ActiveStorage is the file path in s3 (or other storage back ends) and id is the id of the database record pointing to S3

spec/controllers/works_controller_spec.rb
carolyncole: @mccalluc as the test did previously, it is replacing the first and third items. https://github.com/pulibrary/pdc_describe/blob/main/spec/controllers/works_controller_spec.rb#L291 I believe the context explains why not all...

Issue #487, opened 2022-10-13
Augment cytoskeletal migration fixture to include related identifier for paper


bess:
Once the tickets described in [this ticket](https://github.com/pulibrary/pdc_describe/issues/486) are complete, augment [the cytoskeletal migration fixture](https://github.com/pulibrary/pdc_describe/blob/main/spec/system/cytoskeletal_form_submission_spec.rb) to include the missing related identifier for the journal publication (paper) associated with this dataset.

### Acceptance Criteria
- [x] The system spec puts the related identifier into the right place on the form
- [x] Add related objects to the DataCite serialization
- [x] The DataCite record that is produced is re-validated with RDOS to ensure this field is being mapped as expected

bess:
```xml
<?xml version="1.0"?>
<resource xsi:schemaLocation='http://datacite.org/schema/kernel-4 http://schema.datacite.org/meta/kernel-4/metadata.xsd' xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance' xmlns='http://datacite.org/schema/kernel-4'>
  <identifier identifierType='DOI'>10.34770/r2dz-ys12</identifier>
  <creators>
    <creator>
      <creatorName>Taylor, Jenny A</creatorName>
      <givenName>Jenny A</givenName>
      <familyName>Taylor</familyName>
    </creator>
    <creator>
      <creatorName>Bratton, Benjamin P</creatorName>
      <givenName>Benjamin P</givenName>
      <familyName>Bratton</familyName>
    </creator>
    <creator>
      <creatorName>Sichel, Sophie R</creatorName>
      <givenName>Sophie R</givenName>
      <familyName>Sichel</familyName>
    </creator>
    <creator>
      <creatorName>Blair, Kris M</creatorName>
      <givenName>Kris M</givenName>
      <familyName>Blair</familyName>
    </creator>
    <creator>
      <creatorName>Jacobs, Holly M</creatorName>
      <givenName>Holly M</givenName>
      <familyName>Jacobs</familyName>
    </creator>
    <creator>
      <creatorName>DeMeester, Kristen E</creatorName>
      <givenName>Kristen E</givenName>
      <familyName>DeMeester</familyName>
    </creator>
    <creator>
      <creatorName>Kuru, Erkin</creatorName>
      <givenName>Erkin</givenName>
      <familyName>Kuru</familyName>
    </creator>
    <creator>
      <creatorName>Gray, Joe</creatorName>
[...]

PR #578, opened 2022-10-31
Add another missing related object to cytoskeletal migration test

bess:
Work toward #487

### DIFF
--- a/spec/system/cytoskeletal_form_submission_spec.rb
+++ b/spec/system/cytoskeletal_form_submission_spec.rb
@@ -77,8 +77,14 @@ RSpec.describe "Form submission for migrating cytoskeletal", type: :system, mock
 
       # Select Additional Metadata Tab
       click_on "v-pills-additional-tab"
+
+      # Related Objects
       fill_in "related_identifier_1", with: related_identifier
       find("#related_identifier_type_1").find(:xpath, "option[@value='#{related_identifier_type}']").select_option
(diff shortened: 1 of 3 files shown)


hectorcorrea (approved):

spec/system/cytoskeletal_form_submission_spec.rb
```diff
@@ -77,9 +77,15 @@
...
       click_on "v-pills-additional-tab"
+
+      # Related Objects
       fill_in "related_identifier_1", with: related_identifier
       find("#related_identifier_type_1").find(:xpath, "option[@value='#{related_identifier_type}']").select_option
       find("#relation_type_1").find(:xpath, "option[@value='#{relation_type}']").select_option
+      click_on "Add Another Related Object"
+      fill_in "related_identifier_2", with: "https://doi.org/10.7554/eLife.52482"
+      find("#related_identifier_type_2").find(:xpath, "option[@value='DOI']").select_option
+      find("#relation_type_1").find(:xpath, "option[@value='#{relation_type}']").select_option
```
mccalluc, line 88:
```suggestion
      find("#relation_type_1").find("option[value='#{relation_type}']").select_option
```
I don't think we need the xpath syntax here, though it can be useful.

mccalluc (approved):
Looks good: `:xpath` probably isn't needed.

PR #580, opened 2022-10-31
Tweaks after reviewing codebase

mccalluc:
Looking through the code, there were some places I thought it could be simpler... in some cases rspec told me I was wrong, so I learned something from it.

It would be fine to close this unmerged, too.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -55,7 +55,6 @@ class UsersController < ApplicationController
     end
 
     def can_edit?
-      return true if current_user.id == @user.id
-      current_user.super_admin?
+      current_user.id == @user.id or current_user.super_admin?
     end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -13,16 +13,13 @@ class UsersController < ApplicationController
     @unfinished_works = Work.unfinished_works(@user)
     @completed_works = Work.completed_works(@user)
     @withdrawn_works = Work.withdrawn_works(@user)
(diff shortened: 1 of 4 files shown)


app/controllers/users_controller.rb
```diff
@@ -13,16 +13,13 @@ def show
...
 
   # GET /users/1/edit
   def edit
-    if can_edit?
-      render "edit"
-    else
+    unless can_edit?
       Rails.logger.warn("Unauthorized to edit user #{@user.id} (current user: #{current_user.id})")
-      redirect_to user_path(@user)
+      redirect_to user_path(@user) and return
```
carolyncole, line 22: Is the and return needed? I have never seen that syntax before.  Is the and a short cut for a semicolon?
If so I would prefer the return on the next line, so that it is not hidden.

carolyncole (changes requested):
@mccalluc nice updates!  just one question...

app/controllers/users_controller.rb
mccalluc: will do. I think the motivation where I've seen it used is to make sure the return immediately follows the redirect, to make sure there isn't a double render, but moving it to its own line would be fine.

carolyncole (approved):

mccalluc:
Test failure may just be a sporadic timeout:
```
  1) View status of data in S3 when a dataset has a DOI and its data is in S3 when item is approved when user is a curator shows data from S3
     Failure/Error: example.run
     
     Net::OpenTimeout:
       execution expired
```
rerunning.

PR #582, opened 2022-10-31
Move collection JS out of ERB

### LINKED ISSUE #579: Factor out `views/collections/_edit_javascript.html.erb`
It should be analogous to `edit_utils.js`... which should be renamed to clarify that it is only for works.

mccalluc:
- Fix #579


mccalluc:
oops: ERB syntax sneaked through...

app/views/collections/edit.html.erb
```diff
@@ -89,4 +89,15 @@
 <% end %>
 </ul>
 
-<%= render 'edit_javascript' %>
+<script>
+pdc = {
+  ...pdc,
```
hectorcorrea, line 94: What does the `...pdc` syntax do?

hectorcorrea: oh! I see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

hectorcorrea (approved):

PR #586, opened 2022-11-01
add admin privs to readme

mccalluc:
Took a little digging to find this, but not confident that it belongs in the readme, either


escowles:
👍🏻 to adding this to the README since it's important for local manual testing, and Figgy also has the in its README.

hectorcorrea (approved):

PR #593, opened 2022-11-03
Moving the Work form fields `Publisher`, `Publication Year`, and `Collection` to the `Curator Controlled` section

### LINKED ISSUE #561: Describe Form: Move specified properties to a different tab
## Priority

Not on fire, but needed before user testing and launch.

## Description

The following properties must be moved from the "Additional Metadata" tab to the "Curator Controlled" tab.
- Publisher
- Publication Year
- Collection

These are properties that we do not want depositors to be able to change themselves. However, they may request a curator to change when appropriate.

## Acceptance Criteria 
- [x] The specified properties (Publisher, Publication Year, Collection) will move to the "Curator Controlled" tab. 
- [x] An automated test to be certain that only curators and super admins have access to the properties and that they reside on the curator controlled tab

jrgriffiniii:
Resolves #561

### DIFF
--- a/app/views/works/_additional_form.html.erb
+++ b/app/views/works/_additional_form.html.erb
@@ -1,24 +1,4 @@
 <div class="section-title">
-  <div class="field">
-    Publisher:<span class="required-field">*</span><br>
-    <input type="text" id="publisher" name="publisher" class="input-text-long" value="<%= @work.resource.publisher %>" />
-  </div>
-
-  <div class="field">
-    Publication Year:<span class="required-field">*</span><br>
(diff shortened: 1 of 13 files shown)


mccalluc (approved):
Looks good.

I worry a little about the duplication between `_curator_controlled_display` and `_currator_controlled_form`. Not urgent, but I wonder if it would make sense either to factor out the shared bits into partials, or to merge the two into a single partial with a little logic to handle the differences... I'll make a diff and file a follow-up issue.

jrgriffiniii:
Thank you @mccalluc!

Issue #592, opened 2022-11-02
fixes to CKavity migration spec


Twade968:
This is the original record that we are migrating: https://dataspace.princeton.edu/handle/88435/dsp015999n626m?mode=full

### Acceptance Criteria
- [x]  put the three dc.creator fields in the original record into the DataCite Creator fields
- [x] Do not repeat any of the creators as a contributor(never have a person as both contributor and creator)
- [x] Do not put funding agencies as contributors(National Science foundation should be omitted)
- [x] There should not be a title and subtitle in the same field. Title should go in the Title field, subtitle needs to go in subtitle field and alternative title needs to go into alternative title field.
- [x] Publication needs to be set to 2019 instead of 2020( if there is no publication date use the dc.accessioned date as the publication year)
- [x] Publisher need to be set to Princeton University if it is apart of Princeton University (ex. if "Princeton University Lewis-Sigler Institute" is set as publisher just use "Princeton University" instead)
- [x] Use abstract in the description field but omit table of contents in the description field.

### Notes from Matt
there are several issues here. Let me preface everything by saying we (PRDS) did not curate this one, and I'm not sure which submission form was used for it.

I'm looking at the full item record to understand it better: https://dataspace.princeton.edu/handle/88435/dsp015999n626m?mode=full

The dc.contributor.author field was not used. Instead, there are three dc.creator fields in the original record that should be translated to the DataCite Creator field. One of the names listed as a creator was repeated as a contributor, and we do not want to replicate that error. In addition, we typically do not input funding agencies as contributors, so the Creator field you have for the National Science Foundation should be omitted.

The original record has a title as well as an alternative title, and the title field already includes a subtitle, so I do not think that the two title fields should be merged in DataCite. Instead, the alternative title from the original record can be translated as a second title with the type "AlternativeTitle" (see https://support.datacite.org/docs/datacite-metadata-schema-v44-mandatory-properties#3-title).

For some reason, this record does not have an issue date (which is generally required). However, I can see that the dc.date.accessioned is in 2019, so the DataCite record should have the Publication Year field set to 2019 instead of 2020.

While the original record does have the publisher set to "Princeton University Lewis-Sigler Institute", our current practice is to correct such entries to "Princeton University".

The original record has two description fields filled: abstract and table of contents. My understanding is that for the purposes of migration, we are copying over abstracts as Description Type "Other" and omitting other description fields for now.

Twade968:
```
<?xml version="1.0"?>
[...]

PR #595, opened 2022-11-07
Upgraded cklavity migration spec test

Twade968:
Fixes for ticket #592 

redescribed From this data set: https://dataspace.princeton.edu/handle/88435/dsp015999n626m?mode=full

```
Cannot generate XML at this time due to a known bug in code. 
Fixes for this bug are coming ASAP.
```


bess (changes requested):
Looking good! Why did you remove the keywords / subjects? Was that a request from Matt? I would think we would want them. 

Also, you've got a stray file. This PR should only contain your rspec file, not the yarn.lock file.

bess (approved):

PR #598, opened 2022-11-11
Validating the related objects

carolyncole:
Additionally stop empty resources from the form

### DIFF
--- a/app/models/pdc_metadata/related_object.rb
+++ b/app/models/pdc_metadata/related_object.rb
@@ -5,12 +5,25 @@ module PDCMetadata
   # let(:related_identifier_type) { "arXiv" }
   # let(:relation_type) { "IsCitedBy" }
   class RelatedObject
-    attr_accessor :related_identifier, :related_identifier_type, :relation_type
+    attr_accessor :related_identifier, :related_identifier_type, :relation_type, :errors
 
     def initialize(related_identifier:, related_identifier_type:, relation_type:)
       @related_identifier = related_identifier
       @related_identifier_type = related_identifier_type
       @relation_type = relation_type
+      @errors = []
+    end
+
+    def valid?
+      valid = related_identifier.present? && valid_related_identifier_type? && valid_relation_type?
+      return valid if valid
+      if related_identifier.blank?
+        errors << "Related identifier is missing"
+      else
+        errors << "Related Identifier Type is missing or invalid for #{related_identifier}" unless valid_related_identifier_type?
+        errors << "Relationship Type is missing or invalid for #{related_identifier}" unless valid_relation_type?
+      end
+      false
     end
 
     def value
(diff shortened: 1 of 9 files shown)


app/models/pdc_metadata/related_object.rb
```diff
@@ -5,12 +5,25 @@ module PDCMetadata
...
+    end
+
+    def valid?
+      valid = related_identifier.present? && valid_related_identifier_type? && valid_relation_type?
+      return valid if valid
+      if related_identifier.blank?
+        errors << "Related identifier is missing"
+      else
+        errors << "Related Identifier Type is missing or invalid for #{related_identifier}" unless valid_related_identifier_type?
+        errors << "Relationship Type is missing or invalid for #{related_identifier}" unless valid_relation_type?
```
mccalluc, line 24:
```suggestion
        @errors << "Related identifier is missing"
      else
        @errors << "Related Identifier Type is missing or invalid for #{related_identifier}" unless valid_related_identifier_type?
        @errors << "Relationship Type is missing or invalid for #{related_identifier}" unless valid_relation_type?
```

I might be missing something, but wouldn't plain `errors` just be a local variable in the method?

hectorcorrea (approved):

app/models/pdc_metadata/related_object.rb
```diff
@@ -36,5 +49,23 @@ def self.relation_type_options
...
+        @valid_relation_type ||= valid_relationship_types.include?(relation_type&.upcase)
+      end
+
+      def valid_type_values
+        @valid_type_values ||= Datacite::Mapping::RelatedIdentifierType.to_a.map(&:value).map(&:upcase)
+      end
+
+      def valid_relationship_types
+        @valid_relationship_types ||= Datacite::Mapping::RelationType.to_a.map(&:value).map(&:upcase)
+      end
```
mccalluc, line 69:
```suggestion
      def valid_related_identifier_type?
        valid_type_values.include?(related_identifier_type&.upcase)
      end

      def valid_relation_type?
        valid_relationship_types.include?(relation_type&.upcase)
      end

      def valid_type_values
        Datacite::Mapping::RelatedIdentifierType.to_a.map(&:value).map(&:upcase)
      end

      def valid_relationship_types
        Datacite::Mapping::RelationType.to_a.map(&:value).map(&:upcase)
      end
```

I don't know the bigger picture, but is the idea just to store the result to avoid recomputation? Unless these are really expensive operations, I'm not sure the additional complexity is worth it.

mccalluc (changes requested):
I think the `@` is needed... but then is this path not exercised by the tests? I'm probably confused.

app/models/pdc_metadata/related_object.rb
carolyncole: Not needed because of the attr_accessor

PR #608, opened 2022-11-15
Updating yarn lock file

carolyncole:
This keeps occurring on my machine.  Not sure why, but I thought I would open a PR to discuss.  I am utilizing yarn 1.22.19

I believe @Twade968 is also seeing similar changes.  It might be due to the M1, or possibly due to yarn versions.


carolyncole:
@kelynch tested and this update occurs when a yarn install is run with version 1.22.19 and some additional versions. 

We believe this is safe to ignore until @mccalluc gets his new machine.  Although I wonder if our .tool-versions should include yarn so we are utilizing a standard version

carolyncole:
Never mind we do have a yarn version in the .tool-versions.  I just did not utilize asdf to install it...

carolyncole:
superseded by https://github.com/pulibrary/pdc_describe/pull/611

PR #609, opened 2022-11-15
Fixes bug fetching user's own works

### LINKED ISSUE #389: When I administer a collection I can not see works in other collections on my dashboard
As a collection administrator, and a depositor I would like to see my works AND other people's works on my dashboard.

If the works that I submitted are in a collection I do not administer, they disappear from my dashboard.

I currently have 10 works locally, 9 of which are unfinished:
```
[27] pry(main)> Work.unfinished_works(me).count
=> 9
[28] pry(main)> me.add_role("collection_admin", Collection.find(4))
[29] pry(main)> Work.unfinished_works(me).count
=> 2
```

carolyncole:
Fixes bug in code to fetch works by state to always include the user own works (even if they are an administrator)

Big thanks to @carolyncole for providing the initial test that surfaced the bug.

Fixes #389

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -116,10 +116,11 @@ class Work < ApplicationRecord
 
       def works_by_user_state(user, state)
         works = []
-        if user.admin_collections.count == 0
-          # Just the user's own works by state
-          works += Work.where(created_by_user_id: user, state: state)
-        else
+
+        # The user's own works by state (if any)
+        works += Work.where(created_by_user_id: user, state: state)
+
+        if user.admin_collections.count > 0
           # The works that match the given state, in all the collections the user can admin
           # (regardless of who created those works)
           user.admin_collections.each do |collection|
@@ -133,7 +134,7 @@ class Work < ApplicationRecord
           works << work unless already_included
         end
 
-        works.sort_by(&:updated_at).reverse
+        works.uniq(&:id).sort_by(&:updated_at).reverse
       end
 
       # Returns an array of work ids where a particular user has been mentioned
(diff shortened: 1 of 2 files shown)


app/models/work.rb
```diff
@@ -116,10 +116,11 @@ def publish_test_doi?
 
       def works_by_user_state(user, state)
         works = []
-        if user.admin_collections.count == 0
-          # Just the user's own works by state
-          works += Work.where(created_by_user_id: user, state: state)
-        else
+
+        # The user's own works by state (if any)
+        works += Work.where(created_by_user_id: user, state: state)
```
carolyncole, line 121: Instead of a `+=` here, could you combine line. 118 and 121 (Since it is now happening every time)
```suggestion
        works = Work.where(created_by_user_id: user, state: state)
```

carolyncole (commented):
@hectorcorrea Since I created this PR, I can not approve or request changes, but I can comment.

Other than the requested change it looks good to me...

app/models/work.rb
hectorcorrea: Funny. I tried that first, but turns out `Work.where` returns an `ActiveRelation` that looks like an array but it is not an array. Since it is not an array the code blows up a few lines down when it does `works << work unless already_included`. I could probably come up with a clever way to handle this but I thought the simplest and cleanest was to use a plain array and append to it.

carolyncole: Oh!  You just need a `to_a` on the end of the query.

bess (approved):

PR #613, opened 2022-11-16
Refactor XML Validation

### LINKED ISSUE #604: Refactor DataCite XML validation
The code for DataCite XML validation currently lives in a controller. Instead, it should live in the DataCite serialization model, since schema validation is a core behavior of that XML serialization. 

Validation against the official DataCite 4.4 schema happens in the `datacite-mapping` gem, but we also need to validate against our local fork of the schema because we are starting to diverge in several ways (e.g., we require description). We will need a way to audit our records against our local DataCite requirements. 

### Acceptance criteria
- [ ] Refactor this method so it is easier to test and call, and move it to `app/models/pdc_serialization/datacite.rb`: https://github.com/pulibrary/pdc_describe/blob/28b4f3c664b654b57906b0c9e1a20a4894e40798/app/controllers/works_controller.rb#L231-L242

bess:
Ensure the datacite record produced validates against our local copy of the datacite schema.
 This will allow us to evolve our local datacite standards and test our records against them.

Fixes #604

### DIFF
--- a/app/models/pdc_serialization/datacite.rb
+++ b/app/models/pdc_serialization/datacite.rb
@@ -22,16 +22,35 @@ module PDCSerialization
   #
   # rubocop:disable Metrics/ClassLength
   class Datacite
-    attr_reader :mapping
+    attr_reader :mapping, :errors
 
     def initialize(mapping)
       @mapping = mapping
+      @errors = []
     end
 
     # Returns the XML serialization for the Datacite record
     # Note that the actual XML serialization is done by the datacite-mapping gem.
     def to_xml
       @mapping.write_xml
+    end
+
+    # Validate this DataCite XML serialization against the official DataCite schema
+    # By default we validate against DataCite 4.4. This will shift over time as new
+    # versions of the datacite schema are released.
+    # @return [Boolean]
+    def valid?
+      @errors = []
+      datacite_xml = Nokogiri::XML(to_xml)
+      schema_location = Rails.root.join("config", "schema")
+      Dir.chdir(schema_location) do
+        xsd = Nokogiri::XML::Schema(File.read("datacite_4_4.xsd"))
+        xsd.validate(datacite_xml).each do |error|
+          @errors << error
+        end
+      end
+      return true if @errors.empty?
+      false
     end
 
     ##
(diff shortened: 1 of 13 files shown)


app/models/pdc_serialization/datacite.rb
```diff
@@ -34,6 +35,24 @@ def to_xml
...
+    # @return [Boolean]
+    def valid?
+      @errors = []
+      datacite_xml = Nokogiri::XML(to_xml)
+      schema_location = Rails.root.join("config", "schema")
+      Dir.chdir(schema_location) do
+        xsd = Nokogiri::XML::Schema(File.read("datacite_4_4.xsd"))
+        xsd.validate(datacite_xml).each do |error|
+          @errors << error
+        end
```
mccalluc, line 50: Assuming `validate` is an iterator, could you just call `to_a`?

app/models/pdc_serialization/datacite.rb
```diff
@@ -34,6 +35,24 @@ def to_xml
...
+      end
+      return true if @errors.empty?
+      false
```
mccalluc, line 53:
```suggestion
      return @errors.empty?
```

spec/models/pdc_serialization/datacite_spec.rb
```diff
@@ -246,4 +246,15 @@
...
+
+  context "schema validation" do
+    let(:resource) { described_class.new_from_work(work) }
+    context "valid XML record" do
+      let(:work) { FactoryBot.create(:distinct_cytoskeletal_proteins_work) }
+      it "easily validates against our local datacite schema" do
+        expect(resource.valid?).to eq true
+        expect(resource.errors.empty?).to eq true
+      end
+    end
```
mccalluc, line 258: Can you add a negative test as well?

mccalluc (commented):
I think it'd be good to add a test that fails validation as well. Could be something as simple as a document with the wrong root element.

bess:
I agree, and I will do that when we get to the point where we can create an invalid record. Right now, we are using the CDL datacite-mapping gem to make the records, and they literally cannot create an invalid one. They will throw an error if they try. However, as we move forward and encode our local practices into our local copy of the schema, then we will start being able to observe places where we have datacite valid records that are not valid according to our local practices.

mccalluc (approved):
ok then! sorry to slow things down.

Issue #605, opened 2022-11-14
Move migration specs to their own directory


bess:
### Acceptance criteria
- [ ] Move all of the migration specs to their own subdirectory, called `data_migration`

PR #616, opened 2022-11-18
 Moved migration specs to its own directory 

Twade968:
Moved flume, cytoskeletal & femtosecond migration spec test to another folder. ticket #605


mccalluc (commented):
This much is fine, but I'm wondering if you should go ahead and move all the specs? Or maybe the others are somehow different?

I think it might also be simpler to define a rule to ignore this whole directory, so it doesn't need to be updated file-by-file.

bess (approved):

PR #623, opened 2022-11-28
Moved migration spec to its own folder

Twade968:
Moved form submission spec to another folder ticket #605


mccalluc (approved):
Is this spec a duplicate of [sowingseeds_spec.rb](https://github.com/pulibrary/pdc_describe/blob/main/spec/system/sowingseeds_spec.rb)? If it plays a different role, feel free to merge... but if they both cover the same territory, this one could probably be removed.

(The file name is like the others, but short one word.)

Twade968:
@mccalluc thank you for pointing this out. Bess and I will address this when its time for the full fledged data migration to start.

Issue #603, opened 2022-11-14
Validate XML produced by migration specs


bess:
In order to ensure that our migration specs are producing valid datacite records, add a step at the end of each one that validates the datacite produced against the DataCite schema. 

There is an example to follow here: https://github.com/pulibrary/pdc_describe/blob/main/spec/system/attention_form_submission_spec.rb#L56-L59

### Acceptance Criteria
- [x] Add a step at the end of each migration spec that validates the XML

PR #629, opened 2022-11-29
validate bitklavierimage xml datacite schema

Twade968:
Validating bitklavierimage xml against local datacite schema. ticket #603


mccalluc (approved):
It shouldn't block this PR, but down the road we'll want to avoid duplicating code with copy-and-paste. 
- We want to make sure that every case gets the same checks.
- If the check needs to change, we want to make that change in just one location.

As long as that's on your radar, this doesn't need to block.

PR #635, opened 2022-11-30
store gems in project

mccalluc:
As I pick up `pdc_describe` I’m hitting the incompatible library version error again. When I asked about this before, the advice was to make sure I prefix every command with bundle, and I really believe I’m doing that.

It fixes the problem if I gem uninstall each problematic gem, and then bundle install. (I need to repeat this for 6 gems.) When I do this, the version of the freshly installed gem matches what I uninstalled…. but if I copy the gem to /tmp , uninstall, reinstall, and then diff:
```
diff -r ~/.gem/gems/byebug-11.1.3 /tmp/byebug-11.1.3
```
Even though the gem version numbers match, there are significant differences between the version installed under Ruby 3.1.0 (for `tiger-data-app`) and Ruby 3.0.3 (for `pdc_describe`), and this seems consistent with the error message.

There really seems to be a problem with globally installing gems and different Ruby versions. It seems to resolve this problem if I install in the project rather than rely on the global install, and I think this makes a good default.

(If we wanted to leave this to user discretion, we could add to the README an option to run `bundle config set --local path 'vendor/bundle'`... but seems better to just do the more robust thing by default.)


bess (changes requested):
@mccalluc I see you encountering a legitimate difficulty here, but I'd like us to think of a different solution. Storing the gems in the project would be a significant departure from how all of our other projects work and I'm concerned it might lead to deployment issues.

mccalluc:
I've asked on the devs channel if there are other approaches to this problem. So that the other work that's on this branch isn't blocked, I'll revert the bundle changes.

mccalluc:
Last failure was a timeout... Network dependencies during tests aren't ideal, but not something this PR should address. 
```
  1) Form submission for a legacy dataset happy path produces and saves a valid datacite record
     Failure/Error: bl.call(klass)
     
     Net::OpenTimeout:
       execution expired
```

I'll update branch and re-run tests.

mccalluc:
huh: Same error after latest merge. Re-running...

bess:
@mccalluc I think we want to close this PR, since we decided not to store the gems in the project, right? Is there any work from this branch you want to put into a separate PR?

mccalluc:
Sorry for the confusion around this: I've reverted the bundle changes at the tip of this branch, but I'm guessing you're prefer to have separate PRs for each feature, so I'll make need PRs to main that recapitulate these:
- #636
- #646
- #644 (and for this one, I'll separate the refactoring from the the new work.)

I'll switch this PR to draft to take the spotlight off it, and then I'll close when the new analogous PRs have been merged.

mccalluc:
New PRs from main have been made to cover everything in here. Closing!

PR #636, opened 2022-11-30
Datestamps in history

### LINKED ISSUE #632: Describe Form: Method to record time in the provenance log
## Description

The provenance log feature is important for capturing the actions for all users that change the record in some way. The current method for recording the time for each change or set of changes would become confusing as time goes on. The current method uses hours, days, weeks, months, (and presumably years) past. We would like this changed so that looking back on this in the future, it's easy to determine when the actions occurred. Using a timestamp in place of the current method is one possible solution, but others may be suggested.

## Image shows time is currently recorded as hours, days, weeks, months, (years?)
Screen Shot 2022-11-21 at 4 33 08 PM

## Acceptance Criteria
- [x] Log time is written date followed by timestamp. For Example: November 01, 2022 13:40. (Please see image in comment below.)
- [x] Log time is recorded after the username, which is the same order as the current method.

mccalluc:
- Fix #632
- Add "Unknown user outside the system" failover everywhere
- Factor copy-and-paste code into partial
- Rename partials for clarity

In the test, I don't really understand why the UTC time is offset: I hope that this test doesn't depend on the system timezone.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -25,5 +25,9 @@ module DrdsRailsTemplate
       port: ENV.fetch("APPLICATION_PORT", "3000"),
       protocol: ENV.fetch("APPLICATION_HOST_PROTOCOL", "http")
     }
+
+    # Explicitly set timezome rather than relying on system,
+    # which may be different in CI environment.
+    config.time_zone = 'America/New_York'
   end
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -28,6 +28,6 @@ module DrdsRailsTemplate
 
     # Explicitly set timezome rather than relying on system,
     # which may be different in CI environment.
-    config.time_zone = 'America/New_York'
+    config.time_zone = "America/New_York"
   end
 end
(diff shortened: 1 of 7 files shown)


mccalluc:
Hoping that CI will be happy with the timezone, but if there's a better way to manage this, happy to change direction!

mccalluc:
Strange: Previous failure was in an unrelated test suite. Will try rerunning...
```
  1) WorksController valid user login when only some file uploads are replaced for an existing Work with uploads handles the update page
     Failure/Error: expect(saved_work.pre_curation_uploads[0].blob.filename.to_s).to eq(File.basename(temp_file2.path))
     
       expected: "temp_file220221201-1082-vmul8h"
            got: "us_covid_2019.csv"
```

config/application.rb
```diff
@@ -25,5 +25,9 @@ class Application < Rails::Application
       port: ENV.fetch("APPLICATION_PORT", "3000"),
       protocol: ENV.fetch("APPLICATION_HOST_PROTOCOL", "http")
     }
+
+    # Explicitly set timezome rather than relying on system,
+    # which may be different in CI environment.
+    config.time_zone = "America/New_York"
```
jrgriffiniii, line 31: 👍

PR #643, opened 2022-12-01
Validating flume xml datacite schema

Twade968:
Validating flume xml against local datacite schema. ticket #603


mccalluc (approved):
As with the others, down the road we'll want to make sure we don't copy and paste boilerplate between tests, but that doesn't need to block this.

PR #647, opened 2022-12-02
Disable headers in markdown

### LINKED ISSUE #622: Describe Form: Remove header options from the comment feature
## Description

Headings are a useful means to organize textual content. However, there isn't sufficient space or need to support this in the Describe Form. For example, the screenshot below shows how distracting this feature can be when trying to read the provenance and messaging for this record. The bold and italics are sufficient to organize this information, although we may consider other markdown options for the future.

## Image shows the distracting size of the headers
Screen Shot 2022-11-21 at 4 33 08 PM

## Acceptance Criteria
- [ ] Remove heading options
- [ ] Leave the remaining markdown options, such as **bold** and *italic*

mccalluc:
- Fix #622

I have mixed feelings about this: Every special case we add is something that could be confusing down the line... And from my own experience, sometime h3 or h4 headers are useful in organizing a longer comment.

### DIFF
--- a/app/models/work_activity.rb
+++ b/app/models/work_activity.rb
@@ -59,8 +59,10 @@ class WorkActivity < ApplicationRecord
         user_info = user&.display_name_safe || uid
         "<a class='comment-user-link' title='#{user_info}' href='{USER-PATH-PLACEHOLDER}/#{uid}'>#{at_uid}</a>"
       end
-      # allow ``` for code blocks (Kramdown only supports ~~~)
+      # Allow ``` for code blocks (Kramdown only supports ~~~)
       text = text.gsub("```", "~~~")
+      # Escape line-starting hash marks: Headers are distracting.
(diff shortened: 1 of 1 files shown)


hhkitsune:
@mccalluc I don't disagree that smaller headers (h3 or h4) might be useful. However, h1 and h2 create a visual distraction in our log that will make it harder to quickly find the change content, as the present design of this feature favors showing the message content. I would have liked to have messages and changes separated. I keep pointing to the example of provenance used for PSU ScholarSphere to understand what I want to capture. That example shows a clean, standardized timeline of changes to the record. The messaging feature, which is also important, somehow got blended. Perhaps it's something we can talk about separating in the future.

mccalluc:
Bess points out that this is better solved with CSS. Closing this PR, will clarify approach in the orig issue. For the rest, we should talk in slack or at the next meeting. Thanks!

Issue #660, opened 2022-12-05
Describe Form: Description property help text


hhkitsune:
## Description
We are adding help text for the description property. This property is located on the required metadata tab page.

The help text will appear in regular text (not bolded) below the title of the property set.

### For Example:

**Description**
Please enter a brief summary of this item specifically -- not identical to the abstract of a corresponding paper. The focus should be on the nature of the materials constituting the item (scope, purpose, methods, etc.), as opposed to substantive claims made in related publications.

**Note:** -- should be an em-dash

## Text to use

Please enter a brief summary of this item specifically -- not identical to the abstract of a corresponding paper. The focus should be on the nature of the materials constituting the item (scope, purpose, methods, etc.), as opposed to substantive claims made in related publications.

## Acceptance Criteria
- [ ] Help text (provided in this ticket) will be added to the title property on the required metadata tab.

mccalluc:
There's an earlier issue to add hint text:
- #483

Are both of these needed? Or can we get by with just one or the other?

hhkitsune:
@mccalluc (also: @hectorcorrea @bess ) The ticket for help text within the boxes (rather than without) was from back in October. Upon implementation of a few of these, and after some changes were made to the anticipated size of certain property fields, there is an obvious visual problem that help text is cut off for some fields. I wouldn't consider ticket #483 is no longer needed, but we do need to reassess it. The most recent tickets that have help text outside of the property field boxes should have priority. These are requests coming directly from Neggin and are needed for the curation service to be successful once we switch from DataSpace. I'm available to talk about this with any of you if you need more information.

Issue #658, opened 2022-12-05
Describe Form: Title property help text


hhkitsune:
## Description

We are adding help text for the Title property. Please use the text provided in this ticket.

The help text will appear in regular text (not bolded) below the title of the property. This will show on both pages containing the title property: 1) The initial entry page for a new deposit 2) required metadata tab.

### For example:

**Title**
Provide an informative and distinctive title for the item as a whole -- not identical to that of a corresponding paper.

## Text to use

Provide an informative and distinctive title for the item as a whole -- not identical to that of a corresponding paper.

## Acceptance Criteria
- [ ] Help text (provided in this ticket) will be added to the title property on both the new submission page and the required metadata tab.

mccalluc:
There's an earlier issue to add hint text:
- #483

Are both of these needed? Or can we get by with just one or the other?

mccalluc:
Wait for feedback on 
- #672

before implementing this.

Issue #483, opened 2022-10-12
Describe: Hint text within input boxes


hhkitsune:
# Description
Hint text within input boxes are expected to help guide the depositor towards inputting the correct information. This ticket describes which fields should have hint text within the associated box, and exactly what text to include. These are organized by the tab section below. Each section will be followed by a wireframe that illustrates the hint text.

# Properties with Hint Text

### Title

**Hint Text:** The title should be unique and different from one used for a corresponding paper

### Description

**Hint Text:** Please enter a brief summary describing the unique attributes of this submission. (e.g., The purpose, scope, methods, dependencies, etc. for the object(s) deposited to this submission.) It should be unique from a description or abstract used for a corresponding paper.

![Screen Shot 2022-10-11 at 11 49 01 AM](https://user-images.githubusercontent.com/50670293/195354627-1795d28c-7575-4478-868b-61af11d6b0e2.jpeg)

### Funder Name

**Hint Text:** The name of the funder. (Example: National Science Foundation)

### Award Number

**Hint Text:** The grant number or funding code for your award

### Award URI 

**Hint Text:** The URL web address for your award

### Related Identifier

**Hint Text:** The URL web address for a related publication or other resource

### Keywords

**Hint Text:** Three or more unique keywords separated by commas that assist discovery. 

![Screen Shot 2022-10-12 at 3 50 51 PM](https://user-images.githubusercontent.com/50670293/195435177-a33c9ab3-3a1d-4b9e-be6d-0d5f170bcae0.jpeg)

![Screen Shot 2022-10-11 at 11 49 24 AM](https://user-images.githubusercontent.com/50670293/195354751-05071a5a-a954-4625-84d0-7a81231d52f6.jpeg)

### Collection Tags

**Hint Text:** Collection(s) the object belongs to separated by commas

### Related Item

**Hint Text:** Citation for a related resource that does not have a URI

![Screen Shot 2022-10-11 at 11 49 44 AM](https://user-images.githubusercontent.com/50670293/195354789-b15d430c-1c26-425a-ba8b-7da99d6fcbaa.jpeg)

# Acceptance Criteria
For convenience, this has been broken into the same tabbed sections as above. Hint text will be added to the following properties:

## Required Metadata Tab
- [ ] **Title** hint text has been added
- [ ] **Description** hint text has been added

## Additional Metadata Tab
- [ ] **Funder Name** hint text has been added
- [ ] **Award Number** hint text has been added
[...]

PR #663, opened 2022-12-05
Add placeholders to fields

mccalluc:
Towards
- #483

Many of the requested fields have not been implemented, so this probably shouldn't close that issue?

I'm also not sure that this is the best way to provide additional information to the users doing the data entry: As soon as they start to type, the placeholder goes away.

> Note: The placeholder attribute is not as semantically useful as other ways to explain your form, and can cause unexpected technical issues with your content. See [Labels](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#labels) for more information.

from: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#placeholder

Also see more recently filed issues in the same area:
- #660 
- #658


hectorcorrea:
I think the `placeholder` attribute is a good place to start.

PR #679, opened 2022-12-07
Fix Draft validation to not allow nil collections

### LINKED ISSUE #631: Describe Form: Error message for record selection
## Description

**This happened to more than one person testing PDC Describe.**

An error screen appears when attempting to click on a record to open that record. This doesn't seem to happen for all records. The two separate instances occurred with different users for (test) records they recently created.  

## Image shows user tried to access "Purple Haze.." record.
Screen Shot 2022-11-29 at 4 18 16 PM

## Image shows the error screen after selecting the record from the home screen list
Screen Shot 2022-11-29 at 4 18 23 PM

## Acceptance Criteria
- [ ] The record will open when selected from the home screen list

carolyncole:
Collection validation was being lost when the work was in a draft state due to the clearing of errors. Also add a check of parameters to not allow a nil collection coming from the form fixes #631

This may have an impact of if old test data on staging can be saved.  I ran into an instance where there was no version in addition to no collection, which caused me to not be able to save the work.

I wonder if clearing old data from staging might not be a good step prior to deploying this.


app/controllers/works_controller.rb
```diff
@@ -332,5 +332,17 @@ def process_updates
...
       end
     end
+
+    def params_collection_id
+      # Do not allow a nil for the collection id
+      @params_collection_id ||= begin
+        collection_id = params[:collection_id]
+        if collection_id.blank?
+          collection_id = current_user.default_collection.id
+          Honeybadger.notify("We got a nil collection as part of the parameters #{params} #{request}")
```
mccalluc, line 342: Am I understanding this correctly that it's an error state that we want to be notified about, but that it also shouldn't happen?

mccalluc (commented):
This looks good, but I think it would be good to also have review from someone who is more familiar with the code.

app/controllers/works_controller.rb
carolyncole: Yes, that is correct.  I am unsure how it occurred, so I want to log if it ever happens again...

bess (approved):

PR #680, opened 2022-12-07
Server-side publish files to post-curation buckets

### LINKED ISSUE #538: Utilize s3 copy_to instead of downloading and uploading
Currently the code downloads and then uploads the S3 object to transfer from pre-curation to post-curation.

This will be too costly for large files. Instead we should utilize the S3 copy_to, which does it all in Amazon in one step.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/example_s3_CopyObject_section.html

Here is the line that does a download and an upload: https://github.com/pulibrary/pdc_describe/blob/13ea421954f969fa735e61c88c79bba1638df337/app/models/work.rb#L719

Acceptance criteria:

- [ ] The copy to api is utilized in the work (see example link above)
- [ ] There is an automated to make sure we are not doing a get and put on amazon S3

hectorcorrea:
Changed code to copy of files from pre-curation to post-curation S3 buckets to use Amazon's own copy method so that files are copied between S3 buckets rather than require a server download and re-upload.

Closes #538

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -420,20 +420,6 @@ class Work < ApplicationRecord
 
   delegate :bucket_name, to: :s3_query_service
 
-  # Transmit a PUT request to upload an ActiveStorage Blob file to the post-curation S3 Bucket
-  # @param blob [ActiveStorage::Blob]
-  # @param bucket_name [String]
-  # @return [Aws::S3::Types::PutObjectOutput]
-  def put_post_curation_s3_object(bucket_name:, blob:)
-    s3_client.put_object({
-                           bucket: bucket_name,
-                           key: blob.key,
-                           body: blob.download
-                         })
-  rescue Aws::S3::Errors::ServiceError => error
-    raise(StandardError, "Failed to upload the following to the post-curation S3 Bucket object #{blob.key}: #{error}")
-  end
-
   # Transmit a HEAD request for an S3 Object in the post-curation Bucket
   # @param key [String]
   # @param bucket_name [String]
@@ -521,11 +507,6 @@ class Work < ApplicationRecord
       blob = ActiveStorage::Blob.create_before_direct_upload!(**params)
       blob.key = s3_file.filename
       blob
-    end
-
-    def post_curation_s3_file?(filename:)
-      post_curation_uploads_keys = post_curation_uploads.map(&:filename)
-      post_curation_uploads_keys.include?(filename)
     end
 
     def publish(user)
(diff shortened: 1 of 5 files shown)


app/models/work.rb
```diff
@@ -712,28 +693,24 @@ def publish_precurated_files
...
       s3_dir = find_post_curation_s3_dir(bucket_name: s3_post_curation_query_service.bucket_name)
       raise(StandardError, "Attempting to publish a Work with an existing S3 Bucket directory for: #{s3_object_key}") unless s3_dir.nil?
 
-      pre_curation_uploads.each do |attachment|
-        next if post_curation_s3_file?(filename: attachment.filename.to_s)
+      # Copy the pre-curation S3 Objects to the post-curation S3 Bucket...
+      transferred_files = s3_post_curation_query_service.publish_files
 
-        put_post_curation_s3_object(bucket_name: s3_post_curation_query_service.bucket_name, blob: attachment.blob)
+      # ...check that the files are indeed now in the post-curation bucket...
```
hectorcorrea, line 706: We might *not* need to check that the files were transferred since we are letting AWS do this for us, we used to do the transfer manually so the check made sense. I left the check for now since it does not really hurt and it probably won't slow things down.

spec/support/work_s3_requests_specs.rb
```diff
@@ -23,6 +23,12 @@ def build_s3_list_objects_response(work:, file_name:)
 
   # rubocop:disable Metrics/AbcSize
   def stub_work_s3_requests(works: [], work: nil, file_name: nil)
+    # Use the built-in AWS S3 stub
+    # (https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/stubbing.html)
+    s3 = Aws::S3::Client.new(stub_responses: true)
+    allow(Aws::S3::Client).to receive(:new).and_return(s3)
+    s3.stub_responses(:head_object, [Aws::S3::Errors::NotFound.new("1", "2"), true])
+
```
hectorcorrea, line 31: This is the new stubbing referenced in https://github.com/pulibrary/pdc_describe/issues/700

app/services/s3_query_service.rb
```diff
@@ -175,6 +175,27 @@ def data_profile
     { objects: [], ok: false }
   end
 
+  ##
+  # Copies the existing files from the pre-curation bucket to the post-curation bucket.
+  # Notice that the copy process happens at AWS (i.e. the files are not downloaded and re-uploaded).
+  # Returns an array with the files that were copied.
+  def publish_files
```
jrgriffiniii, line 182: 👍

jrgriffiniii (approved):

PR #685, opened 2022-12-08
Connect WorkActivity to WorkActivityNotifications so we have a cleaner deletion process

carolyncole:
This allows a user to run `work_activity.destroy` and have the dependent WorkActivityNotifications also destroyed.

Without this, you need to run  `WorkActivityNotification.where(work_activity_id: work_activity.id).each(&:destroy)` before running the `work_activity.destroy`

Otherwise you will get:
```
ActiveRecord::InvalidForeignKey:
       PG::ForeignKeyViolation: ERROR:  update or delete on table "work_activities" violates foreign key constraint "fk_rails_2a0cd9ebed" on table "work_activity_notifications"
       DETAIL:  Key (id)=(3) is still referenced from table "work_activity_notifications".
```

### DIFF
--- a/app/models/work_activity.rb
+++ b/app/models/work_activity.rb
@@ -1,7 +1,10 @@
 # frozen_string_literal: true
 
+# rubocop:disable Metrics/ClassLength
 class WorkActivity < ApplicationRecord
   belongs_to :work
+
+  has_many :work_activity_notifications, dependent: :destroy
 
   USER_REFERENCE = /@[\w]*/.freeze # e.g. @xy123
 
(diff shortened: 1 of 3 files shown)


app/models/work_activity.rb
```diff
@@ -1,8 +1,11 @@
 # frozen_string_literal: true
 
+# rubocop:disable Metrics/ClassLength
```
carolyncole, line 3: It was right at 100 lines so adding my 1 line took it over the top.

mccalluc (approved):

Issue #682, opened 2022-12-08
How does our Active Storage connection get configured?


carolyncole:
We utilize ActiveStorage: https://edgeguides.rubyonrails.org/active_storage_overview.html

Which can be configured:
* https://github.com/pulibrary/pdc_describe/blob/main/config/environments/test.rb#L43
* https://github.com/pulibrary/pdc_describe/blob/main/config/storage.yml

Questions:
  - [ ] How is this being utilized in our system?  
  - [ ] Should it be in both places?  
  - [ ] Can we simplify?

Acceptance:
- [ ] Questions should be documented in the Readme once we understand

mccalluc:
In the meeting this morning decided that we should:

- Check if publication even works with local storage.
- Explain on the README – if you’re developing features that use s3, s3 needs to be enabled.

PR #689, opened 2022-12-08
Fix URI bug when editting a work in a local development environment

### LINKED ISSUE #662: `URI::InvalidURIError in Works#edit`
This same error message was part of 
- #519

That is marked as fixed, but it's possible this is a regression.

To reproduce:
- Create a new record: On the first page of the wizard, just a title, and "x", "y" as the creator, "asdf" description, and pick a license.
- Hit the "Save work" button.
- Select the default: "under 100GB and I want to upload the files directly."
- create a zero length file locally and upload.
- Hit "Complete"
- From the list of all submissions, click on the one you've just created.
- and then click "Edit"

I get the following error:
```
URI::InvalidURIError in Works#edit
Showing /Users/chuck/github/rdss/pdc_describe/app/views/works/_uploads_form.html.erb where line #31 raised:
bad URI(is not URI?): nil
```
The error occurs on this line:
```
<a href="<%=upload.url%>" class="documents-file-link" target="_blank" title="<%= upload.key %>"><%= upload.filename %></a>
```
 In the console we can narrow down the problem:
```
>> upload.url
URI::InvalidURIError: bad URI(is not URI?): nil
	from /Users/chuck/github/rdss/pdc_describe/app/views/works/_uploads_form.html.erb:31:in `block in _app_views_works__uploads_form_html_erb___1404918409379364812_148280'
```
The `url` method is [documented here](https://api.rubyonrails.org/classes/ActiveStorage/Blob.html#method-i-url).

- Can other devs reproduce this?
- Does it still happen with a non-zero-length file?

mccalluc:
- Seems to fix #662: I can now edit a work. [Rails docs](https://edgeapi.rubyonrails.org/classes/ActiveStorage/SetCurrent.html).
- Leaving #682 open: The configuration process confuses me, [particularly in development.rb](https://github.com/pulibrary/pdc_describe/pull/301/files?diff=unified&w=1#diff-d3c4b3f41072daa416f1920511e9b2e26caea8c5cec0a14cb9508589a4dafa47):
```diff
-  config.active_storage.service = :local
+  config.active_storage.service = :amazon
```

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 class ApplicationController < ActionController::Base
+  include ActiveStorage::SetCurrent
+
   protect_from_forgery with: :exception
   before_action :authenticate_user!
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 class ApplicationController < ActionController::Base
+  # This is necessary only for localhost development, with storage configured for the filesystem,
+  # but it shouldn't cause problems in other environments that use S3.
+  # Including this concern lets the disk service generate URLs using
+  # the same host, protocol, and port as the current request.
   include ActiveStorage::SetCurrent
 
   protect_from_forgery with: :exception


app/controllers/application_controller.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 class ApplicationController < ActionController::Base
+  include ActiveStorage::SetCurrent
```
carolyncole, line 3: Since this is entirely untested, it would be nice to have some comments around this in the code.  If we remove it the tests all still pass, so someone might be tempted to revert this commit.

carolyncole (changes requested):

carolyncole (approved):

PR #691, opened 2022-12-08
Remove Title placeholder and add details folddown

### LINKED ISSUE #658: Describe Form: Title property help text

mccalluc:
- Fix #658

Removing the placeholder not explicitly requested, but since the text is much the same, and the detail fold-down seems to be preferred, thought it would be ok to delete.

### DIFF
--- a/app/views/works/_required_title.html.erb
+++ b/app/views/works/_required_title.html.erb
@@ -1,13 +1,16 @@
 <!-- Render the main title (notice that we don't display a dropdown for this title) -->
 <div class="field">
-  Title:<span class="required-field" title="Title for your submission must be indicated">*</span>
-  <span id="title-required-message" class="hidden">
-    <i class="bi bi-exclamation-diamond-fill required-field"></i>&nbsp;Must provide a title
-  </span>
-  <br>
+  <details>
+    <summary>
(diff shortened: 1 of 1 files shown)


carolyncole:
I'm not sure I love the triangle.  If I wanted help on what the field was I would not think to click on a triangle.  I'll add this to the agenda so we can discuss...
Screen Shot 2022-12-09 at 9 08 30 AM

mccalluc:
Good point. I've filed a separate issue for this:
- #694

Since this is not the first use of the fold-down triangle, I'd prefer for this PR not to block on the styling change.

carolyncole (approved):

PR #692, opened 2022-12-08
Separate comments from other changes

### LINKED ISSUE #677: Separate message log from change log
- Message log should come above change log on the page
- On messages, who sent it is most important; on changes, when it happened is most important

mccalluc:
- Fix #677
- Change "Activity History" to "Change History" for clarity
- Remove view toggle, since they are separate sections.
- Add a "No comments" / "No changes" message if needed, rather than hiding the entire section. (With more sections, I think conditional display can get more confusing.)

Screen Shot 2022-12-08 at 6 16 11 PM


carolyncole:
I would think having the toggle would still be a good thing. Maybe toggle the change history closed by default.  I'm not sure I would be inclined to remove it.

bess (approved):

Issue #645, opened 2022-12-01
When works attribute is changed, don't show "removed" and "added" in prov log


mccalluc:
We're saying work related objects / creators / contributors are "removed" and "added", when we should be able to just say they're "changed" 
- There is logic for "changed", but it might not be working as intended?
- I also wonder if we might be better off with an off-the-shelf diff library, rather than trying to make our own. https://github.com/samg/diffy is high in the search results... but it might start a subprocess for each diff?! https://rubygems.org/gems/diff-lcs/versions/1.4.4 seems safer.

PR #693, opened 2022-12-09
Diff simple text fields in the change log

mccalluc:
- Towards #645

Want to get feedback on this much before going farther. Right now it's doing a character-by-character diff: If the change was a typo fix, this makes it easy to see... but if the whole string has changed, it will still try to find a minimal set of differences, and the result may be hard to read.

In this PR, the diff is done when the data changes. An alternative approach would be to leave `resource_compare_service.rb` as is and do the diff at render time.

This adds a little CSS, but another option would be to increase [`text-decoration-thickness`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-thickness).
Screen Shot 2022-12-08 at 11 34 25 PM

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -94,3 +94,5 @@ end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
+
+gem "diff-lcs", "~> 1.5"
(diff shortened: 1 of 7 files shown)


mccalluc:
Thinking it over, it will be better to only persist the old and new values... The details of the diffing should be left till render time. Marking as draft.

hectorcorrea:
Awesome idea!

mccalluc:
replaced by #706. closing.

Issue #684, opened 2022-12-08
Clean out staging data


carolyncole:
This is different from production as we need to keep around some test sets:

  https://pdc-describe-staging.princeton.edu/describe/works/112
  https://pdc-describe-staging.princeton.edu/describe/works/110
  https://pdc-describe-staging.princeton.edu/describe/works/69
  https://pdc-describe-staging.princeton.edu/describe/works/113

PR #703, opened 2022-12-09
Making UserWork explicitly related to Work so it is easier to delete a work

carolyncole:
Allows the deletion to occur automatically when a work is deleted.

Refs #684

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -8,6 +8,7 @@ class Work < ApplicationRecord
   class InvalidCollectionError < ::ArgumentError; end
 
   has_many :work_activity, -> { order(updated_at: :desc) }, dependent: :destroy
+  has_many :user_work, -> { order(updated_at: :desc) }, dependent: :destroy
   has_many_attached :pre_curation_uploads, service: :amazon_pre_curation
 
   belongs_to :collection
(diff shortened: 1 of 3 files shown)


spec/models/work_spec.rb
```diff
@@ -955,4 +955,13 @@
       expect(work).not_to be_valid
     end
   end
+
+  describe "delete" do
+    it "cleans up all the related objects" do
+      work.complete_submission!(user)
+      expect { work.destroy }.to change { Work.count }.by(-1)
+                                                      .and change { UserWork.count }.by(-1)
```
carolyncole, line 963: Rubocop did this to the spacing 😢

app/models/work.rb
```diff
@@ -8,6 +8,7 @@ class Work < ApplicationRecord
   class InvalidCollectionError < ::ArgumentError; end
 
   has_many :work_activity, -> { order(updated_at: :desc) }, dependent: :destroy
+  has_many :user_work, -> { order(updated_at: :desc) }, dependent: :destroy
```
jrgriffiniii, line 11: 👍

jrgriffiniii (approved):

PR #706, opened 2022-12-10
Present changes to string fields as a diff

mccalluc:
- Replace #693
- Towards #645

Same note as on the original, only the implementation is more clean now:
> Want to get feedback on this much before going farther. Right now it's doing a character-by-character diff: If the change was a typo fix, this makes it easy to see... but if the whole string has changed, it will still try to find a minimal set of differences, and the result may be hard to read.

I'm pretty sure this is better as a stand-alone class rather than a service or a concern... but I'm not sure where the best place for it to live is: `lib` seems to be one convention. Feedback welcome!


bess (dismissed):
😍

jrgriffiniii (approved):

PR #707, opened 2022-12-12
Adding a rake task to clean out works for a user so we can clean out old tests data from staging & production

### LINKED ISSUE #664: Create a rake task to clear out old user by net id
There may be some old crufty data that the developers have generated.

We would like a way to clean up the staging (or production) by user, not just remove all the data.

We should destroy the user and any items created by that user.

### LINKED ISSUE #684: Clean out staging data

carolyncole:
fixes #664
fixes #684

This allows us the delete works that we create, but not works others create.

### DIFF
--- a/config/storage.yml
+++ b/config/storage.yml
@@ -12,8 +12,8 @@ amazon: &amazon
   region: <%= S3QueryService.pre_curation_config.fetch(:region) %>
   bucket: <%= S3QueryService.pre_curation_config.fetch(:bucket) %>
 
-# development: &development
-#   <<: *amazon
+development: &development
+  <<: *amazon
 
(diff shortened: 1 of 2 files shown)


carolyncole:
@bess This is the code that we ran during our meeting.  It worked correctly.

bess (approved):
⭐ 🎉

PR #708, opened 2022-12-12
In diffs, shorten strings that are 7 or more words

mccalluc:
After the last PR, this was nagging me, but I'm not sure if it's the best course. Feedback welcome.

### DIFF
--- a/app/lib/diff.rb
+++ b/app/lib/diff.rb
@@ -17,8 +17,13 @@ class SimpleDiff
 
   def to_html
     @changes.map do |chunk|
-      old_ellipsis = chunk[:old].gsub(/((?:\S+\s+){3})(.+)((?:\s+\S+){3})/, '\1...\3')
-      new_ellipsis = chunk[:new].gsub(/((?:\S+\s+){3})(.+)((?:\s+\S+){3})/, '\1...\3')
+      too_many_words_re = /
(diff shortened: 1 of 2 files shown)


app/lib/diff.rb
```diff
@@ -17,8 +17,10 @@ def initialize(old_value, new_value)
 
   def to_html
     @changes.map do |chunk|
-      old_html = CGI.escapeHTML(chunk[:old])
-      new_html = CGI.escapeHTML(chunk[:new])
+      old_ellipsis = chunk[:old].gsub(/((?:\S+\s+){3})(.+)((?:\s+\S+){3})/, '\1...\3')
```
hectorcorrea, line 20: It might be a good idea to add a quick description as of what is it that this RegEx does? Not a detailed description, just a short bit giving the next developer a hint as of what it does.

hectorcorrea (dismissed):
Looks great.

spec/lib/diff_spec.rb
```diff
@@ -29,8 +29,8 @@
   end
   it "repeats the entire string" do
```
bess, line 30: Should this heading now read "it does NOT repeat the entire string"?

bess (changes requested):
I love the feature, just requesting a fix to the test heading that's currently confusing.

hectorcorrea (approved):

bess (approved):

PR #709, opened 2022-12-12
Copy button to copy DOI's URL to the clipboard

### LINKED ISSUE #696: Describe Form: Full URL for DOI
## Description

We would like to see the full URL displayed in the Describe Form. This will make it easier for users to copy/paste the DOI info to submit to the publisher or for similar needs.

### Scenario 1
Mario, a researcher, needs to submit his draft DOI. He looks in the form and sees something that isn't a URL. This is confusing and he doesn't know what these strange numbers mean. 

### Scenario 2

Luigi, a researcher, needs to submit their draft DOI. They look in the form and see a URL that can be easily copy/pasted to the email. It's copy/pasted and sent. No problems!

## Acceptance Criteria
- [ ]  Allow a user to copy/paste a full URL of the DOI (note: Implementation method is negotiable, this is just the end result we want to experience.)

hectorcorrea:
Added a button to the DOI display to copy the DOI URL to the clipboard (just like we do in PDC Discovery). 

This is slightly different than what was asked on the original ticket (#696) but I discussed it with @hhkitsune and we thought mimicking PDC Discovery was a good approach.

DOI copy

Closes #696

### DIFF
--- a/app/views/works/show.html.erb
+++ b/app/views/works/show.html.erb
@@ -246,56 +246,11 @@
   <% end %>
 </div>
 
+<%= javascript_include_tag 'show_work_utils' %>
+
 <script>
   $(function() {
-   // Sets the elements to the proper CSS classes once a value has been copied to the clipboard.
-   var setCopiedToClipboard = function(iconEl, labelEl, normalClass, copiedClass) {
-      $(iconEl).removeClass("bi-clipboard");
-      $(iconEl).addClass("bi-clipboard-check");
-      $(labelEl).text("COPIED");
-      $(labelEl).removeClass(normalClass);
-      $(labelEl).addClass(copiedClass);
-    }
-
-    // Resets the elements to the proper CSS classes (e.g. displays as if the copy has not happened)
-    var resetCopyToClipboard = function(iconEl, labelEl, normalClass, copiedClass) {
-      $(labelEl).text("COPY");
-      $(labelEl).removeClass(copiedClass);
-      $(labelEl).addClass(normalClass);
-      $(iconEl).addClass("bi-clipboard");
-      $(iconEl).removeClass("bi-clipboard-check");
-    }
-
-    // Sets icon and label to indicate that an error happened when copying a value to the clipboard
-    var errorCopyToClipboard = function(iconEl, errorMsg) {
-      $(iconEl).removeClass("bi-clipboard");
-      $(iconEl).addClass("bi-clipboard-minus")
-      console.log(errorMsg);
-    }
-
-    // Copies a value to the clipboard and notifies the user
-    // value - value to copy to the clipboard
-    // iconEl - selector for the HTML element with the clipboard icon
-    // labelEl - selector for the HTML element with the COPY label next to the icon
-    // normalClass - CSS to style the label with initially
-    // copiedClass - CSS to style the label with after a value has been copied to the clipboard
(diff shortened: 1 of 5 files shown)


spec/system/work_show_spec.rb
```diff
@@ -16,4 +17,12 @@
     expect(page).to have_content "IS_CITED_BY"
     expect(page).to have_content "https://www.biorxiv.org/content/10.1101/545517v1"
   end
+
+  it "copies DOI to the clipboard" do
+    sign_in user
+    visit work_path(work)
+    expect(page).to have_content "COPY"
+    find("button#copy-doi").click
+    expect(page).to have_content "COPIED"
```
hectorcorrea, line 26: I cannot get this change to be detected on the test. 

The "click" of the button triggers the JavaScript to change the label from "COPY" to "COPIED" and the test passes if I run it with `RUN_IN_BROWSER=true` but not without it.

hectorcorrea: @carolyncole sadly that's not it - the spec in this case already has that setting (https://github.com/pulibrary/pdc_describe/pull/709/files#diff-4c1af586c2edeb41cfdbc5021707f0df6bafd14cecbedc8e52e596e3ab4eaa21R11) 

What is weird about this particular scenario is that the test passes OK if I run in the browser (I can see the old/new text, and the test actually passes), but continues to fails if I don't run it on the browser.

carolyncole: Sorry!  I have defiantly had that same behavior in the past.  I just can not remember what caused it.

mccalluc (changes requested):
My sense was that we wanted to avoid adding more JS to .html.erbs, primarily because it doesn't get linted... but maybe there isn't a consensus on that.

spec/system/work_show_spec.rb
hectorcorrea: I changed this test completely. I am now testing the presence of the HTML `button` tag rather than the JS call. I left a note in the test with the complications in case we later on find a better way to test it.

bess (dismissed):

app/views/works/show.html.erb
```diff
@@ -240,9 +246,11 @@
   <% end %>
 </div>
 
+<%= javascript_include_tag 'show_work_utils' %>
```
carolyncole, line 249: If this is part of the manifest, is it really needed here too?

app/views/works/show.html.erb
```diff
@@ -240,9 +246,11 @@
...
+
 <script>
   $(function() {
-    // Issues HTTP POST to update the curator assigned to the work.
+     // Issues HTTP POST to update the curator assigned to the work.
```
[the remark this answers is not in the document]
hectorcorrea: oops! nope. removed

carolyncole (commented):
Apologies, not sure if I should approve or request changes...

app/views/works/show.html.erb
hectorcorrea: I saw that we did that for the other files and I just did the same (see https://github.com/pulibrary/pdc_describe/blob/main/app/views/works/edit.html.erb#L36 and https://github.com/pulibrary/pdc_describe/blob/main/app/assets/config/manifest.js#L3)

mccalluc (approved):
Approve, but should there be an issue to surface the question of testing? My guess is that `RUN_IN_BROWSER` would slow things down, but maybe it would be useful for a subset of the tests?

PR #712, opened 2022-12-12
Added note to README regarding file storage during development

hectorcorrea:
Documented default storage and how to change it to AWS.

PR #716, opened 2022-12-12
Add note in storage.yml to explain development against AWS

mccalluc:
- Complements #712
- Revert accidental config change from #707


carolyncole (approved):
Thank you!

PR #725, opened 2022-12-13
Updating manifest not to include the javascript for edit

carolyncole:
This seems to do nothing, so I am removing it


carolyncole:
Never mind this breaks things.  I do wonder why it is needed in both places though...

PR #735, opened 2022-12-14
About page (with lorem ipsum text for now)

### LINKED ISSUE #541: Provide route for `about` page
There is no route for `about`. Production seems to mask the error by redirecting to the homepage.

Even if there were a route, we would still need to provide content for the page... Perhaps place this on pause until content is ready? Or perhaps remove the button until there is content for the page?

hectorcorrea:
Added route, controller, and view for About page. Text is Lorem Ipsum for now, we can adjust once we have the real text to use.

Closes #541

### DIFF
--- a/app/controllers/welcome_controller.rb
+++ b/app/controllers/welcome_controller.rb
@@ -3,4 +3,5 @@ class WelcomeController < ApplicationController
   skip_before_action :authenticate_user!
 
   def index; end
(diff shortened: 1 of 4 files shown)


mccalluc (approved):
I'm a little surprised that this wouldn't get its own controller, but I don't really care. 

Filed #739 to get the content.

Issue #717, opened 2022-12-12
Funder data should be multi-valued


bess:
- See also #729

We need to be able to describe funding information that looks like this (from https://dataspace.princeton.edu/handle/88435/dsp01rj4307478?mode=full): 

> The U.S. Department of Energy (DE-SC0018925,DE-AC02-76SF00515), the National Science Foundation (NSF) (EAR-1644614, EAR-1446969, EAR-1725349), and the German Research Foundation DFG (AP 262/2-1 and FOR2440).

Mock up: 

<img width="1020" alt="Screenshot 2022-12-12 at 6 11 11 PM" src="https://user-images.githubusercontent.com/65608/207181684-ec737434-5d2f-4513-9a7b-d58f4f3ba43f.png">

## Acceptance criteria
- [ ] Funding reference is multi-valued and a user can keep adding more funding references as needed
- [ ] All funding references appear in the DataCite record that is produced

kelynch:
Per conversation with @astrochun and @hhkitsune, this is not a priority for MVP.  ROR is not widely known or supported for PPPL constituents, and DOE does not currently support ROR for funding information reports. 

We are interested in supporting this feature in the future (alongside free text options), but this is not a high priority right now.

mccalluc:
(Narrowed the scope in the title and description and filed #729 for the follow-up RoR work.)

PR #738, opened 2022-12-14
Proposal: Pass data as JSON rather than hidden HTML

mccalluc:
Working on this issue:
- #717

I would like all the multi-valued fields to work the same way, so if there are things I'd like to tweak first, I'll file small PRs.

In this case, I think just passing the data as JSON is simpler than creating hidden html which we then need to parse with jQuery.


mccalluc:
Even simpler if we populate the form in the ERB: See #761

PR #744, opened 2022-12-14
Proposal: Pass creator as JSON rather than hidden HTML

mccalluc:


mccalluc:
Even simpler if we populate the form in the ERB: See #761

Issue #687, opened 2022-12-08
Configure and test email notification sending


kelynch:
Mailcatcher (SMTP server that runs locally) is installed (possibly not configured) on staging.

## Acceptance Criteria 
- [x] Ensure production server can send out email
- [x] Ensure the email setup is documented
- [x] Ensure [mailcatcher on staging setup](https://github.com/pulibrary/approvals#staging-mail-catcher) is documented so developer can use it for testing
- [ ] Emails should go out ONLY if user has email enabled in their account, in the following cases:
  - [x] In work state cases documented [here](https://github.com/pulibrary/pdc_describe/blob/main/docs/work_states.md) 
  - [ ] Comments in the system with "@" their username

carolyncole:
Once #853 is closed I believe this can be closed.  The issue with collections and mail for non curators is documented here: https://github.com/pulibrary/pdc_describe/issues/762

PR #747, opened 2022-12-14
Changing mailing to happen in one line so that it will send

carolyncole:
For some reason if you set the message to a local variable you get an error like `You've accessed the message before...` 
Changing it to a one liner allows the messages to send 

refs #687


carolyncole:
Never mind... This only happens in the console

PR #748, opened 2022-12-14
Create a partial that wraps data spans

mccalluc:
Before tackling the multi-funders, I would like to clean up the existing code. Copy-pasting the existing pattern would be digging myself into a hole. One pain point is where field names change between contexts: It will be less confusing to be consistent. Down the road I think it would be simpler just to serialize the data to JSON, and read that as the input, but I don't want to change too many things at once.

Spans for data are also used in `app/views/collections/edit.html.erb`, but with more special case logic for the values of some fields, so I don't want to touch those right now.

"draft" until I can get four errors fixed:
```
rspec ./spec/system/curator_controlled_metadata_spec.rb:62 # Curator Controlled metadata tab As a super admin allows editing of curator controlled fields
rspec ./spec/system/curator_controlled_metadata_spec.rb:44 # Curator Controlled metadata tab As a collection admin allows editing of curator controlled fields
rspec ./spec/system/curator_controlled_metadata_spec.rb:19 # Curator Controlled metadata tab As a princeton submitter does not allow editing of curator controlled fields
rspec ./spec/system/work_spec.rb:66 # Creating and updating works Handles Rights field
```

### DIFF
--- a/app/assets/javascripts/edit_work_utils.js
+++ b/app/assets/javascripts/edit_work_utils.js
@@ -215,7 +215,7 @@ $(() => {
       el = $(creatorSpans[i]);
       creator = {
         num: el.data('num'),
-        orcid: el.data('orcid'),
+        identifier: el.data('identifier'),
         givenName: el.data('given-name'),
         familyName: el.data('family-name'),
         sequence: el.data('sequence'),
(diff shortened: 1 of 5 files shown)


mccalluc:
Better approach will be to just populate the form in the ERB.
- See #761

PR #756, opened 2022-12-15
Adds search to the user's dashboard

### LINKED ISSUE #205: Curator dashboard - search feature
Curators can use the search bar to quickly find any records using DOI, Title, Depositor/submitter, ARK.  Allow partial terms in the title and the depositor.

(see https://github.com/pulibrary/pdc_describe/issues/176 for background and screenshot)

This is complete when:

- [ ]  Indexes in the database on our search terms
- [ ] There is a search bar in the UI that has a a text field to enter any of doi title depositor and ark
- [ ] A search will show the dashboard filtered to just the works that match the search
- [ ] There is a way to clear the search and go back to all results
- [ ] There are system tests that prove this works from the user's perspective

See https://www.scimedsolutions.com/blog/arel-part-i-case-insensitive-searches-with-partial-matching for example

hectorcorrea:
Adds ability to search for works from the user's dashboard.

search

Example searches in staging (results might vary depending on your access):
* Find in title: https://pdc-describe-staging.princeton.edu/describe/users/hc8719?q=automeris
* Find in DOI: https://pdc-describe-staging.princeton.edu/describe/users/hc8719?q=+10.80021%2F0hj6-9f47&commit=Search
* Find in contributors: https://pdc-describe-staging.princeton.edu/describe/users/hc8719?q=Sylvia&commit=Search
* Not found: https://pdc-describe-staging.princeton.edu/describe/users/hc8719?q=blahblahblah

Closes #205

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -127,7 +127,7 @@ input[readonly] {
 }
 
 .activity-history-item .comment-html {
-  // For all fields we just do a string diff; 
+  // For all fields we just do a string diff;
   // For arrays, we join the values with "\n",
   // but we need to make sure this is preserved when rendered.
   white-space: pre-wrap;
@@ -171,4 +171,13 @@ details {
 .copy-doi-label-copied {
   font-size: 8pt;
   color: green;
+}
+
+.search-bar {
+  display: flex;
+  justify-content: flex-end;
+}
+
+.search-form > form > input[type=text]{
+  width: 300px;
 }
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -178,6 +178,10 @@ details {
   justify-content: flex-end;
 }
 
+.search-form > form {
+  margin-bottom: -20px;
+}
+
 .search-form > form > input[type=text]{
   width: 300px;
 }
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -91,6 +91,12 @@
   margin-bottom: 25px;
 }
 
+.dataset-section-no-works {
+  background-color: rgb(247, 177, 177);
+  margin-top: 25px !important;
+  margin-bottom: 25px;
+}
+
 .dashboard-options {
   margin-top: -5px;
   margin-bottom: -25px;
@@ -180,8 +186,7 @@ details {
 
 .search-form > form {
   margin-bottom: -20px;
-}
-
-.search-form > form > input[type=text]{
-  width: 300px;
+  & > input[type=text] {
+    width: 300px;
+  }
 }
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -12,7 +12,7 @@ class UsersController < ApplicationController
 
   # GET /users/1
   def show
-    @search_terms = params["q"].present? ? params["q"] : nil
+    @search_terms = params["q"].presence
     @can_edit = can_edit?
     @my_dashboard = current_user.id == @user.id
     @unfinished_works = Work.unfinished_works(@user, @search_terms)
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -4,7 +4,7 @@ class UsersController < ApplicationController
   COLLECTION_MESSAGING_DISABLED = "0"
   COLLECTION_MESSAGING_ENABLED = "1"
 
-  before_action :set_user, only: %i[show edit update, dashboard_search]
+  before_action :set_user, only: %i[show edit update]
 
   def index
     @users = User.all
@@ -12,22 +12,12 @@ class UsersController < ApplicationController
 
   # GET /users/1
   def show
+    @search_terms = params["q"].present? ? params["q"] : nil
     @can_edit = can_edit?
     @my_dashboard = current_user.id == @user.id
-    @unfinished_works = Work.unfinished_works(@user)
-    @completed_works = Work.completed_works(@user)
-    @withdrawn_works = Work.withdrawn_works(@user)
-  end
-
-  def dashboard_search
-    byebug
-    search_terms = params["search_terms"].present? ? params["search_terms"] : nil
-    @can_edit = can_edit?
-    @my_dashboard = current_user.id == @user.id
-    @unfinished_works = Work.unfinished_works(@user, search_terms)
-    @completed_works = Work.completed_works(@user, search_terms)
-    @withdrawn_works = Work.withdrawn_works(@user, search_terms)
-    render "show"
+    @unfinished_works = Work.unfinished_works(@user, @search_terms)
+    @completed_works = Work.completed_works(@user, @search_terms)
+    @withdrawn_works = Work.withdrawn_works(@user, @search_terms)
   end
 
   # GET /users/1/edit
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -18,7 +18,7 @@ class UsersController < ApplicationController
     @unfinished_works = Work.unfinished_works(@user, @search_terms)
     @completed_works = Work.completed_works(@user, @search_terms)
     @withdrawn_works = Work.withdrawn_works(@user, @search_terms)
-    @works_found = @unfinished_works.count + @completed_works.count + @withdrawn_works.count
+    @works_found = @unfinished_works.length + @completed_works.length + @withdrawn_works.length
   end
 
   # GET /users/1/edit
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -4,7 +4,7 @@ class UsersController < ApplicationController
   COLLECTION_MESSAGING_DISABLED = "0"
   COLLECTION_MESSAGING_ENABLED = "1"
 
-  before_action :set_user, only: %i[show edit update]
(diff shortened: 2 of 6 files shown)


hectorcorrea:
I don't like that a search by DOI works (e.g. "10.80021/0hj6-9f47") but a search by the full URL does not (e.g. "https://doi.org/10.80021/0hj6-9f47"). 

This behavior it's because we don't store the full URL but I think it is confusing to users and we should make it transparent to them. But this is a first pass.

app/models/work.rb
```diff
@@ -85,16 +85,25 @@ def administered_by?(user)
...
-    def withdrawn_works(user)
-      works_by_user_state(user, "withdrawn")
+    def withdrawn_works(user, search_terms = nil)
+      works = works_by_user_state(user, "withdrawn")
+      filter_works(works, search_terms)
+    end
+
+    def filter_works(works, search_terms)
+      return works if search_terms.nil?
+      terms = search_terms.strip.downcase
```
carolyncole, line 105: This will become slow if the user's work list is large.  Likely that is why they are searching.

It would be better to send the query to the database.  I believe a like query can be done on a JSONB field.

carolyncole (changes requested):
We can do the query at the database level vs in the code:
` Work.where("CAST(metadata AS VARCHAR) like '%Carolyn%'")`

app/assets/stylesheets/application.scss
```diff
@@ -172,3 +172,16 @@ details {
...
+  justify-content: flex-end;
+}
+
+.search-form > form {
+  margin-bottom: -20px;
+}
+
+.search-form > form > input[type=text]{
+  width: 300px;
+}
```
mccalluc, line 187: Nest these? I think the syntax is:
```suggestion
.search-form > form {
  margin-bottom: -20px;
  & > input[type=text] {
    width: 300px;
  }
}
```

app/models/work.rb
```diff
@@ -85,16 +85,25 @@ def administered_by?(user)
...
 
-    def withdrawn_works(user)
-      works_by_user_state(user, "withdrawn")
+    def withdrawn_works(user, search_terms = nil)
+      works = works_by_user_state(user, "withdrawn")
+      filter_works(works, search_terms)
+    end
+
+    def filter_works(works, search_terms)
+      return works if search_terms.nil?
```
mccalluc, line 104:
```suggestion
      return works unless search_terms
```

app/models/work.rb
mccalluc: Agree that this should probably move to the database, eventually, but for a slightly different reason: When we have enough records that query performance becomes a problem, we will need to page the results, and we would need to filter before paging.

Unless paging has already been implemented, I don't know that this is essential, but it wouldn't be bad to tackle it now, either.

app/controllers/users_controller.rb
```diff
@@ -12,11 +12,13 @@ def index
...
+    @search_terms = params["q"].presence
     @can_edit = can_edit?
     @my_dashboard = current_user.id == @user.id
-    @unfinished_works = Work.unfinished_works(@user)
-    @completed_works = Work.completed_works(@user)
-    @withdrawn_works = Work.withdrawn_works(@user)
+    @unfinished_works = Work.unfinished_works(@user, @search_terms)
+    @completed_works = Work.completed_works(@user, @search_terms)
+    @withdrawn_works = Work.withdrawn_works(@user, @search_terms)
+    @works_found = @unfinished_works.count + @completed_works.count + @withdrawn_works.count
```
mccalluc, line 21: My understanding is that `count` always does a separate SQL `COUNT` query, while `size` is usually preferred because it will use `length` if the data is already in memory.

app/models/work.rb
```diff
@@ -487,6 +496,17 @@ def attach_s3_resources
...
 
+  # Returns true if the work matches the search terms. It searches inside
+  # the resource, the curator, and the work fields.
+  # Uses a brute-force approach to search for the value.
+  # A more fine tuned version could be developed to search within certain specific fields.
+  def match?(search_terms)
+    return true if resource && resource.to_json.downcase.include?(search_terms)
+    return true if curator && curator.to_json.downcase.include?(search_terms)
+    return true if to_json.downcase.include?(search_terms)
+    false
```
mccalluc, line 507: Slight preference to just make this one big boolean.

app/views/users/show.html.erb
```diff
@@ -2,6 +2,19 @@
...
+<div class="search-bar">
+  <div class="search-form">
+    <%= form_with(url: user_path(@user.uid), method: "get") do %>
+      <%= text_field_tag(:q, @search_terms, {placeholder: "Search by DOI, Title, Name, ARK"}) %>
+      <%= submit_tag("Search", class: "btn btn-primary") %>
+    <% end %>
+
+    <% if @works_found == 0 && @search_terms.present? %>
+      <p>No works found for <i><%= @search_terms %></i></p>
+    <% end %>
```
mccalluc, line 14: In UI terms, slight preference to put the "no results" message where the results would be, if there were any.

If this were an application-wide search, then up top makes sense, but otherwise better to place to the search box right above the results.

But just my opinion!

mccalluc (changes requested):
I concerned that if we went with this approach we'd get feedback back very fast that the search wasn't functioning as expected. This would be doing an exact phrase search, and on single words it doesn't do any stemming.

If we need to offer this, it's probably worth looking into [Postgres support for free text search](https://www.postgresql.org/docs/current/textsearch.html).

app/models/work.rb
carolyncole: @mccalluc I think better in positive terms and try to avoid unless.  Is there a solid reason for this change?  I would prefer @hectorcorrea's code.

app/controllers/users_controller.rb
carolyncole: Since this is an Array, I doubt it is sending a SQL query.    But prioritizing length over count seems like a reasonable thing.  
```
[4] pry(main)> unfinished.count
   (12.3ms)  SELECT COUNT(*) FROM "works"
=> 2
[5] pry(main)> unfinished.to_a.count
=> 2
[6] pry(main)> unfinished.length
=> 2
```

app/models/work.rb
hectorcorrea: Moved the search to the database.

app/assets/stylesheets/application.scss
hectorcorrea: Good catch. Done.

app/controllers/users_controller.rb
hectorcorrea: Changed to length.

app/views/users/show.html.erb
hectorcorrea: Good idea. Done.

app/models/work.rb
hectorcorrea: Method is now gone.

app/models/work.rb
hectorcorrea: Method is now gone.

app/models/work.rb
```diff
@@ -120,20 +120,30 @@ def publish_test_doi?
...
     private
 
-      def works_by_user_state(user, state)
-        # The user's own works by state (if any)
-        works = Work.where(created_by_user_id: user, state: state).to_a
+      def search_terms_where_clause(search_terms)
+        if search_terms.nil?
+          Work.where("true")
+        else
+          Work.where("CAST(metadata AS VARCHAR) ilike :search_terms", search_terms: "%" + search_terms.strip + "%")
```
hectorcorrea, line 127: @carolyncole This is the search via through Postgres

app/models/work.rb
```diff
@@ -120,20 +120,30 @@ def publish_test_doi?
...
+        end
+      end
+
+      def works_by_user_state(user, state, search_terms)
+        search_terms_filter = search_terms_where_clause(search_terms)
+
+        # The user's own works (if any) by state and search terms
+        works = Work.where(created_by_user_id: user, state: state).and(search_terms_filter).to_a
```
hectorcorrea, line 135: I had never chained WHERE clauses like this but it became handy for this case.

carolyncole: I have always chained the queries using a memo, but this is a nice way too: https://github.com/pulibrary/lockers_and_study_spaces/blob/17260cee97525c04e955d0a50f6e898aba30af42/app/models/locker_assignment.rb#L18

app/models/work.rb
```diff
@@ -120,20 +120,30 @@ def publish_test_doi?
 
     private
 
-      def works_by_user_state(user, state)
-        # The user's own works by state (if any)
-        works = Work.where(created_by_user_id: user, state: state).to_a
+      def search_terms_where_clause(search_terms)
+        if search_terms.nil?
+          Work.where("true")
```
carolyncole, line 125: @hectorcorrea I like .all better than .where("true"), but not going to hold up the PR for it...

carolyncole (approved):

app/models/work.rb
mccalluc, line 127:
```suggestion
          Work.where("CAST(metadata AS VARCHAR) ILIKE :search_terms", search_terms: "%" + search_terms.strip + "%")
```

Capitalize keywords.

mccalluc (approved):
One capitalization tweak, and I'll merge.

mccalluc (approved):

PR #757, opened 2022-12-15
Use a multi-line JSON string in spec for readability

mccalluc:
Needed to touch this in the process of making funders multi-valued, and thought it could be more readable. This PR only changes the formatting. Surprised that there were no rubocop warnings locally.

### DIFF
--- a/spec/models/work_spec.rb
+++ b/spec/models/work_spec.rb
@@ -894,21 +894,53 @@ RSpec.describe Work, type: :model do
 
   describe "resource=" do
     let(:resource_json) do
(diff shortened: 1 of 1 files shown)


carolyncole (changes requested):
Would rather the json be put into a file and then read back vs changing the indentation like this...

mccalluc:
@carolyncole , does indenting the JSON suffice? Since this is just a clean up, I really wanted to keep the changes small, and if the JSON is in-line, having it formatted seems more readable than it was.

(Looking at this code because the work on funders will prompt a change here, and I'd like that change to be readable.)

carolyncole (approved):

PR #759, opened 2022-12-15
Support multiple funders on work

### LINKED ISSUE #717: Funder data should be multi-valued

mccalluc:
- Fix #717

This is simpler than the approach taken with the other multi-valued fields: Using Rails param conventions for repeating fields means we don't need JS to copy data into the table or to keep track of the row count, and the ruby form processing code is also simpler. The JS for adding a row is generic and could be used to add rows to any table. If this much is acceptable, the same patterns could be applied to other fields:
- #761 
- #760
- #765 

Questions / TODO:
- Could this much be merged, or do we need the items below?
- Besides adding rows, do we want to be able to reorder and delete rows?
- Should we use a table for display of funders even when not editing? Or something else.
- More testing.
- Make sure funder changes show up in prov log.

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -99,15 +99,19 @@ class Work < ApplicationRecord
     end
 
     def find_by_doi(doi)
-      models = all.select { |work| !work.doi.nil? && work.doi.include?(doi) }
-      raise ActiveRecord::RecordNotFound if models.empty?
-      models.first
+      # This does a string compare to allow for partial matches
+      # I'm not entirely sure why we are allowing partial matches in a find by
+      Work.find_by!("metadata->>'doi' like ? ", "%#{doi}")
+      # I think we should be doing the following instead, which matches exactly
+      # Work.find_by!('metadata @> ?', "{\"doi\":\"#{doi}\"}")
     end
 
     def find_by_ark(ark)
-      models = all.select { |work| !work.ark.nil? && work.ark.include?(ark) }
-      raise ActiveRecord::RecordNotFound if models.empty?
-      models.first
+      # This does a string compare to allow for partial matches
+      # I'm not entirely sure why we are allowing partial matches in a find by
+      Work.find_by!("metadata->>'ark' like ? ", "%#{ark}")
+      # I think we should be doing the following instead, which matches exactly
+      # Work.find_by!('metadata @> ?', "{\"ark\":\"#{ark}\"}")
     end
 
     delegate :resource_type_general_options, to: PDCMetadata::Resource


app/models/pdc_serialization/datacite.rb
```diff
@@ -215,11 +215,11 @@ def rights_from_work_resource(resource)
...
 
-        def funding_reference_from_work_resource(resource)
-          return nil unless resource.award_number.present? && resource.funder_name.present?
-          award = ::Datacite::Mapping::AwardNumber.new(uri: resource.award_uri, value: resource.award_number)
-          funding_reference = ::Datacite::Mapping::FundingReference.new(name: resource.funder_name, award_number: award)
-          [funding_reference]
+        def funding_references_from_work_resource(resource)
+          resource.funders.map do |funder|
+            award = ::Datacite::Mapping::AwardNumber.new(uri: funder.award_uri, value: funder.award_number)
+            ::Datacite::Mapping::FundingReference.new(name: resource.funder_name, award_number: award)
```
carolyncole, line 221: Shouldn't this be funder.funder name here?  I feel like this code must be untested.

app/services/form_to_resource_service.rb
```diff
@@ -28,6 +25,10 @@ def convert(params, work)
       # Process contributors
       resource = process_contributors(params, resource)
 
+      # Process funders
+      # (New pattern: method modifies resource in place.)
+      process_funders(params, resource)
```
carolyncole, line 30: I would prefer if the method returns the updated resource,  Just to make the code clear that the resource is being changed by the method.  I think that is the pattern process_contributors takes.  That said we should make it all constant.

app/services/form_to_resource_service.rb
```diff
@@ -113,5 +114,20 @@ def new_contributor(given_name, family_name, orcid, type, sequence)
         return if family_name.blank? && given_name.blank? && orcid.blank?
         PDCMetadata::Creator.new_contributor(given_name, family_name, orcid, type, sequence)
       end
+
+      # Funders:
+
+      def process_funders(params, resource)
+        # (New pattern: Use rails param name conventions rather than numbering fields.)
```
carolyncole, line 121: I'm unsure what the params are going to look like here.  That does not indicate a code change as long as it is show in the tests.

app/services/form_to_resource_service.rb
```diff
@@ -113,5 +114,20 @@ def new_contributor(given_name, family_name, orcid, type, sequence)
...
+        resource.funders = (params[:funders] || []).filter_map do |funder|
+          new_funder(funder[:funder_name], funder[:award_number], funder[:award_uri])
+        end
+        # (New pattern: no return -- method modifies resource in place.)
+      end
+
+      def new_funder(funder_name, award_number, award_uri)
+        return if funder_name.blank? && award_number.blank? && award_uri.blank?
```
carolyncole, line 129: This explains why the need for filter map...

app/views/works/_form_hidden_fields.html.erb
```diff
@@ -8,8 +8,6 @@
   <input type="text" id="new_title_count" name="new_title_count" value="0" class="hidden" />
   <input type="text" id="creator_count" name="creator_count" value="<%= @work&.resource&.creators&.count || 0%>" class="hidden" />
   <input type="text" id="contributor_count" name="contributor_count" value="<%= @work&.resource&.contributors&.count || 0%>" class="hidden" />
-  
-  <!-- The count tells the controller how many related objects to expect -->
```
carolyncole, line 12: Why remove this comment.  It still seems relevant

spec/system/work_edit_spec.rb
```diff
@@ -132,13 +132,13 @@
       sign_in user
       visit edit_work_path(work)
       click_on "Additional Metadata"
-      fill_in "funder_name", with: "National Science Foundation"
-      fill_in "award_number", with: "nsf-123"
-      fill_in "award_uri", with: "http://nsg.gov/award/123"
+      fill_in "funders[][funder_name]", with: "National Science Foundation"
```
carolyncole, line 135: Need to test multiple

carolyncole (changes requested):
There seems to be some untested code.  Also it would be good to test multiple items to show your new params pattern.  Additionally the javascript is entirely untested.  A system test could help that.

app/models/pdc_serialization/datacite.rb
mccalluc: thanks! will do.

app/services/form_to_resource_service.rb
mccalluc: My preference would be that it either modify in place, or assign to a new instance. Right now it's doing both, and I think that could be confusing. `process` is pretty vague, so I'll rename the methods for clarity.

app/services/form_to_resource_service.rb
mccalluc: Right: This is pattern that is already used for creator and contributor, but with an extra `if` at the end. If there's an empty row in the table, we don't create a record for it.

app/views/works/_form_hidden_fields.html.erb
mccalluc: The same hold for all the fields above: `new_title_count` tells the controller how many new titles to expect, `creator_count` tells the controller how many creators to expect, etc. 

In the slightly longer run, we won't need these fields as we move towards using the `[]` rails convention for repeated field names.

spec/system/work_edit_spec.rb
mccalluc: Thanks!

mccalluc:
Back to draft. Need to add
- [ ] Test of serialization
- [ ] Test of data entry for multiple
- [ ] Rename `process_*` fields
- [ ] Test of prov logging

Carolyn didn't comment on these, so I think they can wait for a future PR?
- reorder and delete
- better display when not editting; table?

mccalluc:
Probably hold off on reviewing this until #771 is reviewed and merged: It handles part of the testing, and I've merged that branch into this.

Looking at your comments, I was unsure about this:
> That said we should make it all constant.

Do you mean that immutable objects are good? Generally, I'd agree, but when it's just breaking the constructor down into smaller pieces, using private methods, I'm ok with it.

The more important thing to me is that a set of methods are consistent about whether they modify the object, or return a new object, but these methods were doing both.

app/views/works/show.html.erb
```diff
@@ -174,17 +174,17 @@
...
-  <% if @work.resource.award_number.present? %>
-    <p>Award Number: <%= @work.resource.award_number %></p>
-  <% end %>
-
-  <% if @work.resource.award_uri.present? %>
-    <p>Award URI: <%= @work.resource.award_uri %></p>
-  <% end %>
+  <% if @work.resource.funders.count > 0 %>
+    <p>Funders:
+      <% @work.resource.funders.each do |funder| %>
```
carolyncole, line 179: Are we loosing the Headings here? `Funder Name` `Award Number` `Award URI:`

carolyncole (approved):
Just one question on the show page.  That said I do not think that should block this PR, but we should know what the fields represent.  Maybe ticket deciding how the Funders should be shown?

app/views/works/show.html.erb
mccalluc: I had thought it might be clear enough from context: Each subfield has a distinct format. Since this will repeat, I'm not sure if this would be better presented as a table. 
- Filed #785 for follow-up.

PR #763, opened 2022-12-16
Adding production smtp configuration

carolyncole:
Email successfully sent from production:
Screen Shot 2022-12-19 at 8 40 56 AM


mccalluc (approved):
Is there a separate issue for the server name in the message, or is that a config issue that's outside this codebase?

PR #765, opened 2022-12-16
Cleanup form_to_resource_service

mccalluc:
- At the top level, in `convert`, group all the simple field setters, and group all the calls to `process` methods, and reorder the calls and definitions to be consistent.
- If `resource` is modified in place, we don't need to return and reassign it: Make the methods simple and consistent.
- Create consistent paired `process_fields`/`new_field` methods.
- Use `filter_map` to drop the empty rows.

(I'll be adding to this file for #759, and I wanted to establish a pattern I can follow.)

### DIFF
--- a/app/services/form_to_resource_service.rb
+++ b/app/services/form_to_resource_service.rb
@@ -20,8 +20,8 @@ class FormToResourceService
       resource.award_uri = params["award_uri"]
 
       process_curator_controlled(params, resource)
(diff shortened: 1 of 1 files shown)


carolyncole (approved):
Not sure I am in love with filter_map, but it is not awful.  I find when reading code some of the more magical methods in Rails tend to make the code harder to read in the long run.

PR #766, opened 2022-12-16
Send Notifications to the User and Curators on State changes

carolyncole:
refs #687


carolyncole:
Added this test to #836

PR #770, opened 2022-12-19
Moving curator to an ActiveRecord relation to make it more consistent with other work relationships

carolyncole:
This puts the relationship up top and makes it more consistent with the other relationships the work has such as Activities and Files.

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -12,6 +12,7 @@ class Work < ApplicationRecord
   has_many_attached :pre_curation_uploads, service: :amazon_pre_curation
 
   belongs_to :collection
+  belongs_to :curator, class_name: "User", foreign_key: "curator_user_id", optional: true
 
   attribute :work_type, :string, default: "DATASET"
   attribute :profile, :string, default: "DATACITE"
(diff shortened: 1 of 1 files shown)


app/models/work.rb
```diff
@@ -12,6 +12,7 @@ class InvalidCollectionError < ::ArgumentError; end
   has_many_attached :pre_curation_uploads, service: :amazon_pre_curation
 
   belongs_to :collection
+  belongs_to :curator, class_name: "User", foreign_key: "curator_user_id", optional: true
```
mccalluc, line 15: Is there any argument for accepting the default foreign key? We've just consistently said `<role>_user_id` instead of `<role>_id`, so we should stick with it?

mccalluc (approved):
One small question, but if this seems like the right course, feel free to merge.

app/models/work.rb
carolyncole: This was keeping the code the same and just changing the mechanism.  We may want to ticket changing the name if we think it is desired.  That will impact many more places.

PR #771, opened 2022-12-19
Test Resouce JSON to Datacite XML conversion

mccalluc:
- Review of #759 pointed out gap in test coverage.
- Read resource JSON as YAML so comments can be added, if needed.
- I think that many of the existing tests could be pruned back: Instead of making assertions one-by-one for individual elements, we will have better coverage checking the entire document.
- Add gem for robust xml comparison.
- More fixtures could be added to the directory, if needed.


app/models/work_activity.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require_relative "../lib/diff"
+require_relative "../lib/simple_diff"
```
carolyncole, line 3: Didn't this change already get merged?

spec/models/pdc_serialization/datacite_spec.rb
```diff
@@ -247,6 +248,27 @@
...
+
+        datacite_path = resource_path.gsub(".resource.yaml", ".datacite.xml")
+        datacite_xml_expected = File.read(datacite_path)
+        expect(datacite_xml).to be_equivalent_to(datacite_xml_expected)
+      end
+    end
+
+    it "does not contain stray files" do
+      Dir.glob("#{fixtures_dir}/*").each do |path|
+        expect(path).to match(/\.resource\.yaml$|\.datacite\.xml$/)
```
carolyncole, line 267: Where would these stray files come from??

carolyncole (changes requested):
Just a couple updates.  It looks like it might need to be rebased.  Also can you add a comment on why stray files would show up...

spec/models/pdc_serialization/datacite_spec.rb
mccalluc: From me, perhaps: In the past, I have found that having directories of fixtures that are turned into tests can be useful. But at least once I've tripped up because I added a fixture that didn't match the expected glob pattern, and so it wasn't testing against an example I had provided.

This prevents that.

carolyncole: Could you add a comment in the test to say this is why... Maybe something like If this test fails you have an example file that is not being tested...

spec/models/pdc_serialization/datacite_spec.rb
```diff
@@ -247,6 +248,27 @@
...
+        resource = PDCMetadata::Resource.new_from_json(resource_json)
+        datacite_xml = resource.to_xml
+
+        datacite_path = resource_path.gsub(".resource.yaml", ".datacite.xml")
+        datacite_xml_expected = File.read(datacite_path)
+        expect(datacite_xml).to be_equivalent_to(datacite_xml_expected)
+      end
+    end
+
+    it "does not contain stray files" do
```
carolyncole, line 265:
```suggestion
    # We want to be certain that the files in the fixtures directory are all getting utilized.
    # If this test fails a file has been added to the directory, bit this test has not been updated to include it.
    it "does not contain stray files that are not explicitly being tested above" do
```

carolyncole (dismissed):

carolyncole (approved):

Issue #761, opened 2022-12-16
Simplify form population by creating inputs in ERB


mccalluc:
Right now, data from the backend is put into hidden elements, which are then scanned when the page loads and inputs are created and filled by JS. The same `add*Html` routines are also used to add new blank rows to the tables.

This could be simplified by:
- Creating the html inputs with plain ERB.
- To add a row, take the previous row as a template, but just clear the input values.

It is fine for the input names to repeat across rows: We don't need counters.
- See #760.

PR #773, opened 2022-12-20
Move add-row buttons into tfoot

mccalluc:
It will simplify the JS if we have more generic event handlers: Buttons can add rows to the tables where they are located, instead of needing one handler per button.

This PR moves the buttons, and consolidates the JS for row-reordering.

- Pattern demonstrated in #759
- Towards #761


app/views/works/_related_objects_table.html.erb
```diff
@@ -4,14 +4,22 @@
...
         <th>Related Identifier Type</th>
         <th>Relation Type</th>
       </tr>
-    </tbody>
+    </thead>
+    <tbody id="related-objects-table"></tbody>
+    <tfoot>
+      <tr>
+        <td colspan="3">
+          <%= link_to "Add Another Related Object", "#", id: "btn-add-related-object", class: "btn btn-secondary" %>
```
carolyncole, line 19: Is there code that should be deleted from Related?  Maybe this needs a rebase??

mccalluc: I'm not sure that I understand: The button that's added just above was moved/deleted from farther down in the file.

carolyncole: The PR seems to indicate that this is a refactor.  This seems to add new functionality.

mccalluc: I don't understand your comment: If we're talking about this line, the `Add Another Related Object` is being moved from a div underneath the table to a tfoot inside the table (which corresponds to the title on the PR).

Let me know what I'm missing / what needs to change?

carolyncole (approved):

PR #781, opened 2022-12-21
change findy_by_doi and find_by_ark to utilize a json query in the database

carolyncole:
This will be entirely more efficient when the database gets to be larger than a few records.

I have made the code match the functionality that exists in the current tests, but I am unsure why we are doing a partial match in a find_by, so I put the code that would work for an exact match also commented out.


mccalluc (commented):
I agree that this is a better approach, and the exact match is even better, but the behavior when there are no matches is changing -- It used to raise an exception, but now it returns nil.

Since this didn't cause a test failure, maybe there's a gap in the tests?

carolyncole:
> I agree that this is a better approach, and the exact match is even better, but the behavior when there are no matches is changing -- It used to raise an exception, but now it returns nil.
> 
> Since this didn't cause a test failure, maybe there's a gap in the tests?

Actually the `find_by!` does in fact raise an error when there is no match: https://apidock.com/rails/v4.0.2/ActiveRecord/FinderMethods/find_by%21

mccalluc:
Oh, sorry, that's what the `!` does. Thanks!

mccalluc (approved):

PR #798, opened 2022-12-22
Fix bug with validation of multi-word relation types, and consistently use the "value" rather than the "key" of Datacite enumerations

### LINKED ISSUE #784: Trying to save a Related Object does not work
It causes a validation error even though a type is chosen
Screen Shot 2022-12-21 at 9 53 41 AM

mccalluc:
- Fix #784: Multi-word relation types now validate.
- Use `Datacite::Mapping` values throughout, rather than mixing keys and values. If we always use value, much less likely to hit bugs like this in the future.
- Remove `upcase`: If we're being consistent, it's not necessary; If we're not consistent, it can mask errors.
- Also, noticed that an explicit `to_a` is not necessary before `map`, so cleaned it up on all Mappings.
- Remove a couple unused methods and tests of those methods 

This does change the values that are allowed in the resource json, and I'd guess we'd want a migration rake task? But before investing any more time on this wanted to check in and see if this much is headed in the right direction.


spec/models/pdc_metadata/related_object_spec.rb
```diff
@@ -11,24 +11,6 @@
     expect(new_related_object.value).to eq related_identifier
   end
 
-  ##
```
jrgriffiniii, line 14: 👍 this did not reduce the test coverage: https://coveralls.io/builds/55432856

jrgriffiniii (dismissed):

mccalluc:
Thanks! I'm going to move this back to draft -- On the agenda in January I've put an item to ask about migration. More work may or may not be required.

mccalluc:
I think this PR represents the end goal we'd like to reach, but adding the backward compatibility code and the rake task will make it even more complicated. Instead, I'm going to make smaller PRs that only migrate single fields. After the migrations, either this PR can be revived, or the same work can be finished in a new, simpler PR that just removes the backward compatibility and the migration tasks.

mccalluc:
I think this is ready to go... but it is a lot of changes. In the middle of the effort, it seemed like it would be simpler to try to handle all the cases the same way, but in the end it involved more tweaks than I had anticipated. This does not include a migration: Instead, it just makes sure we can read both new (datacite value) and old (datacite key/upcased value) JSON. Writing a migration task could be a follow-up, unless reviewers think it should also be added to this PR.

bess:
Tested this on staging and looks like the bug is fixed! Thank you! ⭐

bess (approved):

PR #801, opened 2022-12-28
Ensuring that Work#approve logs a message indicating that the Work was successfully published by the curator user

### LINKED ISSUE #361: Notify the Submitter when a DataSet is approved
Utilize the notification system created by #360 to notify the Submitter when a dataset is approved.
This can be accomplished by adding to AASM approved state after action [publish] (https://github.com/pulibrary/pdc_describe/blob/main/app/models/work.rb#L35) in [this routine.](https://github.com/pulibrary/pdc_describe/blob/main/app/models/work.rb#L351-L354
)

This is complete when:
- [ ] The submitter of a dataset receives an notification when a dataset is marked as approved
- [ ] There is an automated test that checks that these notifications get sent

jrgriffiniii:
Resolves #361

### DIFF
--- a/spec/models/work_spec.rb
+++ b/spec/models/work_spec.rb
@@ -577,6 +577,20 @@ RSpec.describe Work, type: :model do
 
         expect(draft_work.curator).to eq(curator_user)
       end
+
+      it "sends a notification for the approval of the work" do
+        allow(WorkActivity).to receive(:add_system_activity)
+        draft_work = FactoryBot.create(:draft_work)
+        file_name = uploaded_file.original_filename
+        stub_work_s3_requests(work: draft_work, file_name: file_name)
+        draft_work.pre_curation_uploads.attach(uploaded_file)
+
+        draft_work.complete_submission!(user)
+        draft_work.update_curator(curator_user.id, user)
+        draft_work.approve!(curator_user)
+
+        expect(WorkActivity).to have_received(:add_system_activity).with(draft_work.id, "successfully published", curator_user.id)
+      end
     end
 
     it "is approved" do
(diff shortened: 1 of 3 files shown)


spec/models/work_spec.rb
```diff
@@ -604,6 +604,20 @@
...
+        draft_work = FactoryBot.create(:draft_work)
+        file_name = uploaded_file.original_filename
+        stub_work_s3_requests(work: draft_work, file_name: file_name)
+        draft_work.pre_curation_uploads.attach(uploaded_file)
+
+        draft_work.complete_submission!(user)
+        draft_work.update_curator(curator_user.id, user)
+        draft_work.approve!(curator_user)
+
+        expect(WorkActivity).to have_received(:add_system_activity).with(draft_work.id, "successfully published", curator_user.id)
```
mccalluc, line 619: I think the original issue sets a higher bar for testing:
> There is an automated test that checks that these notifications get sent

I think this is checking that the right method is getting called, but I don't think it's checking that the email actually gets sent.

Maybe there are other tests that already cover that? I wouldn't block if this seems sufficient to other folks.

mccalluc (commented):
May need more testing? Or check with others to confirm that this is enough.

spec/models/work_spec.rb
jrgriffiniii: I am very sorry for this, I will extend the test suite in order to ensure that the e-mail messages are sent.

mccalluc: My memory of the conversation is that Bess agreed that this is sufficient, and Carolyne's work will handle the end-to-end process of sending a message, but whatever you think best.

jrgriffiniii: I apologize, due to being unwell, and wishing to advance this, I will prepare this for review and merge please.

jrgriffiniii:
This was deprecated by https://github.com/pulibrary/pdc_describe/pull/836

PR #812, opened 2023-01-04
Factored out the list of works, which makes the work model smaller.

carolyncole:
In an effort to not continually add items to the work class this factors out the lists of works into thier on class. 
Also completed test coverage for the different methods


bess (approved):
Great refactor. Sandi Metz would be proud. ⭐ 😁

PR #813, opened 2023-01-04
Moving S3 blob generation into the S3 File to make the work model smaller

carolyncole:
This just moves the code out of the work into the S3 file, which is where I believe it belongs.

### DIFF
--- a/spec/models/s3_file_spec.rb
+++ b/spec/models/s3_file_spec.rb
@@ -39,4 +39,37 @@ RSpec.describe S3File, type: :model do
       expect(CGI.unescape(url_file)).to eq(filename.split("/").last)
     end
   end
+
+  describe "#to_blob" do
+    let(:file) do
+      S3File.new(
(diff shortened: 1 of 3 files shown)


spec/models/s3_file_spec.rb
```diff
@@ -39,4 +39,37 @@
...
+      before do
+        persisted = ActiveStorage::Blob.create_before_direct_upload!(
+          filename: file.filename, content_type: "", byte_size: file.size, checksum: ""
+        )
+        persisted.key = file.filename
+        persisted.save
+      end
+
+      it "finds the blob" do
+        expect { file.to_blob }.to change { ActiveStorage::Blob.count }.by(0)
```
mccalluc, line 71:
```suggestion
        expect { file.to_blob }.not_to change { ActiveStorage::Blob.count }
```
Slightly more readable.

mccalluc (approved):
One possible change to test syntax, but fine to merge as-is, too.

mccalluc (approved):

PR #817, opened 2023-01-04
In the select options, only show contributor roles that are apply to individual contributors

### LINKED ISSUE #815: Limit contributor roles pulldown to personal roles
- Follow-up from #811
- See also #816

For each contributor, instead of listing all possible roles, the existing pulldown should be limited to roles that individuals can fill.

mccalluc:
- Fix #815


bess (dismissed):

mccalluc:
(Not sure why the main merge caused problems, but looking into it...)

bess (approved):

PR #818, opened 2023-01-05
Add license text to the last page of the wizard workflow

### LINKED ISSUE #764: Users must accept our license when submitting
Per a requirement from OGC, users legally must accept a license (what we should regard as "Terms and Conditions") when submitting an item to PDC Describe.  This should largely follow the workflow for license agreement that is currently in DataSpace.

Example in DataSpace:

As the last step of the submission process, the user is presented with the license agreement text:
Screen Shot 2022-12-16 at 11 53 05 AM
Screen Shot 2022-12-16 at 11 53 10 AM

Once they click on the "I Grant the License" button, their submission is complete.

## Acceptance criteria:
- [ ] The [license text](https://docs.google.com/document/d/1aEvbsMFJ8ZD1TEBieX6tcYO8WB4TdAtXs70r2EGrui4/edit) is reviewed and finalized
- [ ] The license is presented to the user as the final stage of submission, with a button acknowledging/granting the license
- [ ] The text of the license is also available to the public

mccalluc:
- Fix #764

Screen Shot 2023-01-05 at 12 22 50 PM

I believe this is a simple way to do what is needed, though it might not be exactly what was envisioned. Some points to consider:
- It does not add a separate page to the wizard: The provided screenshot has "License" as a step preceding "Complete".
- It does not use a fixed-width font for the license-text, nor does it wrap the license text in a scrollable div, both of which are common in license UIs.
- It does not add license-approval as an extra bit of state in the model: Instead I've just added "Grant License" to the button.
- It also does not give an option to proceed without granting the license, which is an option in the DataSpace screenshot.
- ... so there is no need for this text "Not granting the license will not delete your submission..."
- It only has this sentence once, instead of repeating it in the text of the license: "In order for Princeton University's Princeton Data Commons to reproduce, translate, and distribute..."

Finally, it makes one change to the existing text on the page:
```diff
- Data curators are carefully reviewing the submission ...
+ Data curators will review the submission ...
```
This has seemed a little out of place, even with the current workflow, but with the addition of the license approval, it feels less confusing to put the review in the future.

Some of these would be trivial to change, but making license approval an explicit part of work state would be a bit of work.

### DIFF
--- a/app/views/works/review.html.erb
+++ b/app/views/works/review.html.erb
@@ -1,9 +1,58 @@
 <div class="wizard-area">
   <h1>New Submission (<%= @work.title %>)</h1>
   <%= render "wizard_progress", wizard_step: 3 %>
-  <p>Thanks so much!</p>
+  <p>
+    In order for Princeton University's Princeton Data Commons to reproduce, translate, and distribute
+    your submission for worldwide public access your agreement to the following terms is necessary.
+    Please take a moment to read the terms of this license.
+  </p>
 
-  <p>Data curators are carefully reviewing the submission for accuracy and completeness to make
+  <blockquote style="border: 1px solid; padding: 1em;">
+    <p>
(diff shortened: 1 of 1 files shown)


bess (approved):
Thoughtfully done. Thank you!

Issue #782, opened 2022-12-21
Put Notes in Messages


hhkitsune:
## Description

The notes from the submission are currently being rendered as part of the record metadata. However, the notes people may leave are messages and should appear with the messaging feature. The note should be placed with other messages and the curator will see a notification if a note was included with the submission.

## Image shows a note (see next to arrow) in the record metadata area. This is a message for the curators.

![Screen Shot 2022-12-21 at 9 28 35 AM](https://user-images.githubusercontent.com/50670293/208929756-488d821d-d863-4989-aa58-5392243bde67.jpeg)

## Acceptance Criteria

- [ ] Move note to messages
- [ ] Curator receives a numbered notification if a note was included.
- [ ] The numbered notification resolves when a curator reads the note

PR #819, opened 2023-01-05
Move submission notes out of metadata and into the message list

mccalluc:
- #814 should be reviewed and merged before reviewing this PR.
- This moves towards #782 but doesn't fully address it.

Screen Shot 2023-01-05 at 1 36 58 PM

This PR just moves the HTML around, but looking at the original issue:

> - Curator receives a numbered notification if a note was included.
> - The numbered notification resolves when a curator reads the note

To do all of that, we'll need to change the model code: should be possible, but it is more complicated.

If this seems valuable as a partial solution, it could be approved... but if it seems better to implement the full solution, then this could be closed.

### DIFF
--- a/app/models/s3_file.rb
+++ b/app/models/s3_file.rb
@@ -15,4 +15,18 @@ class S3File
     encoded_filename = filename.split("/").map { |name| CGI.escape(name) }.join("/")
     File.join(Rails.configuration.globus["post_curation_base_url"], encoded_filename)
   end
+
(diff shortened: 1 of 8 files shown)


bess:
⭐ Looks great! 

I made a separate ticket to refactor the code to use consistent language for message instead of comment: https://github.com/pulibrary/pdc_describe/issues/825

bess (approved):

PR #822, opened 2023-01-05
Exporting data migration records

hectorcorrea:
Sort of automates the process to export and import the test data from one instance to another. This is heavily based on the steps indicated https://github.com/pulibrary/pdc_describe/blob/main/docs/test_data_in_production.md

The general workflow is:

```
# Export the data to JSON files 
DATA_MIGRATION=true bundle exec rspec ./spec/system/data_migration

# Delete existing data (optional)
bundle exec rake works:delete_user_works[your-net-id]

# Import the data from the JSON files
bundle exec rake works:import_works[./tmp/data_migration]
```

Work towards https://github.com/pulibrary/pdc_discovery/issues/340

### DIFF
--- a/app/assets/javascripts/edit_work_utils.js
+++ b/app/assets/javascripts/edit_work_utils.js
@@ -86,7 +86,7 @@ $(() => {
 
     const rowHtml = `<tr id="${rowId}" class="related-objects-table-row">
       <td>
-        <input type="text" id="${relatedIdentifierId}" name="${relatedIdentifierId}" value="${related_identifier}" data-num="${num}" placeholder="The URL web address for a related publication or other resource" />
+        <input type="text" id="${relatedIdentifierId}" name="${relatedIdentifierId}" value="${related_identifier}" data-num="${num}"/>
       </td>
       <td>
         ${relatedIdentifierTypeHtml}
--- a/app/views/works/_related_objects_table.html.erb
+++ b/app/views/works/_related_objects_table.html.erb
@@ -7,7 +7,7 @@
   <table>
     <thead>
       <tr class="header-row">
-        <th>Related Identifier (Example: URI for a related publication or resource)</th>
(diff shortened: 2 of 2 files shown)


lib/tasks/works.rake
```diff
@@ -58,4 +58,27 @@ namespace :works do
...
+    path = File.join(args[:path], "*.json")
+    approver = User.first
+    puts "Importing files from: #{path}"
+    Dir.glob(path).each do |file_name|
+      hash = JSON.parse(File.read(file_name))
+      resource = PDCMetadata::Resource.new_from_json(hash)
+      work = Work.new(resource: resource)
+      work.collection = Collection.research_data
+      work.created_by_user_id = approver.id
+      work.state = "approved"
```
hectorcorrea, line 79: Set to approved for now because that is what I need to get the records to show in the RSS feed but we could tweak this.

hectorcorrea:
@bess I sort of automated the process to export/import the data from the tests based on your instructions in the docs folder. This is something that I am using as I work on https://github.com/pulibrary/pdc_discovery/issues/340 and I thought could be useful in general.

spec/system/data_migration/baldwin_form_submission_spec.rb
```diff
@@ -64,6 +64,8 @@
       # This will allow us to evolve our local datacite standards and test our records against them.
       datacite = PDCSerialization::Datacite.new_from_work(baldwin_work)
       expect(datacite.valid?).to eq true
+
+      export_spec_data("baldwin.json", baldwin_work.to_json)
```
hectorcorrea, line 68: Notice that by default the call to `export_spec_data` does nothing, it only exports the data if `ENV["DATA_MIGRATION"] == "true"`

bess (approved):
🤩

Issue #823, opened 2023-01-06
Messages: button and "right-side" messages


hhkitsune:
## Description

We want the messages feature to be immediately findable for users. It is located on the record pages in Describe, which also contains the object's record and provenance information. Moving the messages to another page may be too complicated at this time, but having it with the other information can be confusing with everything else on the page. It was suggested we move the messages to the right-side of the record and have a button that would allow messages to be made visible or moved from view. 

## Value added

The suggestion to have the messages appear at the right-side of the record would be useful for users to be able to compare information in the messages with the record itself. For example, some messages from a curator may suggest changes to a specific field in the record, such as the "Description" field. The submitter may see what they have written on the left and quickly comment back to the curator to receive feedback for a change they might make.

Having the messages appear or disappear is useful for the different purposes the record is being used. For example, messages are needed for communication during the curation process, but they may not be needed if a moderator is just checking a record for remediation or analytic purposes. 

## Image shows the addition of a blue "messages" button
![Screen Shot 2023-01-06 at 10 24 14 AM](https://user-images.githubusercontent.com/50670293/211046766-ff384117-6df2-4c96-ab36-35f764fa9740.jpeg)

## Image shows the messages appear after the blue "messages" button is selected and would disappear is selected again
![Screen Shot 2023-01-06 at 10 29 14 AM](https://user-images.githubusercontent.com/50670293/211046858-bd6d8eba-4d7c-4f25-ac59-d64f207a87f3.jpeg)

## Acceptance Criteria
- [ ] A bright blue button is added for the messages 
- [ ] Messages will appear (at right) when button is selected and disappear when it is selected again
- [ ] The message input box is still below the messages (include a "submit" button)
- [ ] The record information fields may need to move from the right side of the screen for this feature to work properly

PR #824, opened 2023-01-06
Minimalist UI for prov and messages

mccalluc:
@hhkitsune should review this and see if the simplified UI is closer to what she wants, and if it is better, what additional tweaks are needed, if any.

Not that there are other PRs in the wings which will also make improvements here:
- #814
- #821 
- #819

before and after:
Screen Shot 2023-01-06 at 1 32 05 PM

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -163,6 +163,7 @@ details {
   }
 
   summary.show-changes {
+    font-weight: normal;
     &::after {
       content: 'Show Changes';
     }
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -207,3 +207,17 @@ dl {
     grid-column-start: 2;
   }
 }
+
+ul.beads {
+  padding-left: 1em;
+  margin-left: 1em;
+  border-left: 2px solid;
+  li {
+    list-style: none;
+  }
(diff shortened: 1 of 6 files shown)


hhkitsune:
@mccalluc I love the appearance of the provenance! It is perfect. Thank you so much. : )

I'm not sure why the messages (currently called "comments") look this way though. It should only be the provenance that has this appearance and message feature should be more consistent with how messages typically look. Regarding messages, I just wrote a ticket for changes to the messaging feature that was discussed in our last feature refinement meeting. Bess just took a look at it and said the ticket was actionable. You might also want to have a look: #823. Please feel welcome to comment or ask questions.

mccalluc:
> I'm not sure why the messages (currently called "comments") look this way though.

Both sections were using the same style before, so the simplest change was to make a single change, rather than adding special cases. But that is doable. I'll update this PR.

hhkitsune:
@mccalluc Thanks, much appreciated!

mccalluc:
Updated screen shot, with a distinct style for messages, and other recent features merged in. Tests are looking for particular elements, but the structure of the page has changed. Will fix.

#823 will be handled separately: We want to keep this to a small set of changes.

Screen Shot 2023-01-06 at 4 36 30 PM

bess (approved):

Issue #760, opened 2022-12-16
Simplify form handling with Rails parameter naming conventions


mccalluc:
We have a lot of code, on both the JS and Ruby sides, to handle the submission of multi-valued fields. We can simplify this by using [Rails parameter naming conventions](https://guides.rubyonrails.org/form_helpers.html#combining-them).

In JS:
```diff
-  function addCreatorHtml(num, givenName, ...) {
-    ...
-    const givenNameId = `given_name_${num}`;
-    ...
-    const rowHtml = `...
-      <td>
-        <input type="text" id="${givenNameId}" name="${givenNameId}" value="${givenName}" />
-      </td>
+  function addCreatorHtml(givenName, ...) {
+    const rowHtml = `...
+      <td>
+        <input type="text" name="" value="creators[][given_name]" />
+      </td>
```
- (Or we might be able to drop this JS entirely: See #761)

In Ruby:
```diff
      def process_creators(params, resource)
-       (1..params["creator_count"].to_i).each do |i|
-         creator = new_creator(params["given_name_#{i}"], params["family_name_#{i}"], params["orcid_#{i}"], params["sequence_#{i}"])
+       params[:creators].each do |params_creator|
+         creator = new_creator(params_creator[:given_name], params[:family_name], params[:orcid], params[:sequence])
          resource.creators << creator unless creator.nil?
        end
        resource
      end
```

PR #826, opened 2023-01-06
Simplify handling of related objects

mccalluc:
Towards
- #760
- #761

This code uses the Datacite enum values, rather than the keys. This PR should be merged first: 
- #798

Tests will break because of inconsistency in the values. Filing as draft until that is resolved.


mccalluc:
Out of date; There's not much here, so best just to close to clean up the PR list.

PR #830, opened 2023-01-09
Change the state to be titleized when shown to the user

carolyncole:
This stop awaiting_approval to be displayed.

Also utilized constants in WorkActivity for the types.


spec/factories/work_activity.rb
```diff
@@ -5,6 +5,6 @@
     message { "test work activity message" }
     activity_type { "SYSTEM" }
     created_by_user_id { FactoryBot.create(:user).id }
-    work_id { FactoryBot.create(:work).id }
+    work_id { FactoryBot.create(:draft_work).id }
```
mccalluc, line 8: Not a problem, but also not sure what motivates this particular change...

mccalluc (approved):
Approved, but it looks like Bess may already be working on getting this up to date and merging.

spec/factories/work_activity.rb
carolyncole: You are right this is an unrelated change.  This factory was not utilized an errors when you do. I'll update the PR comment to highlight that this is also changed.

PR #831, opened 2023-01-09
Adding a notification WorkActivity type 

carolyncole:
fixes #687

Issue #675, opened 2022-12-07
Separate each field as a separate event in prov log


mccalluc:
Hannah's goal:
> The entries are standardized and look clean. I think part of the difference is each change is its own entry, rather than lumping them together

Currently, changes to title and author are displayed like this:
<img width="310" alt="Screen Shot 2022-12-07 at 12 08 42 PM" src="https://user-images.githubusercontent.com/730388/206244837-3870f5ad-4a03-4662-b8a9-10f841a41d8d.png">

Instead the changes for each field should be separate, with the timestamp and user restated for each, like this:
<img width="326" alt="Screen Shot 2022-12-07 at 12 16 22 PM" src="https://user-images.githubusercontent.com/730388/206246415-7ad202f5-af9b-458a-8402-e26ed79ed497.png">

Prefer to implement this as only a change to the display code, but if there's a problem with that, consider tweaking the model.

(Separately, we also want to improve the display of changes, instead of showing add + remove events. See #645.)

mccalluc:
There are higher level goals with the display of provenance, but the particular issue here is just splitting the events, and that was fixed in 
- #695

PR #836, opened 2023-01-09
Sending a notification to the Depositor and the Collection Administrators when state changes occur

carolyncole:
Moves the code from the Work Model to it's own class WorkStateTransitionNotification

refs #675 

This is related to #831, but it leaves the notification as SYSTEM messages, which means they will show up in the comments/messages section of the work.

This is the smallest change to advance #675 without implementing a Notification screen we do not know if the curators will want.

The only unfortunate part of this change is that the Depositor and the Curators receive the same note. If we sent two different ones then the comments/message section would grow even more.  This may be good enough.  I could see that it would be nice to tell the depositor that this is just a receipt and that no action is required on their part for any of the notifications.

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -32,7 +32,7 @@ class Work < ApplicationRecord
     end
 
     event :complete_submission do
-      transitions from: :draft, to: :awaiting_approval, guard: :valid_to_submit, after: :notify_collection_curators
+      transitions from: :draft, to: :awaiting_approval, guard: :valid_to_submit
     end
 
     event :request_changes do
@@ -505,6 +505,7 @@ class Work < ApplicationRecord
       uw = UserWork.new(user_id: user.id, work_id: id, state: state)
       uw.save!
       WorkActivity.add_system_activity(id, "marked as #{state.to_s.titleize}", user.id)
+      WorkStateTransitionNotification.new(self, user.id).send
     end
 
     def data_cite_connection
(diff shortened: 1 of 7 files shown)


carolyncole:
@jrgriffiniii I think you and I may have come at this same functionality from two angles.  Your PR #801 and this PR both send a notification when the state has changed to Approved.  This one also adds one for Draft and Awaiting Approval.

How do you feel about this PR?  I'd be happy to discuss how to combine them, or merge yours and then adapt mine.  I do not have the Successfully Published method, so maybe both PRs are needed.

jrgriffiniii:
Thank you very much for this @carolyncole! My preference would be to please merge your PR, close my own PR, and then offer a second PR with a proposal for any changes which I felt my first PR still addressed (if there are any - this is really excellent and may well indeed cover the cases which mine sought to cover).

app/models/work.rb
```diff
@@ -32,7 +32,7 @@ class InvalidCollectionError < ::ArgumentError; end
     end
 
     event :complete_submission do
-      transitions from: :draft, to: :awaiting_approval, guard: :valid_to_submit, after: :notify_collection_curators
+      transitions from: :draft, to: :awaiting_approval, guard: :valid_to_submit
```
jrgriffiniii, line 35: 👍

app/models/work_state_transition_notification.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def send
+    return if notification.nil?
+
+    WorkActivity.add_system_activity(id, notification, current_user_id, activity_type: WorkActivity::SYSTEM)
+  end
+
+    private
+
+      def notification_for_transition
```
jrgriffiniii, line 29: 👍

jrgriffiniii (approved):

PR #839, opened 2023-01-09
Allow free text notes in provenance for a limited set of users

### LINKED ISSUE #720: Mechanism for free-text notes in provenance log
- Related issue: #718

We've just separated the display of comments from the provenance log, but it seems like there may be cases where we need to describe an event with free-text. Before implementation:
- [x] confirm with users that there is a use case for this feature.
- [ ] confirm that there isn't code that already does this.
- [ ] get buy-in on: input UI / internal representation / output UI

## Acceptance Criteria
- [ ] Only developers + Hannah (defined by a list of netids?) have access to this feature. It will initially be used only for migration purposes.
- [ ] People who are migrating records (initially, Tyler, Bess, and Kevin from CaMS) can enter short comments (e.g., "Initially created in DataSpace") 
- [ ] Each comment has a comment body, an editable date, and a user (Clarification: Yes, we want to be able to back-date these, but the user is just the logged-in user.)
- [ ] The comment goes into the provenance log

mccalluc:
- Eventually, fix #720 

Tests failing right now. I'd like to get feedback about this approach before digging deeper.

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -188,6 +188,9 @@ class WorksController < ApplicationController
     @uploads = @work.uploads
     @wizard_mode = true
     @work.complete_submission!(current_user)
+
+    # Do not add provenance notes until validated:
+    @work.add_provenance_notes(params["provenance_notes"], current_user.id)
     redirect_to user_url(current_user)
   end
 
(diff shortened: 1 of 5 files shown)


mccalluc:
- Close in favor of #848

PR #843, opened 2023-01-10
Switched to contributors a few authors in the bitKlavier spec

hectorcorrea:
Some of the people for this data set are contributors rather than authors (see original DSpace record https://dataspace.princeton.edu/handle/88435/dsp01nv9356017) and the new spec mimics this.

Part of https://github.com/pulibrary/pdc_discovery/issues/340


carolyncole (approved):
Good Catch!

PR #844, opened 2023-01-10
Change find to select to allow the test to wait

carolyncole:
Some tests were continuously failing on my machine.  It may be because my machine is a bit faster. 
In general it is better to utilize the rspec methods over find as they will delay until the page fully loads
I also find it easier to understand which option is being chosen


hectorcorrea:
Nice, I also like the `select` syntax better than what were doing with `find` and xpath.

hectorcorrea (approved):

PR #848, opened 2023-01-11
Let admins add free-text, back-dated messages to change-history

### LINKED ISSUE #720: Mechanism for free-text notes in provenance log

mccalluc:
- Replace #839
- Fix #720

Notes: 
- User list is hard-coded, as indicated in the issue: I'm the only one on it for now: please add netIDs as needed.
- On the screen we're calling this the "Change History", but I also hear "Provenance" -- It looks like the name "provenance" hasn't really be used in the code before, so I would changes this to "change-history"... unless we should be moving existing references to "provenance"?
- There's no validation of the date string; there could be. The Rails date parsing has some strange behavior:

```
January 12, 2023 00:00 by Chuck
Date was "12"

January 11, 2023 16:35 by Chuck
Date was "1234"

January 11, 2023 00:00 by Chuck
Date was "123"

January 01, 0012 00:00 by Chuck
Date was "12345"
```

Screen Shot 2023-01-11 at 4 27 52 PM

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -237,7 +237,7 @@ class WorksController < ApplicationController
     work = Work.find(params[:id])
     if params["new-provenance-note"].present?
       new_date = params["new-provenance-date"]
-      new_note = params["new-provenance-note"]
+      new_note = html_escape(params["new-provenance-note"])
 
       work.add_provenance_note(new_date, new_note, current_user.id)
     end
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -233,6 +233,17 @@ class WorksController < ApplicationController
     redirect_to work_path(id: params[:id])
   end
 
+  def add_provenance_note
+    work = Work.find(params[:id])
+    if params["new-provenance-note"].present?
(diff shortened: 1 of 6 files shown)


bess (approved):
This looks great! 
I made a follow up ticket to move the configuration to a config file, but I'm merging this as-is because it unblocks us for data migration.

Follow up ticket: https://github.com/pulibrary/pdc_describe/issues/849

PR #851, opened 2023-01-12
Include files in Work JSON

### LINKED ISSUE #841: Include Collection in Work JSON
The data that we expose for a work in JSON (e.g. http://localhost:3000/works/43.json) does not include the Collection and we need that value in PDC Discovery (see also https://github.com/pulibrary/pdc_discovery/issues/371). 

I believe the culprit is that for the JSON format we export the Resource (which is part of the Work) but the Collection information (e.g. "Research Data") is part of the Work itself and not included in the Resource.

Part of https://github.com/pulibrary/pdc_discovery/issues/340

### LINKED ISSUE #842: Include files in Work JSON
The data that we expose for a work in JSON (e.g. http://localhost:3000/works/43.json) does not include the list of files in the Work and we need that value in PDC Discovery. 

Part of https://github.com/pulibrary/pdc_discovery/issues/340
See also https://github.com/pulibrary/pdc_describe/issues/841

mccalluc:
- fix #842
- fix #841 
- Going forward, work JSON != resource JSON
- Should we be able to get more information on the files?
- I'm not sure if pre- and post-curation files should be distinguished.
- Not seeing an obvious way to get the unsanitized filename from [ActiveStorage](https://api.rubyonrails.org/classes/ActiveStorage/Filename.html). 

<details>
<summary>Questions Hector replies to below</summary>

Draft. Open questions:
- Do you need anything besides filename downstream?
- Should post curation uploads be treated just the same?
- There doesn't seem to already be a test where works with attached files are serialized to json. Thoughts about the best place to add this?
- Should we prohibit direct use of `resource.to_json`?

In the same vein, it worries me that we'll now have these two forms of json that are almost, but not quite the same. A different approach would be more compositional. Instead of what this PR does:
```
{
  ... existing fields ...
  "files": ["file1.txt"]
}
```
... we could move everything down a level, to keep the data separate:
```
{
  "resource": { ... existing fields ... },
  "work": { "files": ["file1.txt"] }
}
```
... but that would involve a lot more work.

Thinking about it, I think there are two separate usecases for the JSON, and maybe we should try to keep them separate:
- A serialization structure for persistence in the DB.
- An export format for discovery.

</details>

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -448,10 +448,14 @@ class Work < ApplicationRecord
     log_file_changes(changes, nil)
   end
 
-  def as_json
+  def as_json(options = nil)
+    if options&.present?
+      raise(StandardError, "Received options #{options}, but not supported")
+      # Included in signature for compatibility with Rails.
+    end
     # to_json returns a string of serialized JSON.
     # as_json returns the corresponding hash.
-    files = pre_curation_uploads.map(&:as_json)
+    files = pre_curation_uploads.map { |upload| { "filename": upload.filename.to_s, "created_at": upload.created_at } }
     {
       "resource" => resource.as_json,
       "files" => files
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -116,7 +116,7 @@ class Work < ApplicationRecord
 
   before_save do |work|
     # Ensure that the metadata JSONB postgres field is persisted properly
-    work.metadata = JSON.parse(work.resource.to_json)
+    work.metadata = JSON.parse(work.to_json)
     work.save_pre_curation_uploads
   end
 
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -455,7 +455,7 @@ class Work < ApplicationRecord
     end
     # to_json returns a string of serialized JSON.
     # as_json returns the corresponding hash.
-    files = pre_curation_uploads.map { |upload| { "filename": upload.filename.to_s, "created_at": upload.created_at } }
+    files = (pre_curation_uploads + post_curation_uploads).map { |upload| { "filename": upload.filename.to_s, "created_at": upload.created_at } }
     {
       "resource" => resource.as_json,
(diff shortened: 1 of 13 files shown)


mccalluc:
Test passes locally. Investigating...

mccalluc:
Oh: Not an ordering problem: The files names are consistently different. Looking into it...

mccalluc:
There's a side effect in the tests that I don't understand:
```
  1) Work approves works and records the change history
     Failure/Error: expect(work.state_history.first.state).to eq "approved"

       expected: "approved"
            got: "awaiting_approval"

       (compared using ==)
     # ./spec/models/work_spec.rb:175:in `block (2 levels) in <top (required)>'
     # ./spec/rails_helper.rb:107:in `block (3 levels) in <top (required)>'
     # ./spec/rails_helper.rb:106:in `block (2 levels) in <top (required)>'

  2) Work resubmit works and records the change history
     Failure/Error: expect(work.state_history.first.state).to eq "draft"

       expected: "draft"
            got: "withdrawn"

       (compared using ==)
     # ./spec/models/work_spec.rb:275:in `block (2 levels) in <top (required)>'
     # ./spec/rails_helper.rb:107:in `block (3 levels) in <top (required)>'
     # ./spec/rails_helper.rb:106:in `block (2 levels) in <top (required)>'
```
Follow-up tomorrow.

app/models/work.rb
```diff
@@ -448,8 +448,31 @@ def attach_s3_resources
...
+        "created_at": upload.created_at
+      }
+    end
+
+    # to_json returns a string of serialized JSON.
+    # as_json returns the corresponding hash.
+    {
+      "resource" => resource.as_json,
+      "files" => files,
+      "collection" => collection.as_json.except("id")
```
hectorcorrea, line 470: cool, I didn't know you could omit fields with `except`

hectorcorrea (approved):

PR #845, opened 2023-01-11
Rename json to jsonb to more accurately represent what the method is doing

carolyncole:
Also rename hash to jsonb_hash to more accurately represent what is being passed as a parameter

fixes #802

PR #852, opened 2023-01-12
Moving Notification  on State Change out of the System ActivityType

### LINKED ISSUE #847: Move User Notifications to the Messaging/comments section
User Notifications on State Changes should not show up in the Change log.

Move them to the Comments section for the moment.  They may move later, so let's add an additional type so that we can classify them.  Change Log should look like the following:

Screen Shot 2023-01-11 at 8 00 51 AM

Change Log Currently includes:

Screen Shot 2023-01-11 at 7 58 22 AM

carolyncole:
Additionally moving logic for finding the different types of messages into WorkActivity in an effort to make the Work Model smaller.
Also included the title of the work in notifications as I though that would make the emails nicer.

fixes #847

### DIFF
--- a/app/models/work_activity.rb
+++ b/app/models/work_activity.rb
@@ -4,11 +4,16 @@ require_relative "../lib/diff_tools"
 
 # rubocop:disable Metrics/ClassLength
 class WorkActivity < ApplicationRecord
-  CHANGES = "CHANGES"
   MESSAGE = "COMMENT" # TODO: Migrate existing records to "MESSAGE"; then close #825.
+  NOTIFICATION = "NOTIFICATION"
+  MESSAGE_ACTIVITY_TYPES = [MESSAGE, NOTIFICATION].freeze
+
+  CHANGES = "CHANGES"
   FILE_CHANGES = "FILE-CHANGES"
-  SYSTEM = "SYSTEM"
   PROVENANCE_NOTES = "PROVENANCE-NOTES"
+  SYSTEM = "SYSTEM"
+  CHANGE_LOG_ACTIVITY_TYPES = [CHANGES, FILE_CHANGES, PROVENANCE_NOTES, SYSTEM].freeze
+
   USER_REFERENCE = /@[\w]*/.freeze # e.g. @xy123
 
   include Rails.application.routes.url_helpers
@@ -16,7 +21,7 @@ class WorkActivity < ApplicationRecord
   belongs_to :work
   has_many :work_activity_notifications, dependent: :destroy
 
-  def self.add_system_activity(work_id, message, user_id, activity_type: "SYSTEM", date: nil)
+  def self.add_system_activity(work_id, message, user_id, activity_type: SYSTEM, date: nil)
     activity = WorkActivity.new(
       work_id: work_id,
       activity_type: activity_type,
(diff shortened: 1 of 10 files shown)


spec/requests/works_spec.rb
```diff
@@ -41,8 +41,6 @@
 
         before do
           allow(work).to receive(:id).and_return("test-id")
-          allow(work).to receive(:changes).and_return([])
```
carolyncole, line 44: These methods no longer are defined on a Work, so they can not be mocked.

app/models/work_activity.rb
```diff
@@ -29,6 +34,22 @@ def self.add_system_activity(work_id, message, user_id, activity_type: "SYSTEM",
     activity
   end
 
+  def self.activities_for_work(work_id, types = [])
+    context = where(work_id: work_id).order(updated_at: :desc)
+    if types.count > 0
+      context = context.where(activity_type: types)
```
carolyncole, line 40: Instead of doing a select after the fact as the old code did add it to the where clause to let the database do what it does best.

mccalluc:
> fixes #845

Is that a typo? That's already closed, and this doesn't seem to relate to renaming.

app/models/work_activity.rb
```diff
@@ -77,7 +98,7 @@ def provenance_notes_type?
   end
 
   def log_event_type?
-    system_event_type? || file_changes_event_type? || changes_event_type? || provenance_notes_type?
+    CHANGE_LOG_ACTIVITY_TYPES.include? activity_type
```
mccalluc, line 101: Are the methods above still used? Maybe `def changes_event_type?` and the rest can be removed.

app/models/work_state_transition_notification.rb
```diff
@@ -21,19 +22,19 @@ def initialize(work, current_user_id)
...
+    WorkActivity.add_system_activity(id, notification, current_user_id, activity_type: WorkActivity::NOTIFICATION)
   end
 
     private
 
       def notification_for_transition
         case to_state
         when :awaiting_approval
-          "#{user_tags} The [work](#{work_url}) is ready for review."
+          "#{user_tags} The [#{work_title}](#{work_url}) is ready for review."
```
mccalluc, line 33: This might read awkwardly:
> The [The Cat in the Hat]() is ready for review.

Maybe drop the "The" outside the link?

spec/models/work_activity_spec.rb
```diff
@@ -45,4 +45,40 @@
...
+        expect(described_class.activities_for_work(work)).to eq([notification, work_activity])
+      end
+
+      it "finds all the activities for the work and type" do
+        expect(described_class.activities_for_work(work, [WorkActivity::SYSTEM])).to eq([work_activity])
+        expect(described_class.activities_for_work(work, [WorkActivity::NOTIFICATION])).to eq([notification])
+        expect(described_class.activities_for_work(work, [WorkActivity::SYSTEM, WorkActivity::NOTIFICATION])).to eq([notification, work_activity])
+      end
+    end
+    describe "#messages_for_work" do
```
mccalluc, line 69:
```suggestion

    describe "#messages_for_work" do
```

carolyncole:
Yes it should be #847

mccalluc (approved):
One suggested change to message format, but looks good.

Check if this is marked as fixing the correct issue number: Maybe a typo?

app/models/work_activity.rb
carolyncole: I believe they are utilized for the HTML formatting, but I did not check them all.

carolyncole:
@mccalluc I fixed up your suggestions.  Thanks for the review!

mccalluc (approved):

Issue #762, opened 2022-12-16
Any user should be able to configure email for collections


carolyncole:
Either I should be able to configure email notifications here:
https://github.com/pulibrary/pdc_describe/blob/main/app/views/users/_form.html.erb#L12-L25

Or I should always be sent email here:
https://github.com/pulibrary/pdc_describe/blob/main/app/models/work_activity_notification.rb#L10

Instead I get neither option:
![Screen Shot 2022-12-16 at 8 09 27 AM](https://user-images.githubusercontent.com/1599081/208106070-3e75a63e-f468-4c40-bea9-40946d12b98b.png)

PR #854, opened 2023-01-13
Allow collection depositors to receive emails

carolyncole:
Allow Users to recieve emails by default for collection they have deposit rights to Allow users to configure thier emails even when they are not collection administrators Fix bug so that you can uncheck email for specific collections by utilizing the form.check_box instead of the check_box_tag

fixs #762

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -39,6 +39,7 @@ RSpec.describe User, type: :model do
       user = described_class.from_cas(access_token)
       expect(user).to be_a described_class
       expect(user.default_collection.id).to eq Collection.default.id
+      expect(user.messages_enabled_from?(collection: user.default_collection)).to be_truthy
     end
 
     it "sets the proper default collection for a PPPL user" do
@@ -46,6 +47,7 @@ RSpec.describe User, type: :model do
       pppl_user = described_class.from_cas(access_token_pppl)
       expect(pppl_user).to be_a described_class
       expect(pppl_user.default_collection.id).to eq pppl_collection.id
+      expect(pppl_user.messages_enabled_from?(collection: pppl_user.default_collection)).to be_truthy
     end
 
     it "sets the CAS info on new" do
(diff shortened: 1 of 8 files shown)


spec/models/user_spec.rb
```diff
@@ -272,31 +274,79 @@
...
         expect(enabled_state).to be true
+      end
+    end
+
+    context "The user can deposit into the collection" do
+      before do
+        user.add_role :submitter, collection
+      end
+
+      it "disables email messages for notifications from a Collection" do
```
mccalluc, line 302:
```suggestion
      it "disables and enables email messages for notifications from a Collection" do
```

spec/models/user_spec.rb
```diff
@@ -272,31 +274,79 @@
...
+        user.disable_messages_from(collection: pppl_collection)
+      end .to raise_error(ArgumentError, "User #{user.uid} is not an administrator or depositor for the collection #{pppl_collection.title}")
+      expect do
+        user.enable_messages_from(collection: pppl_collection)
+      end .to raise_error(ArgumentError, "User #{user.uid} is not an administrator or depositor for the collection #{pppl_collection.title}")
+    end
+  end
+
+  describe "#submitter_or_admin_collections" do
+    it "returns both admin and submit collections" do
```
mccalluc, line 339:
```suggestion
    it "returns both admin and submitter collections" do
```

spec/models/work_activity_notification_spec.rb
```diff
@@ -17,18 +17,13 @@
     end
 
     it "does not enqueue an e-mail message to be delivered for the notification" do
```
mccalluc, line 19:
```suggestion
    it "does not enqueue an e-mail message to be delivered for the notification if disabled" do
```

spec/models/work_activity_notification_spec.rb
```diff
@@ -17,18 +17,13 @@
...
+      user.disable_messages_from(collection: collection)
       described_class.create(user: user, work_activity: work_activity)
       expect(message_delivery).not_to have_received(:deliver_later)
     end
 
     context "when e-mail notifications are enabled for the Collection" do
```
mccalluc, line 25: Connected to the comment just above: This context used to be the special case, but now it's the default case: Maybe move it out, and wrap the test just above with a context that disables messages?

mccalluc (approved):
Looks good. A couple comments on test names and organization. I didn't know about the default "Research Data" collection, so it was helpful for me to read this!

carolyncole:
@bess I rebased this with main and it is green.

bess (approved):

Issue #676, opened 2022-12-07
In prov log, use a beaded line in UI


mccalluc:
Example:
﻿
<img width="307" alt="Screen Shot 2022-12-07 at 12 22 55 PM" src="https://user-images.githubusercontent.com/730388/206247879-e287d0e4-01e9-4276-a064-5211a24d8134.png">

- Title beside the bead is the action
- User and date are below
- Beaded line is to the left

Should our prov log also be in chronological order? Currently we put the more recent items at the top. Get direction from Hannah, and update this issue.

mccalluc:
Beaded UI implemented by
- #824

... but seeing that there's also a change to the order of the fields. Keeping this open.

PR #857, opened 2023-01-13
Provenance and Messages ordered correctly; Explicit notice of backdated records; Show author UID in messages; Improve testing

mccalluc:
- Towards #676

@hhkitsune had requested:

> For the provenance, chronological is preferred.


mccalluc:
Sorry, I should have quoted @hhkitsune more fully:
> For the provenance, chronological is preferred. For the (to be separate) message feature (e.g., the one where curators are communicating with the depositor), most recent at the top is preferred.

I'm reading this as requesting difference in behavior between the two. Messages are most recent at the top; Provenance / changelog is oldest at the top.

bess:
We believe the request is to change the order so that messages are sorted newest first, and provenance is sorted by date of occurrence.

Order needs to be moved out of here: 
https://github.com/pulibrary/pdc_describe/blob/main/app/models/work_activity.rb#L39

mccalluc:
- Provenance and Messages ordered as requested: newest message at the top; oldest provenance at the top.
- Sorting is now done in the ERB, rather than the model, since the two rendering cases have different behavior.
- Moving the sort out of the model required tweaks to a handful of other tests: These really only care about the set membership, not the order, so these changes are not substantive.
- Explicit notice of backdated records, with `updated_at` used to record the actual date, and `created_at` used for the back-dated date. Happy to swap if these are not the desired semantics
- Show author UID in messages. Partially addresses #863.
- Two view tests, corresponding to the two partials, that cover the features listed above.

Screen Shot 2023-01-19 at 10 41 46 PM

Screen Shot 2023-01-19 at 10 41 58 PM

carolyncole (approved):
I would much prefer logic to be in the models, not the views.  That said it is well tested, so we should see when an issue occurs.

PR #858, opened 2023-01-13
Export the work JSON instead of the resource JSON in /works/:id.json

hectorcorrea:
Part of https://github.com/pulibrary/pdc_discovery/issues/373

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -78,7 +78,7 @@ class WorksController < ApplicationController
         @can_curate = current_user.can_admin?(@collection)
         @work.mark_new_notifications_as_read(current_user.id)
       end
-      format.json { render json: @work.resource }
+      format.json { render json: @work.to_json }
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/works_controller.rb
```diff
@@ -78,7 +78,7 @@ def show
         @can_curate = current_user.can_admin?(@collection)
         @work.mark_new_notifications_as_read(current_user.id)
       end
-      format.json { render json: @work.resource }
+      format.json { render json: @work.to_json }
```
mccalluc, line 81: Would it work without the explicit `to_json`? It seems like it used to.

mccalluc: I'll try it locally...

mccalluc (commented):
Curious if the `to_json` is required...

mccalluc (approved):
Looks like the explicit `to_json` is best: Without that, it tries to call `as_json` with options:
```
{:prefixes=>["works", "application"], :template=>"show", :layout=>#<Proc:0x00007fbd23a11610 /Users/chuck/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/actionview-6.1.6.1/lib/action_view/layouts.rb:390>}
```

approved.

PR #865, opened 2023-01-17
Allow Funder ROR to be supplied

### LINKED ISSUE #729: Funder data should use RoR
- See also #717

We need to be able to describe funding information that looks like this (from https://dataspace.princeton.edu/handle/88435/dsp01rj4307478?mode=full): 

> The U.S. Department of Energy (DE-SC0018925,DE-AC02-76SF00515), the National Science Foundation (NSF) (EAR-1644614, EAR-1446969, EAR-1725349), and the German Research Foundation DFG (AP 262/2-1 and FOR2440).

Mock up: 

Screenshot 2022-12-12 at 6 11 11 PM

## Acceptance criteria
- [ ] If an ROR is entered in the ROR field, the funding organization name is populated from the ror.org API
- [ ] ROR is optional -- a user can just enter free text into the funding organization field if they prefer
- [ ] There is a help button next to ROR that explains what an ROR is and links to https://ror.org/

mccalluc:
- Fix #729

Notes:
- This does not add any system tests for the API call. It could.
- This does not add any server-side validation: It could... Would we we want to just check the regex? or hit the API and make sure the ID resolves? or Make sure the ID is in sync with the name?
- I made a change to make the changelog for this field more readable... but it doesn't seem to show up at all. Filed  #867.

### DIFF
--- a/app/assets/javascripts/edit_work_utils.js
+++ b/app/assets/javascripts/edit_work_utils.js
@@ -501,6 +501,13 @@ $(() => {
     }
   });
 
+  // Fetch information for funder via the ROR API
+  $(document).on('change', '.ror-input', (el) => {
+    const $target = $(el.target);
+    const ror = $target.val().trim();
+    console.log(`TODO: Hit ROR API for ${ror}`);
+  });
+
   // Allows the creators and contributors to be reordered via drag and drop.
   // The `cancel` property "prevents sorting if you start on elements matching the selector"
   // https://api.jqueryui.com/sortable/#method-cancel
--- a/app/assets/javascripts/edit_work_utils.js
+++ b/app/assets/javascripts/edit_work_utils.js
@@ -505,7 +505,12 @@ $(() => {
   $(document).on('change', '.ror-input', (el) => {
     const $target = $(el.target);
     const ror = $target.val().trim();
-    console.log(`TODO: Hit ROR API for ${ror}`);
+    fetch(`https://api.ror.org/organizations/${ror}`).then()
+      .then((response) => response.json())
+      .then((responseJson) => {
+        const { name } = responseJson;
+        $target.closest('tr').find('.ror-output').val(name);
+      });
   });
 
   // Allows the creators and contributors to be reordered via drag and drop.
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/edit_work_utils.js
```diff
@@ -501,6 +501,18 @@ $(() => {
     }
   });
 
+  // Fetch information for funder via the ROR API
+  $(document).on('change', '.ror-input', (el) => {
+    const $target = $(el.target);
+    const ror = $target.val().trim();
+    fetch(`https://api.ror.org/organizations/${ror}`).then()
```
jrgriffiniii, line 508: Might it be possible to please request error handling? I can create another issue if you wish to have this merged first (I have no preference at all, and do not mean this to sound critical).

```javascript
fetch(`https://api.ror.org/organizations/${ror}`).then()
      .then((response) => response.json())
      .then((responseJson) => {
        const { name } = responseJson;
        $target.closest('tr').find('.ror-output').val(name);
      })
      .catch((error) => {
        console.error('Error querying the Research Organization Registry (ROR) API:', error);
      });
});
```

app/views/works/_funder_table.html.erb
```diff
@@ -21,7 +30,10 @@
     <% (@work.resource.funders + [nil]).each do |funder| %>
       <tr>
         <td>
-          <input name="funders[][funder_name]" value="<%= funder&.funder_name %>" />
+          <input name="funders[][ror]" value="<%= funder&.ror %>" placeholder="https://ror.org/........" pattern="https://ror.org/0[a-z|0-9]{6}[0-9]{2}" class="ror-input" />
```
jrgriffiniii, line 33: 👍

jrgriffiniii (approved):

app/assets/javascripts/edit_work_utils.js
mccalluc: will do!

mccalluc: ... actually, this doesn't do what you want. Exceptions won't fall through to the end of the chain, so this would only handle an error in the jquery, not in the http, or in parsing the JSON. I'll file an issue for follow-up, and reference this.

PR #866, opened 2023-01-17
Ensuring that multiple titles can be updated for an approved Work

### LINKED ISSUE #833: Only curators should be able to modify approved items
Curators should be able to modify objects after they have been approved.

Test to ensure that curators can modify objects that they themselves submitted after they have been approved.

Additional titles for a Work are not being saved *when the work has already been approved*. Although I can edit the work and add a new title via the `+` button, the title is **not** saved on the work.
 
add_title_edit

No error is reported to the user during Save, but the system goes back to the home page rather than displaying the Show page for the work after save.

The system works as expected when the title is added to a work that has not yet been approved.

Related to https://github.com/pulibrary/pdc_discovery/issues/340

jrgriffiniii:
Resolves #833

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -114,8 +114,6 @@ class WorksController < ApplicationController
     if current_user.blank? || !@work.editable_by?(current_user)
       Rails.logger.warn("Unauthorized attempt to update work #{@work.id} by user #{current_user.uid}")
       redirect_to root_path
-    elsif @work.approved? && @work.submitted_by?(current_user)
-      redirect_to root_path, notice: I18n.t("works.approved.uneditable")
     else
       update_work
     end
--- a/spec/requests/works_spec.rb
+++ b/spec/requests/works_spec.rb
@@ -51,6 +51,99 @@ RSpec.describe "/works", type: :request do
           expect { get work_url(work) }.to raise_error(Work::InvalidCollectionError, "The Work test-id does not belong to any Collection")
         end
       end
+
+      context "when the work has been approved" do
+        let(:work) { FactoryBot.create(:approved_work) }
+        let(:collection) { work.collection }
+
+        context "when appending a second title" do
+          let(:params) do
+            {
+              id: work.id,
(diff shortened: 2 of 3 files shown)


spec/requests/works_spec.rb
```diff
@@ -53,6 +53,99 @@
...
+              expect(titles.length).to eq(2)
+
+              subtitle = titles.last
+              expect(subtitle).to be_a(Hash)
+              expect(subtitle).to include("title" => "the subtitle")
+              expect(subtitle).to include("title_type" => "Subtitle")
+            end
+          end
+
+          context "when authenticated as the submitter of the Work" do
```
jrgriffiniii, line 108: This seems to be the failing case. I can ensure that this is fully resolved.

jrgriffiniii:
I was failing to reproduce these errors locally, I am going to proceed with resolving the outstanding failing test.

app/controllers/works_controller.rb
```diff
@@ -114,8 +114,6 @@ def update
     if current_user.blank? || !@work.editable_by?(current_user)
       Rails.logger.warn("Unauthorized attempt to update work #{@work.id} by user #{current_user.uid}")
       redirect_to root_path
-    elsif @work.approved? && @work.submitted_by?(current_user)
```
carolyncole, line 117: The requirement that a user can not edit an item they deposited was ticketed here: https://github.com/pulibrary/pdc_describe/issues/255.  Does removing this code introduce a bug?

carolyncole (changes requested):
@jrgriffiniii I am unsure if by fixing this "bug" we are introducing a bug.  We should check to see if we still want to limit when a depositor can edit the metadata before this gets merged.

I have no technical issue with this PR and I am not really requesting any changes,  I just want to block it from being merged until we can discuss it with RDOS.

jrgriffiniii:
Given the discussion held on 01/20/23, it was confirmed by the Product Owners that submitters should *not* be able to modify metadata.

PR #853, opened 2023-01-12
Documenting the Mail setup on production

carolyncole:
Also documnenting mailcatcher for development
refs #687

PR #872, opened 2023-01-18
Setting MailCatcher as the default Mail setup for the development environment

jrgriffiniii:
Derived from https://github.com/pulibrary/pdc_describe/pull/853, please see https://github.com/pulibrary/pdc_describe/pull/853#discussion_r1073687851 for the specific context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,10 @@ We use DataCite to mint DOIs and in production you must to define the `DATACITE_
 RDSS uses the same [release and deployment process](https://github.com/pulibrary/rdss-handbook/blob/main/release_process.md) for all projects.
 
 ## Mail
+
+### Mail on Development
+Mailcatcher is a gem that can also be installed locally.  See the [mailcatcher documentation](https://mailcatcher.me/) for how to run it on your machine.
+
 ### Mail on Staging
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -72,6 +72,9 @@ We use DataCite to mint DOIs and in production you must to define the `DATACITE_
 
 RDSS uses the same [release and deployment process](https://github.com/pulibrary/rdss-handbook/blob/main/release_process.md) for all projects.
 
+## Mail on Development
```
carolyncole, line 75: This information would need to be up in the dependencies (https://github.com/pulibrary/pdc_describe/pull/872/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R15).  This change requires the user to have mailcatcher installed and running on their local machine

carolyncole (changes requested):
Looks good except the information about mail catcher would need to be put in as a development dependency.

carolyncole (approved):
@jrgriffiniii thank you for addressing my concern!  It was not my intention to change the Gemfile, but that definitely addresses my concern better than just moving the documentation.

PR #875, opened 2023-01-18
Converting the message to html before sending for html version of the email

### LINKED ISSUE #860: Convert the mark down in the email
Like we do on the work show page, we need to convert the markdown to html before sending it.
Screen Shot 2023-01-13 at 3 24 36 PM

carolyncole:
fixes #860


app/mailers/notification_mailer.rb
```diff
@@ -8,6 +8,7 @@ def build_message
 
     @subject = "[pdc-describe] New Notification"
     @message = @work_activity.message
```
mccalluc, line 10:
```suggestion
```
Might be unused now?

spec/mailers/notification_mailer_spec.rb
```diff
@@ -30,5 +30,27 @@
...
+        expect(message.to).to eq([user.email])
+        expect(message.from).to eq(["noreply@example.com"])
+        expect(message.subject).to eq("[pdc-describe] New Notification")
+        text_part = message.text_part
+        html_part = message.html_part
+        expect(text_part.content_type).to eq("text/plain; charset=UTF-8")
+        expect(html_part.content_type).to eq("text/html; charset=UTF-8")
+        expect(html_part.encoded).to include("Hello #{user.display_name},")
+        expect(html_part.encoded).to include("To view the notification, please browse to http://www.example.com/works/#{work.id}.")
+        expect(html_part.encoded).to include("I like to send &lt;a href=&quot;www.google.com&quot;&gt;links&lt;/a&gt;")
```
mccalluc, line 49:
```suggestion
      let(:work_activity) { FactoryBot.create(:work_activity, work: work, message: "I like to send [links](https://www.google.com)") }
      it "generates the e-mail message" do
        expect(message_delivery).to be_a(ActionMailer::Parameterized::MessageDelivery)
        expect(message_delivery.message).to be_a(Mail::Message)
        message = message_delivery.message
        expect(message.to).to eq([user.email])
        expect(message.from).to eq(["noreply@example.com"])
        expect(message.subject).to eq("[pdc-describe] New Notification")
        text_part = message.text_part
        html_part = message.html_part
        expect(text_part.content_type).to eq("text/plain; charset=UTF-8")
        expect(html_part.content_type).to eq("text/html; charset=UTF-8")
        expect(html_part.encoded).to include("Hello #{user.display_name},")
        expect(html_part.encoded).to include("To view the notification, please browse to http://www.example.com/works/#{work.id}.")
        expect(html_part.encoded).to include("I like to send &lt;a href=&quot;https://www.google.com&quot;&gt;links&lt;/a&gt;")
```

Doesn't change the substance of the test, but the link should include the protocol.

mccalluc, line 49: Is there a reason for `encoded`?

spec/mailers/notification_mailer_spec.rb
```diff
@@ -30,5 +30,27 @@
...
+        expect(text_part.encoded).to include(work_activity.message)
+        expect(text_part.encoded).to include("Hello #{user.display_name},")
+        expect(text_part.encoded).to include("To view the notification, please browse to http://www.example.com/works/#{work.id}.")
```
mccalluc, line 52: Are we expecting the text to include "I like to send ..." markdown? Or are we expecting it not to include it? I'd suggest adding an assertion one way or the other, to make sure we're getting the behavior we're expecting.

mccalluc (commented):
A few comments -- To me, making an assertion about what we expect to see in the text part of the email seems the most valuable.

app/mailers/notification_mailer.rb
carolyncole: This is still used in the text portion of the email and still needed

spec/mailers/notification_mailer_spec.rb
carolyncole: Yes, this is done here: https://github.com/pulibrary/pdc_describe/pull/875/files#diff-9860398629ad665126bc447cc0b151e236450de156e8bc64a313c18db62bc7c3R50

spec/mailers/notification_mailer_spec.rb
carolyncole: I can add the protocol...  I'l make the change locally, since I do not understand why you changed every line of the test in this commit.

bess (approved):

mccalluc (approved):

Issue #869, opened 2023-01-18
Add funders to migration specs


bess:
When we wrote the migration specs we didn't yet have a space for the funders.

## Acceptance Criteria
- [x] Go through the migration specs and add funder information
  - [x] attention
  - [x] baldwin
  - [x] bitklavier
  - [x] bitklavier_image
  - [x] cklibrary
  - [x] cytoskeletal
  - [x] femosecond
  - [x] flume
  - [x] sowingseeds
- [x] Extend the way we test for ORCID integration to include RORs for funding data so this can be tested: https://github.com/pulibrary/pdc_describe/blob/main/spec/support/orcid_specs.rb

PR #878, opened 2023-01-18
Add funders to migration specs

bess:
Work toward https://github.com/pulibrary/pdc_describe/issues/869


spec/fixtures/dataspace_migration/attention/Attention_Awareness_Dorsal_Attention_README.txt
```diff
@@ -0,0 +1,155 @@
+This Attention_Awareness_Dorsal_Attention_readme.txt file was generated on 2020-11-17 by Andrew I. Wilterson
```
mccalluc, line 1: You might consider just naming this `README.txt`: I think it will then be shown when looking at the directory in github. Or there might already be a convention that dictates this pattern.

spec/system/data_migration/femtosecond_form_submission_spec.rb
```diff
@@ -99,6 +126,10 @@
       femtosecond_work = Work.last
       expect(femtosecond_work.title).to eq title
 
+      # Check that RORs were persisted as funder names
+      funders = femtosecond_work.resource.funders.map(&:funder_name).uniq
+      expect(funders).to contain_exactly("United States Department of Energy", "National Science Foundation", "Deutsche Forschungsgemeinschaft")
```
mccalluc, line 131: Normally, we want tests to avoid dependencies on outside services, but this might be an exception: If the ROR API turns out to be flaky, it might be good to surface that, even if could mean sporadic test failures.

mccalluc (approved):

Issue #870, opened 2023-01-18
Error in JSON representation


hectorcorrea:
This record is throwing an HTTP 500: https://pdc-describe-staging.princeton.edu/describe/works/129.json

The record shows OK in HTML: https://pdc-describe-staging.princeton.edu/describe/works/129

https://app.honeybadger.io/projects/99214/faults/92629043

Carolyn found that the problem might be happening only for curated files. The json loads for un-curated items.

ActiveStorage is utilized for pre curated files, since when files are uploaded via the form they are done with active storage.  Post curation files are only in S3.

The json should include:  Please do not download the files to determine this information.  It should be coming from AWS.

- [ ] file name
- [ ] size
- [ ] url to the file in Globus for post curated

## Completion Criteria:

- [ ] json to load for post-curated works including files
- [ ] json to load for pre-curated works, but not include the files
- [ ] An automated test that proves that json loads for pre-curated works
- [ ] An automated test that proves that json loads for post-curated works including files

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-Workcycles-61a4f1a12a399b001730f65a/issues/pulibrary/pdc_describe/870?planning-poker) with Zenhub @bess @hectorcorrea @jrgriffiniii @kelynch @mccalluc

mccalluc:
I can start this, but the honey badger link doesn't work for me. Is there a typo, or am I missing privs?

mccalluc:
Largely addressed by
- #879 

but 
- it does not include globus
- wait to close until we can confirm that the integration works.

PR #879, opened 2023-01-18
Remove files from precuration json; Add files to postcuration json

mccalluc:
- Towards #870... but does not fix because we still need to provide the globus URL.

Of note, I think the s3 objects we see at this point in the code are our own, rather that ones from the AWS SDK, and so only some properties are available, and names have been mapped:
```
    model_uploads.each do |attachment|
      s3_file = S3File.new(query_service: self,
                           filename: attachment.key,
                           last_modified: attachment.created_at,
                           size: attachment.byte_size,
                           checksum: attachment.checksum)
```


mccalluc:
Not sure how best to mock s3... Doing more research.

hectorcorrea (approved):

PR #891, opened 2023-01-23
Move hard-coded list of users out of ERB and into configuration; Add documentation

### LINKED ISSUE #849: Free text provenance permission should be managed in a config file
Only users who are explicitly named in the application can enter free-text provenance notes. In the first pass at this feature, the list of users with this ability is hard-coded. Change this to a config file and document how to update it so that when we have staff turnover it can be easily updated. Maybe it would make sense to combine this with "superadmin" status? We might need some feedback from stakeholders here.

Screenshot 2023-01-12 at 8 04 42 AM

## Acceptance Criteria
- [ ] There is a config file in the config directory containing netids of users who are allowed to enter free text provenance notes
- [ ] There is documentation in the docs directory, with a pointer from the README (or the RDSS handbook?) describing how to update it.

mccalluc:
- Fix #849

That issue specifies the implementation, but as I re-read it, it also proposes to combine this with admin privs. I don't care strongly which approach is taken.

- This with have a merge conflict with #857: That is older and more complicated, so I'd prefer to get that reviewed and merged before this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,8 @@ To stop database services:
 2. Enter the rails console: `bundle exec rails console`
 3. Elevate your privs: `User.new_super_admin("your NetID here")`
 
+Separate from admin, users may also be [given permission](docs/how_to_add_postdated_provenance.md) to add post-dated notes in the provenance / change log.
+
 ### Development with AWS resources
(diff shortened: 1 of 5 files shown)


hectorcorrea (dismissed):

hectorcorrea:
@mccalluc looks good - could you take care of the merge conflict?

mccalluc:
resolving conflict locally...

hectorcorrea (approved):

PR #894, opened 2023-01-23
Send just one notification if creator is also admin

### LINKED ISSUE #855: If the depositor is a curator they get two notifications
During the demo Kate got 4 notifications instead of two, because they are both a curator and the depositor.

https://github.com/pulibrary/pdc_describe/blob/main/app/models/work_state_transition_notification.rb#L16

Should utilize logical and (&) or .unique

Acceptance criteria:

- [ ] if the depositor is a curator they should get one notification
- [ ] There is an automated test for when the user is the depositor and the collection curator.

mccalluc:
- Fix #855
- Refactor test to cover four cases: 2 work states (associated with a particular transition method to call) times 2 admin conditions (creator is or is not admin)

### DIFF
--- a/app/models/work_state_transition_notification.rb
+++ b/app/models/work_state_transition_notification.rb
@@ -13,7 +13,7 @@ class WorkStateTransitionNotification
     @collection_administrators = work.collection.administrators.to_a
     @work_url = Rails.application.routes.url_helpers.work_url(work)
     @work_title = work.title
-    @users = @collection_administrators + [@depositor]
+    @users = @collection_administrators | [@depositor] # Depositor may also be an admin, but should only be listed once.
     @notification = notification_for_transition
     @id = work.id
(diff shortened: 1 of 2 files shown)


app/models/work_state_transition_notification.rb
```diff
@@ -13,7 +13,7 @@ def initialize(work, current_user_id)
     @collection_administrators = work.collection.administrators.to_a
     @work_url = Rails.application.routes.url_helpers.work_url(work)
     @work_title = work.title
-    @users = @collection_administrators + [@depositor]
+    @users = @collection_administrators | [@depositor] # Depositor may also be an admin, but should only be listed once.
```
hectorcorrea, line 16: This is pretty neat, I didn't know you could use `|` like this to prevent duplicates.

PR #898, opened 2023-01-24
Adding a license page to the welcome controller

### LINKED ISSUE #897: License information
The text of the license terms that users agree to when submitting items to PDC Describe is available to the public without signing in.

Approved license text [here](https://docs.google.com/document/d/1aEvbsMFJ8ZD1TEBieX6tcYO8WB4TdAtXs70r2EGrui4/edit).

Related to https://github.com/pulibrary/pdc_describe/issues/764

## Acceptance Criteria
- [ ] Create a static content page containing the text of the license
- [ ] The license page can be accessed without signing into any system

carolyncole:
Also change the work license to a partial so we can reuse it in welcome fixes #897

Screen Shot 2023-01-24 at 1 28 03 PM

### DIFF
--- a/app/views/works/_license_text.html.erb
+++ b/app/views/works/_license_text.html.erb
@@ -1,0 +1,50 @@
+<p>
+In order for Princeton University's Princeton Data Commons to reproduce, translate, and distribute
+your submission for worldwide public access your agreement to the following terms is necessary.
+Please take a moment to read the terms of this license.
+</p>
+
+<blockquote style="border: 1px solid; padding: 1em;">
+<p>
+  By clicking through this license, you (the author(s) or creator(s) or copyright owner(s)) grant to Princeton University the non-exclusive right to reproduce, translate (as defined below), and/or distribute your submission (including the abstract) worldwide in electronic format.
+</p>
+<p>
+  You represent that the submission is your original work, and/or that you have
+  the right to grant the rights contained in this license.  You represent that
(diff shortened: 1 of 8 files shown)


app/views/works/_license_text.html.erb
```diff
@@ -0,0 +1,50 @@
+<p>
+In order for Princeton University's Princeton Data Commons to reproduce, translate, and distribute
+your submission for worldwide public access your agreement to the following terms is necessary.
+Please take a moment to read the terms of this license.
+</p>
```
mccalluc, line 5: As I understand it, this page will be linked from PDC Discovery, where it will describe the terms under which access is granted to this object. In that context, I don't think "your submission" makes sense.

I would leave this paragraph in review.html.erb, and perhaps replace it here with a different paragraph addressed to the end-user.

mccalluc (changes requested):
To me, it seems like a different introduction for a different audience would make sense... but if this doesn't resonate, and someone else can approve, don't let this block.

mccalluc (approved):
Reconsidering... I think the default is to approve PRs to move them along, but if there are remaining questions, put them on the agenda? I'll do that.

kelynch:
I personally think it makes sense to hold off on merging if the PR review is requesting changes to the code that we're proposing adding to the code base in that PR itself, however a follow-up review and (if appropriate) approval should then be made if/when the requested changes are made in the PR.

PR #899, opened 2023-01-24
Move messages up to sidebar on the right which can be toggled

### LINKED ISSUE #823: Messages: button and "right-side" messages

mccalluc:
- Fix #823 

I'm a little uncomfortable with more ad-hoc jquery, but I think this is what is requested.

Defaults to open:
Screen Shot 2023-01-24 at 3 47 40 PM

Can be closed:
Screen Shot 2023-01-24 at 3 47 52 PM

### DIFF
--- a/app/assets/stylesheets/components/_container.scss
+++ b/app/assets/stylesheets/components/_container.scss
@@ -19,7 +19,6 @@
   .btn-primary {
     background-color: #f57e21;
     border-color: #914810;
-    margin: 20px 0px;
   }
   .btn-primary:hover {
     background-color: #914810;
--- a/app/views/works/_work_activity_messages.html.erb
+++ b/app/views/works/_work_activity_messages.html.erb
@@ -23,7 +23,7 @@
 
   <div id="new-message-section">
     <%= form_with url: add_message_work_path(@work) do |f| %>
-      <textarea id="new-message" name="new-message" class="new-message" placeholder="leave a message" rows="5" cols="100"></textarea>
+      <textarea id="new-message" name="new-message" class="new-message" placeholder="leave a message" rows="5" cols="30"></textarea>
       <br/>
       <%= f.submit("Message", class: "btn btn-secondary") %>
       <span class="new-message-help">
--- a/app/views/works/_work_activity_messages.html.erb
+++ b/app/views/works/_work_activity_messages.html.erb
@@ -1,5 +1,4 @@
 <div>
-  <h2>Messages</h2>
   <% if @messages.size == 0 && !@work.submission_notes.present? %>
     No messages
   <% end %>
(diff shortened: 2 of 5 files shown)


mccalluc:
I should add tests. Back to draft.

app/assets/stylesheets/components/_container.scss
```diff
@@ -19,7 +19,6 @@
   .btn-primary {
     background-color: #f57e21;
     border-color: #914810;
-    margin: 20px 0px;
```
mccalluc, line 22: This style on one button was giving the whole button bar more height, but it didn't align with the new column added in this PR.

app/views/works/_work_activity_messages.html.erb
```diff
@@ -1,5 +1,4 @@
 <div>
-  <h2>Messages</h2>
```
mccalluc, line 2: Not in mockup: The button above may be enough of a label.

spec/system/work_create_spec.rb
```diff
@@ -109,11 +109,17 @@
...
       visit(work_path(Work.last))
 
+      has_been_created_message = "#{title} has been created"
       within("ul.work-messages") do
-        expect(page).to have_content("#{title} has been created")
+        expect(page).to have_content(has_been_created_message)
         expect(page).to have_content("#{title} is ready for review")
       end
 
+      click_on "Hide Messages"
```
carolyncole, line 118: ❤️

carolyncole (approved):

Issue #902, opened 2023-01-25
Include Globus URL in JSON endpoint 


kelynch:
Advances #870

PR #904, opened 2023-01-25
Adds Globus URL for each file in JSON endpoint

hectorcorrea:
Screenshot 2023-01-25 at 3 07 53 PM

Moves forward #902

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,8 @@ Then source this file before starting rails:
 $ . ~/s3-envvars.sh; bundle exec rails s -p 3000
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -67,6 +67,8 @@ Then source this file before starting rails:
 $ . ~/s3-envvars.sh; bundle exec rails s -p 3000
 ```
 
+By default `storage.yml` is also configured to use local storage instead of connecting to AWS. Make sure you update your local copy as indicated inside the file if you want to use AWS.
```
hectorcorrea, line 70: This is not technically needed but I keep running into issues because my local env is not uploading all the way to AWS so I wanted to document it.

carolyncole (approved):

PR #910, opened 2023-01-26
Adds support for domains

### LINKED ISSUE #905: Domain field
Add a new `domain` field in the Work (under "Additional Metadata"). The field will be a multi-select field from a controlled vocabulary. The controlled vocabulary will be provided by RDOS, but we can start with the domain values that we are [currently using in Discovery](https://github.com/pulibrary/pdc_discovery/blob/main/lib/traject/domain.rb).

This field should be indexed in PDC Discovery. We already have a field for it in Discovery but the existing field is *single value* and we will need to define a new multi-value field.

(source: https://docs.google.com/document/d/1czy1qAs8P-sMm_7T3zAv5oAiFvYI3Ij4pR5JsTbcpdo/edit#, under January 13, 2023)

hectorcorrea:
Adds support for a multi-value domain field with a controlled vocabulary in the Additional Metadata section of the Work. 

The current values in the controlled vocabulary were copied from what were using for DataSpace records in PDC Discovery but we could easily expand that list in a future PR once we get the official list from RDOS.

Screenshot 2023-01-26 at 4 49 51 PM

Closes #905

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -22,6 +22,13 @@
     <!-- Provides https://jqueryui.com/sortable/ and /autocomplete/-->
     <script src="https://code.jquery.com/ui/1.13.1/jquery-ui.js"></script>
 
+    <!-- Reference bootstrap-select for the modern looking multi-select dropdown list
+      Notice that we are using a beta version of this component since that's the one
+      supported with Bootstrap 5 (https://stackoverflow.com/a/67949799/446681).
+    -->
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-select/1.14.0-beta2/css/bootstrap-select.min.css" integrity="sha512-mR/b5Y7FRsKqrYZou7uysnOdCIJib/7r5QeJMFvLNHNhtye3xJp1TdJVPLtetkukFn227nKpXD9OjUc09lx97Q==" crossorigin="anonymous" referrerpolicy="no-referrer" />
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-select/1.14.0-beta2/js/bootstrap-select.min.js" integrity="sha512-FHZVRMUW9FsXobt+ONiix6Z0tIkxvQfxtCSirkKc5Sb4TKHmqq1dZa8DphF0XqKb3ldLu/wgMa8mT6uXiLlRlw==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
+
     <link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/1.11.3/css/jquery.dataTables.min.css"/>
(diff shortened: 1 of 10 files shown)


app/views/layouts/application.html.erb
```diff
@@ -22,6 +22,13 @@
     <!-- Provides https://jqueryui.com/sortable/ and /autocomplete/-->
     <script src="https://code.jquery.com/ui/1.13.1/jquery-ui.js"></script>
 
+    <!-- Reference bootstrap-select for the modern looking multi-select dropdown list
+      Notice that we are using a beta version of this component since that's the one
+      supported with Bootstrap 5 (https://stackoverflow.com/a/67949799/446681).
+    -->
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-select/1.14.0-beta2/css/bootstrap-select.min.css" integrity="sha512-mR/b5Y7FRsKqrYZou7uysnOdCIJib/7r5QeJMFvLNHNhtye3xJp1TdJVPLtetkukFn227nKpXD9OjUc09lx97Q==" crossorigin="anonymous" referrerpolicy="no-referrer" />
```
mccalluc, line 29: Up till now, I don't think we've had any dependencies on cloudflare. Do we have a general policy about dependence on CDNs?

mccalluc (commented):
I'm concerned about introducing an extra dependency where it's not needed. I would suggest either a native [multi-select](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/multiple), or checkboxes.

Making this a "Comment": If someone else approves, I shouldn't block.

app/views/layouts/application.html.erb
hectorcorrea: @mccalluc We have plenty of CDN dependencies already on that file: https://github.com/pulibrary/pdc_describe/blob/main/app/views/layouts/application.html.erb#L11-L26

mccalluc: I would like to defer to another developer for review. For me, adding a new CDN, "cloudflare" in this case, adds a new point of fragility. Maybe they are all s3 behind the curtain, or maybe not.

Separately: I don't see the additional functionality offered by this plugin as being worth the incremental cost. Checkboxes work well and don't require outside dependencies.

But this is just my opinion: If someone else feels comfortable, please approve.

hectorcorrea: OK. I removed the third party and went with the built-in

hectorcorrea:
@mccalluc I went ahead and removed the third party and switched to the built-in

app/views/layouts/application.html.erb
mccalluc: ... but I think there's also a larger tech debt issue that I'm concerned about. I'll get my thoughts in order and put something on the agenda.

mccalluc (approved):

PR #912, opened 2023-01-27
Clean up utility code to avoid JS injection

### LINKED ISSUE #906: Clean up html encoding and copy-and-paste in `application_helper.rb`
In the course of #900, found problems with the utils in `application_helper.rb`
- We take raw data and concatenate it into html without first html-encoding it.
- Copy-and-paste between `contributor_link` and `person_orcid_link` should be cleaned up.
- Consider moving handling of separator up and out with `safe_join`

mccalluc:
- Fix #906

### DIFF
--- a/app/views/works/_orcid_link.html.erb
+++ b/app/views/works/_orcid_link.html.erb
@@ -1,0 +1,15 @@
+<% display_name = capture do %>
+  <%= person.value %>
+  <% if person.type.present? %>
+    (<%= person.type %>)
+  <% end %>
+<% end %>
+
+<span class="author-name">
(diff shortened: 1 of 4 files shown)


hectorcorrea (approved):

app/views/works/_orcid_link.html.erb
```diff
@@ -0,0 +1,15 @@
+<% display_name = capture do %>
+  <%= person.value %>
+  <% if person.type.present? %>
+    (<%= person.type.titleize %>)
+  <% end %>
+<% end.strip %>
```
hectorcorrea, line 6: nice

PR #916, opened 2023-01-30
Remove params as we read them; flag unexpected params

mccalluc:
- Prompted by #900

Just an idea: If we add a form field, but nothing reads it, that's a problem. By deleting params as we read them, we can detect if that is happening and throw an exception. Draft for now.


mccalluc:
This complexity this would add is not worth the rare error it would catch.

PR #918, opened 2023-01-31
changed vpill code in migration specs to Additional Metadata and/or Curator Controlled

Twade968:
change code to "additional metadata" instead of the "v-pills-additional-tab"
change code to "Curator Controlled" instead of "v-pills-curator-controller-tab"


bess (approved):
⭐ 😄

Issue #913, opened 2023-01-30
Migration spec missing ark


bess:
The migration spec for Sowing Seeds has an ark in the setup, but it is never used. This leads to an error when someone tries to approve the work: https://app.honeybadger.io/projects/99214/faults/92882814

See: https://github.com/pulibrary/pdc_describe/blob/main/spec/system/data_migration/sowingseeds_form_submission_spec.rb#L13

## Acceptance criteria
- [ ] The ark is updated to use the expected format (`ark:/88435/dsp01d791sj97j`)
- [ ] The spec is updated so that the ark is submitted via the form
- [ ] Check all tests for ark formatting (use the ark:/ prefix) and that the ark is populated

PR #919, opened 2023-01-31
Fixing arks in migration specs

Twade968:
Fixes for ticket #913


Twade968:
Thank you so much Carolyn! All checks are passing now!

PR #944, opened 2023-02-02
Add XML fixtures for migration specs

### LINKED ISSUE #917: For all data migration specs, check in Datacite XML fixtures and make sure the output is what we expect
Each test in `spec/system/data_migration` should have a corresponding xml fixture, and we should confirm that the Datacite produced by the actions described is what we expect. At least two kinds of problems we'd like to catch:

- Race conditions in the browser-based tests: If some percentage of the time the data entry does not reliably produce the record we expect, we should be able to detect that.
- Bugs in code not populating a given field. We have a fair number of fields, and we're adding new ones: If one of our fields (perhaps multivalued) doesn't actually go through to the Datacite, we should catch that.

mccalluc:
- Fix #917


spec/system/data_migration/femtosecond_form_submission_spec.rb
```diff
@@ -138,6 +138,7 @@
       # # This will allow us to evolve our local datacite standards and test our records against them.
       # datacite = PDCSerialization::Datacite.new_from_work(femtosecond_work)
       # expect(datacite.valid?).to eq true
+      # expect(datacite.to_xml).to be_equivalent_to(File.read("spec/system/data_migration/femtosecond.xml"))
```
bess, line 141: @mccalluc Why is this commented out?

bess (approved):
@mccalluc Looks great and I'm glad to merge as-is, but I'm curious why the test you added for one of the specs is commented out?

spec/system/data_migration/femtosecond_form_submission_spec.rb
bess: Nevermind. Just looked more closely and I realize this is commented out because of the ROR network call that I'm trying to fix right now. 🤦🏻‍♀️

PR #947, opened 2023-02-03
Only allow uploading 20 files from the UI

carolyncole:
Allow as many files as a person wants if they are attached via s3 or Globus
Files less than 20 chosen
Screenshot 2023-02-03 at 8 16 19 AM
Files more than 20 chosen
Screenshot 2023-02-03 at 8 16 07 AM

This PR also introduces a concept I see over in [Orangelight ](https://github.com/pulibrary/orangelight/blob/main/app/javascript/entrypoints/application.js) for the vite javascript loading.

If I am reading Orangelight correctly we could move all of our [sprokets javascript](https://github.com/pulibrary/pdc_describe/tree/main/app/assets/javascripts) over and utilize the concept of a UI loader to initialize our custom javascript.


app/javascript/entrypoints/pdc/maximum_file_upload.es6
```diff
@@ -0,0 +1,24 @@
+export default class MaximumFileUpload {
+
+    constructor(upload_id, save_id, error_id = "file-error") {
+      this.upload_element = $("#"+upload_id);
+      this.save_element = document.getElementById(save_id);
+      this.error_element = document.getElementById(error_id)
```
mccalluc, line 6: A convention that I've used is to distinguish jQuery objects by prefixing with `$`. Here, things are small enough that it's unlikely we'd get confused, but if there's more code, it's possible to forget what kind of "element" we're dealing with. I'll put a little item on the agenda.

mccalluc (approved):

mccalluc:
Moving towards ESM is great! In this case, another possibility would be to put both classes in one file, and only export the one we need.

PR #951, opened 2023-02-03
Moving DOI copy to clipboard into Vite managed class

carolyncole:
Basically copied and pasted the code, converted it to es6, and wrapped it in a class

### DIFF
--- a/app/javascript/entrypoints/pdc/copy_to_clipboard.es6
+++ b/app/javascript/entrypoints/pdc/copy_to_clipboard.es6
@@ -1,6 +1,20 @@
-$(() => {
+/* eslint class-methods-use-this: ["error",
+  { "exceptMethods": ["setCopiedToClipboard", "resetCopyToClipboard", "errorCopyToClipboard"] }] */
+
+export default class CopytoClipboard {
+  attach_copy() {
+    // Setup the DOI's COPY button to copy the DOI URL to the clipboard
+    $('#copy-doi').click(this.copy_doi.bind(this));
+  }
+
+  copy_doi() {
+    var doi = $('#copy-doi').data('url');
+    this.copyToClipboard(doi, '#copy-doi-icon', '#copy-doi-label', 'copy-doi-label-normal', 'copy-doi-label-copied');
+    return false;
+  }
+
   // Sets the elements to the proper CSS classes once a value has been copied to the clipboard.
-  function setCopiedToClipboard(iconEl, labelEl, normalClass, copiedClass) {
+  setCopiedToClipboard(iconEl, labelEl, normalClass, copiedClass) {
     $(iconEl).removeClass('bi-clipboard');
     $(iconEl).addClass('bi-clipboard-check');
     $(labelEl).text('COPIED');
@@ -9,7 +23,7 @@ $(() => {
   }
 
   // Resets the elements to the proper CSS classes (e.g. displays as if the copy has not happened)
-  function resetCopyToClipboard(iconEl, labelEl, normalClass, copiedClass) {
+  resetCopyToClipboard(iconEl, labelEl, normalClass, copiedClass) {
     $(labelEl).text('COPY');
     $(labelEl).removeClass(copiedClass);
     $(labelEl).addClass(normalClass);
@@ -18,7 +32,7 @@ $(() => {
   }
 
   // Sets icon and label to indicate that an error happened when copying a value to the clipboard
-  function errorCopyToClipboard(iconEl, errorMsg) {
+  errorCopyToClipboard(iconEl, errorMsg) {
     $(iconEl).removeClass('bi-clipboard');
     $(iconEl).addClass('bi-clipboard-minus');
     console.log(errorMsg);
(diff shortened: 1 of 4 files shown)


mccalluc (changes requested):
I should have noticed this earlier, but the `es6` extension isn't being picked up by eslint. It will flag some issues, but it also has a `--fix` option that should handle a lot of it.

bess (dismissed):

bess:
I'm not actually requesting a review, but I can't figure out how to un-select it. Feel free to merge without my input.

app/javascript/entrypoints/pdc/copy_to_clipboard.es6
```diff
@@ -0,0 +1,62 @@
...
+        return false;
+    }
+    
+    // Sets the elements to the proper CSS classes once a value has been copied to the clipboard.
+    setCopiedToClipboard(iconEl, labelEl, normalClass, copiedClass) {
+      $(iconEl).removeClass('bi-clipboard');
+      $(iconEl).addClass('bi-clipboard-check');
+      $(labelEl).text('COPIED');
+      $(labelEl).removeClass(normalClass);
+      $(labelEl).addClass(copiedClass);
```
mccalluc, line 20:
```suggestion
      $(iconEl).removeClass('bi-clipboard')
        .addClass('bi-clipboard-check');
      $(labelEl).text('COPIED')
        .removeClass(normalClass)
        .addClass(copiedClass);
```
method chaining is a jquery idiom.

app/javascript/entrypoints/pdc/copy_to_clipboard.es6
```diff
@@ -0,0 +1,62 @@
...
+  
+    // Copies a value to the clipboard and notifies the user
+    // value - value to copy to the clipboard
+    // iconEl - selector for the HTML element with the clipboard icon
+    // labelEl - selector for the HTML element with the COPY label next to the icon
+    // normalClass - CSS to style the label with initially
+    // copiedClass - CSS to style the label with after a value has been copied to the clipboard
+    // iconEl and labelEl could be any jQuery valid selector (e.g. ".some-id" or a reference
+    //  to an element)
+    copyToClipboard(value, iconEl, labelEl, normalClass, copiedClass) {
```
mccalluc, line 47:
```suggestion
    copyToClipboard({value, iconEl, labelEl, normalClass, copiedClass}) {
```
For longer lists of parameters, consider destructuring instead. JS doesn't complain about having the wrong number of parameters, so long parameter lists are even more bug-prone than in ruby.

app/javascript/entrypoints/pdc/pdc_ui_loader.es6
```diff
@@ -8,5 +9,6 @@ export default class PdcUiLoader {
   setup_fileupload_validation() {
     new MaximumFileUpload("patch_pre_curation_uploads","file-upload");
     new MaximumFileUpload("pre_curation_uploads","btn-submit");
+    new CopytoClipboard();
```
mccalluc, line 12: There will be an eslint warning about using a constructor just for side effects.

app/javascript/entrypoints/pdc/copy_to_clipboard.es6
carolyncole: @mccalluc Can changing the code be a separate PR?  This PR was a direct copy to move it into Vite vs sprokets.  I would hate to look back and think these changes were needed to get Vite to compile the code.

carolyncole:
@mccalluc I think that cleaning up javascript style should either be done before or after this PR.  Again this was just a direct copy to move the code into Vite.  I would like to keep the change as simple as possible to give an example of how to move the code.  In general I think the project JavaScript is a mess.  So making updates to it's style are fine by me.  I do think that putting javascript in the pages, in Sprokets and in Vite is just a recipe for disaster.  My goal with this PR, was to make the code a bit better by heading to all javascript in Vite.  We could certainly ticket the updates you suggest.

app/javascript/entrypoints/pdc/copy_to_clipboard.es6
carolyncole: In addition I do not believe that it is good coding style to cram more than one action onto a line just because the code allows it.  Adding more than one action into a line makes it hard to understand what is happening when you read the code in the future.

mccalluc: I won't block this, but I think there are cases where method chaining makes sense, but I'll try to summarize that in the meeting agenda.

app/javascript/entrypoints/pdc/copy_to_clipboard.es6
mccalluc: I'll make a separate issue for this

app/javascript/entrypoints/pdc/copy_to_clipboard.es6
```diff
@@ -1,6 +1,20 @@
-$(() => {
+/* eslint class-methods-use-this: ["error",
+  { "exceptMethods": ["setCopiedToClipboard", "resetCopyToClipboard", "errorCopyToClipboard"] }] */
```
mccalluc, line 2: I'll file a followup issue for these inline ignores. Thank you for handling it.

mccalluc (approved):

PR #953, opened 2023-02-04
Test flash messages

### LINKED ISSUE #943: Write test confirm that flash messages are being rendered on homepage
Anecdotally, developers report landing on the home page after saving a record and not seeing an error message in the flash UI. Maybe the homepage template is missing the flash slot? It also seems that the flash is just not being set in all cases.
- [ ] Make sure flash is being set any page we redirect to the homepage as an error.
- [ ] Write tests to confirm that the flash is rendered.

```ruby
 def edit
    if current_user && @work.editable_by?(current_user)
      if @work.approved? && !@work.administered_by?(current_user)
        redirect_to root_path, notice: I18n.t("works.approved.uneditable")
      else
        @uploads = @work.uploads
        @wizard_mode = wizard_mode?
        render "edit"
      end
    else
      Rails.logger.warn("Unauthorized attempt to edit work #{@work.id} by user #{current_user.uid}")
      redirect_to root_path # TODO: add flash
    end
  end
```

mccalluc:
- Fix #943
- Add system tests for flash messages. (Previously, the only tests of flash messages were controller tests... so they didn't fail, even if they didn't make it to the screen.) (Tried for a bit to make a lower-level view test, but never got the wiring sorted out... and view tests aren't our default, in any case.)
- In two cases, we weren't actually sending the right message: fix it.
- Remove redundant rendering of flash messages: It should only be done in one place.

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -98,7 +98,7 @@ class WorksController < ApplicationController
     if current_user && @work.editable_by?(current_user)
       if @work.approved? && !@work.administered_by?(current_user)
         Honeybadger.notify("Can not edit work: #{@work.id} is approved but #{current_user} is not admin")
-        redirect_to root_path, notice: I18n.t("works.approved.uneditable")
+        redirect_to root_path, notice: I18n.t("works.uneditable.approved")
       else
         @uploads = @work.uploads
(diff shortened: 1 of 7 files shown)


bess (approved):
Thank you so much for finding these bugs and fixing this!

PR #954, opened 2023-02-06
Adding .es6 files to the eslint

carolyncole:
Also ran eslint with --fix to clean up the mess
```
yarn run eslint app/assets app/javascript --ext .js,.jsx,es6 --fix
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,7 +64,7 @@ jobs:
       - install_dependencies
       - run:
           name: Run eslint
-          command: yarn run eslint app/assets app/javascript    
+          command: yarn run eslint app/assets app/javascript --ext .js,.jsx,es6   
 
   test:
     working_directory: *root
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -64,7 +64,7 @@ jobs:
       - install_dependencies
       - run:
           name: Run eslint
-          command: yarn run eslint app/assets app/javascript    
+          command: yarn run eslint app/assets app/javascript --ext .js,.jsx,es6
```
mccalluc, line 67:
```suggestion
          command: yarn run eslint app/assets app/javascript --ext .js,.jsx,.es6   
```
I'm not sure that it matters. Also wonder if there's a way to add this to the config, so we get the right behavior by default. I'll file a separate issue for that.

mccalluc: Filed #955 for follow-up.

mccalluc (approved):
Thanks. Just waiting for CI to complete.

PR #962, opened 2023-02-07
Fetch pre-curation files from directly from AWS 

### LINKED ISSUE #941: Load file list directly from AWS
In the show page load the file list directly from AWS rather than using ActiveStorage to cache the information and then display it from ActiveStorage's cache. 

The reason for this is that loading the ActiveStorage cache is too slow. @carolyncole has found that it can take 7+ minutes to cache 1000 files (see https://github.com/pulibrary/pdc_describe/issues/933) whereas AWS returns the entire list of files in just a couple of seconds.

hectorcorrea:
Re-route the call to always fetch pre-curation files directly from AWS, bypassing ActiveStorage. 

Fixes #941

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -201,7 +201,7 @@ class Work < ApplicationRecord
         key: upload.key,
         filename: upload.filename.to_s,
         created_at: upload.created_at,
-        url: rails_blob_path(upload, disposition: "attachment")
+        url: upload.url
       }
     end
   end
@@ -352,7 +352,12 @@ class Work < ApplicationRecord
   def uploads
     return post_curation_uploads if approved?
 
-    pre_curation_uploads
+    pre_curation_uploads_fast
+  end
+
+  # Fetches the data from S3 directly bypassing ActiveStorage
+  def pre_curation_uploads_fast
+    s3_query_service.client_s3_files.sort_by(&:filename)
   end
 
   # This ensures that new ActiveStorage::Attachment objects can be modified before they are persisted
@@ -472,6 +477,12 @@ class Work < ApplicationRecord
 
   delegate :ark, :doi, :resource_type, :resource_type=, :resource_type_general, :resource_type_general=,
            :to_xml, to: :resource
+
+  # S3QueryService object associated with this Work
+  # @return [S3QueryService]
+  def s3_query_service
+    @s3_query_service ||= S3QueryService.new(self, !approved?)
+  end
 
   protected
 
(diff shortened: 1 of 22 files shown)


app/services/s3_query_service.rb
```diff
@@ -136,19 +148,18 @@ def client_s3_files(reload: false)
       resp = client.list_objects_v2({ bucket: bucket_name, max_keys: 1000, prefix: prefix })
       resp_hash = resp.to_h
       objects = parse_objects(resp_hash)
-      parse_continuation(resp_hash, objects)
+      objects += parse_continuation(resp_hash)
```
hectorcorrea, line 151: @carolyncole re-wrote this since it was not really appending before. I am not sure why, I would have expected it to append to the objects array we were passing as a parameter but it didn't so I just force the append `+=` here.

app/models/work.rb
```diff
@@ -360,11 +360,6 @@ def pre_curation_uploads_fast
     s3_query_service.client_s3_files
   end
 
-  # Fetches the data from S3 directly bypassing ActiveStorage
-  def pre_curation_uploads_fast
-    s3_query_service.client_s3_files
-  end
-
```
hectorcorrea, line 357: @carolyncole This method was duplicated. I removed the duplicate **there** is still one `pre_curation_uploads_fast` in the code.

app/services/s3_query_service.rb
```diff
@@ -148,21 +148,7 @@ def client_s3_files(reload: false)
...
-        objects = parse_objects(resp_hash)
-
-        while resp_hash[:is_truncated]
-          token = resp_hash[:next_continuation_token]
-          resp = client.list_objects_v2({ bucket: bucket_name, max_keys: 1000, prefix: prefix, continuation_token: token })
-          resp_hash = resp.to_h
-          objects = parse_objects(resp_hash)
-          parse_continuation(resp_hash, objects)
-        end
-      end
```
hectorcorrea: @carolyncole I suspect this code got duplicated during a rebase/merge conflict ¯\_(ツ)_/¯

hectorcorrea:
@carolyncole we are down to only two failing tests in `work_upload_edit_service_spec.rb`. I fixed two of them today, hopefully the last two are easier.

carolyncole:
@hectorcorrea I'll pull and try to solve those this morning

carolyncole (approved):
Looks like it is Ready

PR #965, opened 2023-02-07
Reduce the number of ESLint rules we ignore

### LINKED ISSUE #957: Clean up eslint ignores
Follow-up from: https://github.com/pulibrary/pdc_describe/pull/951#discussion_r1097621714_

I'd suggest this be done in two phases:
- Replace the global ignores in `.eslintrc.yaml` with inline ignores, or if it's easy, just fix the problem. This way new code coming in will be cleaner.
- Handle the inline eslint ignores one by one.

mccalluc:
- Fix #957
- Explicitly list a couple globals we need for now, but add a TODO with next steps.
- Leave disabled `no-alert`: This seems fine for an application like this that doesn't need a stunning UI, but we could discuss this more.
- Restrict `no-console`: In my experience, adding console.log is a fine debugging tool, but it's useful to have linting that reminds you to clean it up before merging.
- We have one error with `no-restricted-syntax`, but I haven't reproduced the CI behavior locally. Leave it as a TODO.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -14,7 +14,6 @@ rules:
   no-alert: off
   no-console: off
   no-param-reassign: off
-  no-plusplus: off
   no-restricted-globals: off
   no-restricted-syntax: off
   no-undef: off
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -10,7 +10,6 @@ ignorePatterns:
   'app/javascript/entrypoints/vendor/**'
 rules:
   block-scoped-var: off
-  camelcase: off
   import/prefer-default-export: off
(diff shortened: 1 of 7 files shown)


mccalluc:
Two CI errors that I couldn't reproduce locally. Will move to draft, rerun CI, and pick up tomorrow.

mccalluc:
Errors come up again. We're now capturing screenshots:
Screen Shot 2023-02-08 at 9 18 36 AM

My guess is that there's a JS error, and so it errors out before the JS backed tables are rendered... and there's a caching issue locally that prevents me from reproducing this? I'll try reverting the parts of the last three commits.

bess (approved):

PR #974, opened 2023-02-09
Reduce use of `html_safe` and `raw`

### LINKED ISSUE #908: Audit usage of `html_safe` and `raw`
Want to confirm that the problem in #906 is not more general.
- Review usage of `html_safe` and `raw` and make sure each is safe.
- Consider moving more of this into ERB, where it's harder to get it wrong.

mccalluc:
- fix #908

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -142,12 +142,6 @@ class User < ApplicationRecord
     end
   end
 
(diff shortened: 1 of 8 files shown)


carolyncole:
Closing until we need this, since it is failing CI.

PR #976, opened 2023-02-09
Destructure some longer JS parameter lists

### LINKED ISSUE #956: Use destructuring instead of long parameter lists in JS
```diff
-    copyToClipboard(value, iconEl, labelEl, normalClass, copiedClass) {
+    copyToClipboard({value, iconEl, labelEl, normalClass, copiedClass}) {
```
For longer lists of parameters, we should destructure instead. JS doesn't complain about having the wrong number of parameters, so long parameter lists are even more bug-prone than in ruby.

_Originally posted by @mccalluc in https://github.com/pulibrary/pdc_describe/pull/951#discussion_r1097451679_

mccalluc:
- Fix #956


carolyncole:
Closing unless we need later, since it is failing CI

Issue #986, opened 2023-02-16
Display and sort enhancements to pre-curated file list in Show page


hectorcorrea:
After https://github.com/pulibrary/pdc_describe/pull/983 has been merge we should update the display in data tables so that

* we display the time in the Eastern time zone
* we generate a value that can be used for sorting in the table
* we display size formatted (12,345 instead of 12345)
* we generate a size value that can be used for sorting in the table

PR #996, opened 2023-02-20
Pre-curated file list display improvements

hectorcorrea:
Display filename without DOI/ID prefix. Display file size with commas (but sorts based on integer value).

Moves forward #986

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -82,15 +82,7 @@ class WorksController < ApplicationController
 
   def file_list
     @work = Work.find(params[:id])
-    # @work.pre_curation_uploads_fast is an array of S3File elements (which contain a
-    # reference to S3QueryService). Because of this reference an innocent call to to_json
-    # also serializes the S3QueryService object and this takes forever on large arrays.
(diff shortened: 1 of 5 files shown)


app/controllers/works_controller.rb
```diff
@@ -82,15 +82,7 @@ def show
...
-    # @work.pre_curation_uploads_fast is an array of S3File elements (which contain a
-    # reference to S3QueryService). Because of this reference an innocent call to to_json
-    # also serializes the S3QueryService object and this takes forever on large arrays.
-    # To prevent this behavior we create a simple hash with the properties that we need
-    # so that the conversion to JSON is plain and fast.
-    files = @work.pre_curation_uploads_fast.map do |file|
-      { filename: file.filename, last_modified: file.last_modified, size: file.size }
-    end
-    render json: files
+    render json: @work.pre_curation_uploads_fast
```
hectorcorrea, line 85: My previous hack is not needed anymore since @carolyncole removed the call to the S3 service in S3file 🎉

carolyncole (approved):

Issue #994, opened 2023-02-20
Publish is utilizing precuration_uploads instead of pre_curation_uploads_fast


carolyncole:
Now that we are no longer storing all precuration files in precuration_uploads, the publish job can no longer utilize it.

PR #998, opened 2023-02-21
Removing attach_works which would slow down edits

carolyncole:
This forced publish to utilize S3 instead of local files, which should be faster. 
Many tests needed to change to not attach files to precuration_uploads and instead mock S3

refs #994


bess (approved):
🚀

PR #1008, opened 2023-02-22
Fixes problem accessing User Edit for guest account

### LINKED ISSUE #997: User edit for guest accounts
PR https://github.com/pulibrary/pdc_describe/pull/995 fixed the issue wog going to the User dashboard for guest account but we still have issues going to the User Edit page when logged in with a guest account

hectorcorrea:
Closes #997

### DIFF
--- a/spec/routing/users_routing_spec.rb
+++ b/spec/routing/users_routing_spec.rb
@@ -11,6 +11,10 @@ RSpec.describe UsersController, type: :routing do
       expect(get: "/users/1/edit").to route_to("users#edit", id: "1")
     end
 
+    it "routes to #edit with guest account" do
+      expect(get: "/users/guestacc@gmail.com/edit").to route_to("users#edit", id: "guestacc@gmail", format: "com")
+    end
+
     it "routes to #update via PUT" do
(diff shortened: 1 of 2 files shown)


spec/routing/users_routing_spec.rb
```diff
@@ -11,6 +11,10 @@
       expect(get: "/users/1/edit").to route_to("users#edit", id: "1")
     end
 
+    it "routes to #edit with guest account" do
```
carolyncole, line 14: @hectorcorrea do you need to put the `.com` back onto the id in the edit?

hectorcorrea: The controller did that already, so once the route was handled correctly (i.e. call `edit()` in the controller) the code takes care of putting the id back together.

But the issue in this case was that Rails was failing to even *recognize the route* if it came with the `.com` as part of the `:id`. 

What I still don't know is why Rails did recognize this kind of route for the Show page? We had issues in the Show page but those were in the controller (i.e. Rails recognized the route and called the proper controller method).

carolyncole (approved):

Issue #925, opened 2023-02-01
Re-load migration specs in staging


bess:
## Acceptance Criteria
- [ ] Re-load the migration specs in pdc-describe-staging
- [x] Document and use the new loading process that ensures the DOIs will be minted in staging
- [ ] Tell PRDS that this is done and ask for feedback

PR #1012, opened 2023-02-27
(Mostly) Automate data migration fixture import process

bess:
Also, improve documentation of the process.

Work toward:
* #925


docs/test_data_in_production.md
```diff
@@ -19,57 +19,22 @@ Work with RDOS to ensure we're identifying representative samples. We want to co
...
\ No newline at end of file
+  1. Regenerate the json representations of the migration data by running:
+     ```
+     DATA_MIGRATION=true bundle exec rspec spec/system/data_migration
+     ```
+
+  2. Copy the newly created .json files to the server where you want to load them: 
+    
+    `scp tmp/data_migration/*.json deploy@p
+dc_describe_staging1.princeton.edu:/tmp`
```
mccalluc, line 30:
```suggestion
    ```
    scp tmp/data_migration/*.json deploy@pdc_describe_staging1.princeton.edu:/tmp
    ```
```

docs/test_data_in_production.md
```diff
@@ -19,57 +19,22 @@ Work with RDOS to ensure we're identifying representative samples. We want to co
...
+
+1. Then, ssh to the server where you want to create this sample work as the `deploy` user. 
+2. Run the import rake task, specifying the location of the .json files and the netid of the user they should import as:
```
mccalluc, line 33:
```suggestion
3. Then, ssh to the server where you want to create this sample work as the `deploy` user. 
4. Run the import rake task, specifying the location of the .json files and the netid of the user they should import as:
```

docs/test_data_in_production.md
```diff
@@ -19,57 +19,22 @@ Work with RDOS to ensure we're identifying representative samples. We want to co
...
+  ```
+  bundle exec rake works:import_works\[/path/to/json/files,bs3097]
```
mccalluc, line 35:
```suggestion
  bundle exec rake works:import_works\[/tmp,bs3097]
```

In the previous step, I think you identify a destination, so I think you could use that same destination here... Except there may be other files you don't want in `/tmp` ... maybe a date-stamped subdirectory just for this transfer? Or maybe the script will skip files that aren't json?

mccalluc (commented):
Markdown formatting issue: I think suggestion will fix it.

I may be missing something but could there already be other files in `tmp` before you `scp`?

docs/test_data_in_production.md
Twade968: The JSON files are not going to /tmp. They are going to a tmp/data_migration directory within the application. When they are copied to the server, they will go wherever the user decides to put them.

Twade968 (approved):

PR #1014, opened 2023-02-27
Creating a folder in Amazon so the curators do not need to do it by hand

### LINKED ISSUE #1001: Create the aws directory for external uploads
At the moment it is very hard to create the globus directory correctly if a user wants to put files in from an external source.  What we should do is create the directory, even in nothing is uploaded.

Background:
I created the directory by copying the doi, but I added a space to the name by mistake.  It took me some time to figure out my mistake, and then correct it.

carolyncole:
fixes #1001

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -138,6 +138,10 @@ class WorksController < ApplicationController
     @wizard_mode = true
     @work.files_location = params["attachment_type"]
     @work.save!
+
+    # create a directory for the work if the curator will need to move files by hand
+    @work.s3_query_service.create_directory if @work.files_location != "file_upload"
+
     next_url = case @work.files_location
                when "file_upload"
                  work_file_upload_url(@work)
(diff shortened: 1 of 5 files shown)


app/controllers/works_controller.rb
```diff
@@ -138,6 +138,10 @@ def attachment_selected
     @wizard_mode = true
     @work.files_location = params["attachment_type"]
     @work.save!
+
+    # create a directory for the work if the curator will need to move files by hand
+    @work.s3_query_service.create_directory if @work.files_location != "file_upload"
```
mccalluc, line 143: I'd suggest factoring out this string and referencing it with a variable where it's needed: `work.rb`, `attachment_select.erb`, and here. It would be easy to introduce a typo, and hard to debug.

mccalluc (commented):
All the test failures seem to be the same:
```
undefined local variable or method `data'
```

Besides that, I'd suggest pulling out one string that's used in multiple locations, but it's not essential.

app/controllers/works_controller.rb
carolyncole: Maybe a ticket, or just leave it.  Not really a part of this change...

PR #1018, opened 2023-02-27
Adds pagination, sorting, and download links to the file list for curated works

### LINKED ISSUE #988: Add pagination for post-curated files in Work Show page
We already implemented pagination for pre-curated files (see https://github.com/pulibrary/pdc_describe/issues/931), but we should also do it for post-curated files. 

Post-curated files only live in S3 (and never in ActiveStorage) so speed might not be as big of a problem when the number of files is large as it was for pre-curated files, but the display still looks awful when we dump 2000+ files on the page without pagination.

hectorcorrea:
Updated to render post-curation files via DataTables and AJAX. Merged the two partials for file rendering into one now that files always live in S3 and the only difference is the bucket that they come from.

Screenshot 2023-02-27 at 6 09 19 PM

Closes #988


app/views/works/show.html.erb
```diff
@@ -248,21 +248,11 @@
...
-          <h2 class="h2"><%= t('works.uploads.pre_curation.heading') %></h2>
-          <%= render(partial: 'works/uploads/pre_curation_s3_resources') %>
-        </section>
-      <% end %>
-    <% end %>
+    <section class="uploads">
+      <h2 class="h2"><%= @work.approved? ? "Curated Files" : "Uncurated Files" %></h2>
+      <%= render(partial: 'works/s3_resources') %>
+    </section>
+
```
hectorcorrea, line 255: `works/s3_resources` partial is the merged version of `_post_curation_s3_resources` and `_pre_curation_s3_resources`

app/services/s3_query_service.rb
```diff
@@ -119,9 +119,6 @@ def find_s3_file(filename:)
   # Retrieve the S3 resources uploaded to the S3 Bucket
   # @return [Array<S3File>]
   def client_s3_files(reload: false, bucket_name: self.bucket_name)
-    # Allow migration objects to load locally with an AWS key
-    return [] if Rails.env.development?
-
```
hectorcorrea, line 124: We probably missed this during the refactoring...but that throw me off for a while! 😱

app/controllers/works_controller.rb
```diff
@@ -82,7 +82,7 @@ def show
 
   def file_list
     @work = Work.find(params[:id])
-    render json: @work.pre_curation_uploads_fast
+    render json: @work.uploads
```
hectorcorrea, line 85: `work.uploads()` in the model handles the logic of returning pre/post curated files.

mccalluc (commented):
One missing change:
```
--- a/app/views/works/_form.html.erb
+++ b/app/views/works/_form.html.erb
@@ -43,7 +43,7 @@
           <% if @work.approved? %>
             <section class="uploads">
               <h2 class="h2"><%= t('works.uploads.post_curation.heading') %></h2>
-              <%= render(partial: 'works/uploads/post_curation_s3_resources') %>
+              <%= render(partial: 'works/s3_resources') %>
             </section>
           <% else %>
```

hectorcorrea:
@mccalluc good catch!

app/views/works/_s3_resources.html.erb
```diff
@@ -14,29 +16,32 @@
...
       </div>
     </div>
   </div>
 </section>
 
 <script type="text/javascript">
   $(function() {
+    // This is needed so that we can swap the URL during testing
+    // (for more info see external_identifier_specs.rb)
+    var fileListUrl = '<%= Rails.env.test? ? WORKS_FILE_LIST_URL.gsub(/place-holder/, @work.id.to_s) : work_file_list_path(@work) %>';
```
carolyncole, line 36: For the other URLS, we set it in both cases (test or not).  Then there is no need for the logic here...  See the ORCID for and [example](https://github.com/pulibrary/pdc_describe/search?q=ORCID_URL)

app/views/works/_s3_resources.html.erb
```diff
@@ -14,29 +16,32 @@
           <tbody>
             <% if @work.uploads.empty? %>
               <tr class="files">
-                <td><span><%= t('works.form.curation_uploads.empty') %></span></td>
+                <td><span>No files in S3</span></td>
```
carolyncole, line 19: Curious why you remove the translation here?  Not a blocker at all...

carolyncole (changes requested):
@hectorcorrea One question on the use of the external_identifier_specs server and another curiosity from me...

I'd like to keep our usage consistent...

app/views/works/_s3_resources.html.erb
hectorcorrea: Good catch. Restored.

app/views/works/_s3_resources.html.erb
hectorcorrea: Adding a reference to `work_file_list_path(id: "place-holder")` in `./config/environments/development.rb`  results in error "undefined method `work_file_list_path' for main:Object (NoMethodError)" -- I don't think all the Rails configuration is available at that stage.

carolyncole (approved):
Great to have data tables everywhere.

hectorcorrea:
Hats off to @carolyncole for figuring it out a much simpler testing approach!

PR #1029, opened 2023-03-08
Export PPPL data for demo purposes

bess:


app/services/s3_query_service.rb
```diff
@@ -119,6 +119,7 @@ def find_s3_file(filename:)
   # Retrieve the S3 resources uploaded to the S3 Bucket
   # @return [Array<S3File>]
   def client_s3_files(reload: false, bucket_name: self.bucket_name)
+    return [] if Rails.env.development?
```
jrgriffiniii, line 122: 👍

spec/system/data_migration/dynamic_tension_form_submission_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+      # Check that RORs were persisted as funder names
+      funders = dynamic_tension_work.resource.funders.map(&:funder_name).uniq
+      expect(funders).to contain_exactly("United States Department of Energy")
+
+      # # Ensure the datacite record produced validates against our local copy of the datacite schema.
+      # # This will allow us to evolve our local datacite standards and test our records against them.
+      datacite = PDCSerialization::Datacite.new_from_work(dynamic_tension_work)
+      expect(datacite.valid?).to eq true
+      expect(datacite.to_xml).to be_equivalent_to(File.read("spec/system/data_migration/dynamic_tension.xml"))
```
jrgriffiniii, line 85: 👍

hectorcorrea (changes requested):
See note regarding `return [] if Rails.env.development?`

app/services/s3_query_service.rb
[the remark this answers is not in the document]
bess: @hectorcorrea and I came up with a better solution, which is now documented in the README file.

hectorcorrea (approved):

Issue #1041, opened 2023-03-31
Adding files view Create Dataset does not save initial files


hectorcorrea:
When creating a work via the "Create Dataset" option that bypasses the wizard we allow the user to add fields directly on the form, but these files are NOT saved!

(Files can be added after the fact and they are saved, the issue is when adding them right on the first page after hitting "Create Dataset")

![create_work_0](https://user-images.githubusercontent.com/568286/229217471-765e2990-7d60-4c3f-8aed-84162ed9357c.png)

![create_work_1](https://user-images.githubusercontent.com/568286/229217499-731b5204-308d-416d-b9db-c8c5b2882ac6.png)

![create_work_2](https://user-images.githubusercontent.com/568286/229217525-02c93cd9-68d2-4076-be9f-338b6768427a.png)

PR #1042, opened 2023-03-31
Updated interface in Work Edit page to allow users to add and delete files

### LINKED ISSUE #1010: Describe Form: File upload after work is submitted is deleting previous files
## Description

Users do not currently have a way to upload additional files without deleting what they have already uploaded in the process. The user experience of uploading or deleting files should feel more intuitive and reflect what users have likely encountered for other software applications. 

After discussing our options with RDSS, the "replace all uploads" feature should instead allow users to add new files. The "replace upload" for individual files feature will allow deletion of the file. However, we had discussed implementing a fail-safe for file deletions, so this will be written to a separate ticket. 

#### Image shows upload feature as it is currently
Screen Shot 2023-02-24 at 9 46 44 AM

## Acceptance Criteria
- [ ]  The "replace all uploads" feature will instead allow users to add new files.
- [ ] Another ticket will be written for file deletion, because that will include a fail-safe feature. However, note that the "replace upload" for individual files feature will allow deletion of the file.

### LINKED ISSUE #1021: Describe Form: Delete files + Failsafe and new Upload
## Description

The user experience of uploading or deleting files should feel more intuitive and reflect what users have likely encountered for other software applications. In ticket #1010, we explained that we want both file addition and file deletion, and described where these features should be located on the files feature. This ticket covers the details for how we expect the delete feature should work. Please see the images below.

## Image shows files that were uploaded to a submission
Screen Shot 2023-03-01 at 2 00 57 PM

#### A thick black "X" appears to show which files have been selected for deletion. These files will not delete until "save work" has been selected. This provides a fail-safe against accidental deletion. We use a thick black "X"  because it will be inclusive of visual differences, but still very noticeable. 

Note: We may need additional help text to explain that save work is selected to confirm deletion.
Screen Shot 2023-03-01 at 2 01 09 PM

#### The submitter makes a correction by unchecking the README file they do not want to delete
Screen Shot 2023-03-01 at 2 01 23 PM

## Image shows the Paints.xls file was deleted

Screen Shot 2023-03-01 at 2 01 36 PM

#### Please assume the submitter had used the "add new files" button to add this file. This button will work in the same way as uploading files during the submission process.

Screen Shot 2023-03-01 at 2 01 50 PM

## Acceptance Criteria
- [ ] Ticket #1010 is completed
- [ ] The "replace upload" feature is removed from the file listing
- [ ] A fail-safe is added for files that are selected for deletion
- [ ] The fail-safe appears as a visible thick, black "X" (or contact Hannah if you have another idea)
- [ ] A selected file may be de-selected
- [ ] Selected files will delete after selecting "save work"
- [ ] **Suggested:** We add help text to say that "Selected files will be deleted after saving work." (or a similar statement)
- [ ] New files may be added using the "add new files" button, which was changed in ticket #1010 
- [ ] Adding new files will occur in the same way as adding them during the submission process.

hectorcorrea:
There are three major changes on this PR:
* Updated the work edit page to use DataTables so that files are paginated
* Updated the UX to allow the user to *add* files to an existing work
* Added the ability to *delete* an existing file from a work

Closes #1021
Closes #1010 

Notice that I left a few TODOs in the code that point to an (existing) issue that needs to be addressed separately (https://github.com/pulibrary/pdc_describe/issues/1041)

work-edit


spec/system/work_spec.rb
```diff
@@ -286,27 +286,5 @@ def value_for(label)
...
-      # This is necessary to interact with the rendered JavaScript <table> rows
-      page.scroll_to(target)
-      sleep 1
-
-      source.drag_to(target)
-      sleep 1
-
-      dragged = page.find('.uploads-row[data-upload-key="0"]')
-      expect(dragged[:id]).to eq(source[:id])
-    end
```
hectorcorrea, line 310: I have no idea what this test was attempting to test...but we do not allow the user to re-order the display of files so I removed it.

carolyncole (changes requested):
@hectorcorrea I deployed this to staging and got errors when trying to delete files.
Screenshot 2023-04-03 at 8 36 36 AM

hectorcorrea:
@carolyncole thank you so much tip, the spaces where indeed the problem. 

I've updated the code to use a safe value for the id of the HTML elements (based on the full path of the name) that prevents issue with any non-alphanumeric characters in the name.

carolyncole (approved):
@hectorcorrea now works for me.  Thanks for updating it!

Issue #991, opened 2023-02-17
Track Changes to Files in Globus (AWS) In the Change History


carolyncole:
Track the number of files that are Added Deleted and Replaced in Globus In the Change History.  We should compare the items in the system Name and checksum with the items in AWS each time the item is shown.
 
![Screenshot 2023-04-05 at 2 03 23 PM](https://user-images.githubusercontent.com/1599081/230165513-5f4882e4-4e4a-4243-b804-ea322dfc9b27.png)

We want to know what changes are happening in Globus in our Change History on the work show page.

The Number of files Replaced Added and Removed should be tracked.  See #990 for definition of replaced.
Also should track the file names of the Added, Removed, or Replaces files (but not display them)

### UI Changes for pre MVP

What we have now is a list of the files changed

![Screenshot 2023-04-05 at 2 17 42 PM](https://user-images.githubusercontent.com/1599081/230169039-883806e3-61c1-46e1-891b-cf3f6e2ec93c.png)
In the new way 
```
File Added: 1
```
And if any are removed

```
File Removed: 1
```
And if any are replaced (the checksum is diffed to see in this has been changed)
```
File Replaced: 1
```

Completion criteria
- [x] Track the changes from Globus (really AWS in our system).  This might need to be a new database field or part of the resource. Snapshot the list of files in the database.
- [x] If there are changes we create another snapshot in the database
- [x] We create a diff provenance record when the file list changes
- [x] The diff numbers are displayed on the UI in the provenance log
- [x] There are automated tests

carolyncole:
The current File Changes should be matched up to show just the numbers for now.

![Screenshot 2023-02-17 at 1 01 49 PM](https://user-images.githubusercontent.com/1599081/219740987-ef631b4f-e36c-4c07-8efe-f24cb30975d4.png)

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-Workcycles-61a4f1a12a399b001730f65a/issues/pulibrary/pdc_describe/991?planning-poker) with Zenhub @bess @jrgriffiniii @kelynch @mccalluc

PR #1060, opened 2023-04-14
Integrating Support for Upload Snapshots

jrgriffiniii:
Advances #991


jrgriffiniii:
I am not certain why, but the status check on this is not updating with https://app.circleci.com/pipelines/github/pulibrary/pdc_describe/3000/workflows/579e8eac-b40c-47c6-970b-9ba64545022e

jrgriffiniii:
https://app.circleci.com/pipelines/github/pulibrary/pdc_describe/3002/workflows/c480140b-4b8e-4739-893f-4b2389b2f537 seems to be finally running, I am not quite certain as to why this was delayed.

app/services/s3_query_service.rb
```diff
@@ -119,6 +119,8 @@ def find_s3_file(filename:)
   # Retrieve the S3 resources uploaded to the S3 Bucket
   # @return [Array<S3File>]
   def client_s3_files(reload: false, bucket_name: self.bucket_name, prefix: self.prefix)
+    return [] if Rails.env.development?
```
carolyncole, line 122: @jrgriffiniii this is causing me issue in development.  I just have a script locally that exports the staging environment variables so I can connect to AWS in development.  This causes things not to work for me on this branch.

app/models/work.rb
```diff
@@ -355,8 +355,24 @@ def uploads
     pre_curation_uploads_fast
   end
 
+  def pre_curation_s3_files
```
carolyncole, line 358: @james it looks like you are trying to utilize local storage for development.  I'm not sure this will still work.  It would be better just to get the AWS keys from the staging server.

app/controllers/works_controller.rb
```diff
@@ -173,6 +174,7 @@ def file_uploaded
       upload_service = WorkUploadsEditService.new(@work, current_user)
       @work = upload_service.update_precurated_file_list(files, [])
       @work.save!
+      @work.reload_snapshots
```
carolyncole, line 177: ❤️

app/controllers/works_controller.rb
```diff
@@ -316,7 +318,7 @@ def patch_params
     def pre_curation_uploads_param
       return if patch_params.nil?
 
-      patch_params[:pre_curation_uploads_added]
+      patch_params[:pre_curation_uploads]
```
carolyncole, line 321: This seems like an odd change.  Is it a real change?

app/models/s3_file.rb
```diff
@@ -31,6 +31,15 @@ def globus_url
     File.join(Rails.configuration.globus["post_curation_base_url"], encoded_filename)
   end
 
+  def create_blob(key: nil)
```
carolyncole, line 34: Why are we going back to creating blobs here?  This seems unneeded

app/models/upload_snapshot.rb
```diff
@@ -2,9 +2,29 @@
 class UploadSnapshot < ApplicationRecord
   belongs_to :work
   attr_writer :upload
+  has_one_attached :bitstream
```
carolyncole, line 5: Why is there a bitstream here?  I thought we were just going to utilize the checksum.

config/environments/development.rb
```diff
@@ -32,7 +32,7 @@
   end
 
   # Store uploaded files on the local file system (see config/storage.yml for options).
-  config.active_storage.service = :amazon
+  config.active_storage.service = :local
```
carolyncole, line 35: I believe changing back to local for development is going to cause us pain now that we want to rely on the files existing in S3.  We moved away from ActiveStorage bitstreams because they were super slow.  Adding them back in seems like a mistake.

config/storage.yml
```diff
@@ -10,12 +10,10 @@ amazon: &amazon
   bucket: <%= S3QueryService.pre_curation_config.fetch(:bucket) %>
 
 development: &development
-# AWS is the default as the will not really work without AWS
-#  You can go back to local if needed for some reason, just be aware the uploaded
-#  files will not display in the edit or the show pages
-#  <<: *local
-   <<: *amazon
-
+  # Default to the local file system (this avoids coercion into AWS)
```
carolyncole, line 13: Going back to local here seems like a mistake.

db/migrate/20230303165757_create_upload_snapshots.rb
```diff
@@ -1,7 +1,11 @@
+# frozen_string_literal: true
 class CreateUploadSnapshots < ActiveRecord::Migration[6.1]
   def change
     create_table :upload_snapshots do |t|
-      t.string :uri, null: false
+      t.string :filename
```
carolyncole, line 5: This set of parameters looks correct to me

spec/models/s3_file_spec.rb
```diff
@@ -82,14 +83,53 @@
...
+        checksum: "abc567",
+        work: work
+      )
+    end
+    let(:blob) do
+      s3_file.to_blob
+    end
+    let(:s3_client) { instance_double(Aws::S3::Client) }
+    let(:s3_query_service) { instance_double(S3QueryService) }
+    let(:bitstream) { instance_double(ActiveStorage::Attached::One) }
```
carolyncole, line 100: I think utilizing the bitstream should be avoided at all costs.

spec/requests/works_spec.rb
```diff
@@ -49,4 +50,127 @@
...
+          stub_ark
+          patch work_url(work), params: params
+          work.reload
+        end
+
+        it "renders messages generated for the added files" do
+          get work_url(work)
+
+          expect(response.code).to eq "200"
+          expect(response.body).to include("Files Added: 2")
```
carolyncole, line 116: This is exactly right!

spec/requests/works_spec.rb
```diff
@@ -49,4 +50,127 @@
...
+
+          patch work_url(work), params: params
+          work.reload
+        end
+
+        it "renders messages generated for the replaced files" do
+          get work_url(work)
+
+          expect(response.code).to eq "200"
+          expect(response.body).to include("Files Replaced: 1")
```
carolyncole, line 171: Nice!

carolyncole (changes requested):
Overall there are some great updates in this PR.

I think it is a mistake to go back to utilizing local storage in development as it is not close to what the server will see.  I think we wanted to tend away from ActiveStorage because it is so slow.

@james Could I either take what you have done and update it, or could we pair to remove the ActiveStorage portion and keep the rest?  I'll put a meeting on our calendars.

jrgriffiniii (commented):
@carolyncole thank you very, very much again for the continuous reviews, patience, and additional commits for this 🙏

Issue #1019, opened 2023-02-28
Automatically import data payloads from DataSpace


bess:
For objects that are being migrated from dataspace, we need a way to automatically bring over all files from DataSpace and put them in the right place in the pre-curation bucket. Then, once the work is approved, the files will be moved to post-curation as they would for any other object. 

@hectorcorrea wrote something that can be used for inspiration: https://github.com/hectorcorrea/dsget

## Acceptance criteria
- [ ] The button is not visible to a casual user; we don't want it getting pressed by accident
- [ ] When the button is pushed, it uses the ark to find the right object in DataSpace and brings over its data to attach to the PDC Describe object

carolyncole:
I should be able to retrieve any data in Globus by getting the DOI.  And then searching `prds-dataspace` bucket for the DOI and copying those items into the work pre-curation bucket via AWS.

carolyncole:
smaller dspace with globus sample: https://dataspace.princeton.edu/handle/88435/dsp01cz30pw862

PR #1071, opened 2023-04-18
Moving away from copy_object as that is limited to 5GB files

carolyncole:
Changing all copies to multipart so that we can handle files larger than 5GB during publish and other AWS file movements

Specifically this will be needed for files that are in Dspace that were uploaded to Globus, because some of them are larger than 5GB.  The example  https://dataspace.princeton.edu/handle/88435/dsp01s7526g63n has this issue.

refs #1019


bess (approved):
🎉

PR #1082, opened 2023-04-19
Jammy update

hectorcorrea:
@kayiwa updated to run on Ruby 3.1.0

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 # This file is copied to spec/ when you run 'rails generate rspec:install'
 require "simplecov"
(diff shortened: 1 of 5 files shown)


spec/rails_helper.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 # This file is copied to spec/ when you run 'rails generate rspec:install'
 require "simplecov"
-SimpleCov.start "rails"
```
hectorcorrea, line 4: There is another call to `SimpleCov.start` in line 27. The new version of Ruby does not like that we call it more than once (see https://github.com/simplecov-ruby/simplecov/issues/1003)

Issue #1068, opened 2023-04-17
HTML encoding issue with em dash in Work Edit page


hectorcorrea:
The help text for the description field on the Work edit page is not encoding properly the em dash

![Screenshot 2023-04-17 at 4 11 15 PM](https://user-images.githubusercontent.com/568286/232599425-e17be838-da3b-43b4-a1f6-853efd2020fe.png)

hectorcorrea:
Closed with https://github.com/pulibrary/pdc_describe/pull/1086

PR #1086, opened 2023-04-20
Process uploads when creating a dataset outside the wizard

### LINKED ISSUE #1041: Adding files view Create Dataset does not save initial files

hectorcorrea:
Makes sure files uploaded are processed when creating a new dataset without using the Wizard.

Closes #1041

I also bundled the fix for #1068 since it was a simple typo in a view.

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -44,6 +44,11 @@ class WorksController < ApplicationController
     # We currently do not and therefore there are not saved to the work.
     # See https://github.com/pulibrary/pdc_describe/issues/1041
     @work = Work.new(created_by_user_id: current_user.id, collection_id: params_collection_id, user_entered_doi: params["doi"].present?)
+
+    byebug
+
+    upload_service = WorkUploadsEditService.new(@work, current_user)
+    upload_service.update_precurated_file_list(added_files_param, deleted_files_param)
     @work.resource = FormToResourceService.convert(params, @work)
     if @work.valid?
       @work.draft!(current_user)
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -40,18 +40,12 @@ class WorksController < ApplicationController
   end
 
   def create
-    # TODO: We need to process files submitted by the user in this method.
-    # We currently do not and therefore there are not saved to the work.
-    # See https://github.com/pulibrary/pdc_describe/issues/1041
     @work = Work.new(created_by_user_id: current_user.id, collection_id: params_collection_id, user_entered_doi: params["doi"].present?)
-
-    byebug
-
-    upload_service = WorkUploadsEditService.new(@work, current_user)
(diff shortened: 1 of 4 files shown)


app/views/works/_required_form.html.erb
```diff
@@ -16,8 +16,8 @@
 
 <!-- Description (one for now)-->
 <div class="field">
-  <%= render partial: "required_form_field_details", locals: { name: "description", display_name: "Description", 
-                                                       field_detail: "Please enter a brief summary of this item specifically &mdash; not identical to the abstract of a corresponding paper. The focus should be on the nature of the materials constituting the item (scope, purpose, methods, etc.), as opposed to substantive claims made in related publications.",
+  <%= render partial: "required_form_field_details", locals: { name: "description", display_name: "Description",
```
hectorcorrea, line 19: This is the typo (mdash was not being processed correctly and rendered as `mdash` rather than `-`.

app/controllers/works_controller.rb
```diff
@@ -40,13 +40,12 @@ def new
...
 
   def create
-    # TODO: We need to process files submitted by the user in this method.
-    # We currently do not and therefore there are not saved to the work.
-    # See https://github.com/pulibrary/pdc_describe/issues/1041
     @work = Work.new(created_by_user_id: current_user.id, collection_id: params_collection_id, user_entered_doi: params["doi"].present?)
     @work.resource = FormToResourceService.convert(params, @work)
     if @work.valid?
       @work.draft!(current_user)
+      upload_service = WorkUploadsEditService.new(@work, current_user)
```
jrgriffiniii, line 47: 👍

jrgriffiniii (dismissed):

hectorcorrea:
@jrgriffiniii I've fixed he failing test. Could you take a second look?

spec/system/data_migration/attention_form_submission_spec.rb
```diff
@@ -14,12 +14,17 @@
...
+    "https://example-bucket.s3.amazonaws.com/"
+  end
 
   before do
     stub_datacite(host: "api.datacite.org", body: datacite_register_body(prefix: "10.34770"))
     stub_request(:get, "https://handle.stage.datacite.org/10.34770/9425-b553")
       .to_return(status: 200, body: "", headers: {})
-    stub_s3
+    stub_request(:put, /#{bucket_url}/).to_return(status: 200)
+    stub_s3(data: [file1])
```
hectorcorrea, line 27: We now need to stub this because the controller processes files on create. The original bug that the controller was **not** processing them.

jrgriffiniii (approved):

PR #1089, opened 2023-04-21
Fixes issue with files uploaded when creating a new dataset in the wizard

hectorcorrea:
Files uploaded when creating a new dataset in the wizard were not being processed. @hhkitsune, @neggink, and I noticed this during the last rounds of user testing. 

This PR makes sure these files are processed and uploaded correctly.

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -168,11 +168,12 @@ class WorksController < ApplicationController
 
   def file_uploaded
     @work = Work.find(params[:id])
-    if pre_curation_uploads_param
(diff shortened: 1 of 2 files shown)


app/controllers/works_controller.rb
```diff
@@ -168,11 +168,12 @@ def file_upload
 
   def file_uploaded
     @work = Work.find(params[:id])
-    if pre_curation_uploads_param
-      @work.pre_curation_uploads.attach(pre_curation_uploads_param)
+    files = work_params.dig("patch", "pre_curation_uploads") || []
```
jrgriffiniii, line 171: 👍

jrgriffiniii (approved):

Issue #1084, opened 2023-04-20
Coveralls reporting is incorrect


bess:
According to coveralls, our code coverage just dropped significantly. Coveralls is [reporting](https://coveralls.io/builds/59101877/source?filename=app%2Fmodels%2Fuser.rb) that our `User` model doesn't have any test coverage, but it definitely does: https://github.com/pulibrary/pdc_describe/blob/main/spec/models/user_spec.rb

![Screenshot 2023-04-20 at 10 56 00 AM](https://user-images.githubusercontent.com/65608/233405514-d726f09b-a4d5-4620-8b2a-87a10864fb1f.png)

hectorcorrea:
I wonder if this is related to the SimpleCov change that I made here: https://github.com/pulibrary/pdc_describe/pull/1082/files#diff-053aa3ea46cb7c6a649e0d0fc592ce729b66e1d44be475516fd58857ba695862

PR #1102, opened 2023-04-28
Ensuring that coveralls is replaced with just simplecov

jrgriffiniii:
Advances #1084

### DIFF
--- a/lib/simplecov_helper.rb
+++ b/lib/simplecov_helper.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+require "active_support/inflector"
+require "simplecov"
+
+class SimpleCovHelper
+  DEFAULT_BASE_DIR = "./coverage/"
+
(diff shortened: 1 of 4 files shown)


lib/simplecov_helper.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require "active_support/inflector"
+require "simplecov"
+
+class SimpleCovHelper
+  DEFAULT_BASE_DIR = "./coverage/"
+
+  def self.report_coverage(base_dir: DEFAULT_BASE_DIR, minimum_coverage: 100)
```
bess, line 9: Is this going to enforce coverage at 100%, the way DLS does it? Are we ready to commit to that as a team?

bess (changes requested):
Looks good to me except that CI is failing, possibly because our coverage isn't at 100%? Is that something we are ready to commit to? I'm going to add this to the team agenda for discussion.

PR #1105, opened 2023-05-01
Ensuring that only a single WorkActivity with multiple events will be created for file additions, replacements, and deletions

jrgriffiniii:
This ensures that `WorkActivity` models for `UploadSnapshots` consistently contain multiple actions tracking file modifications.


carolyncole (approved):
Thanks @jrgriffiniii

Issue #1013, opened 2023-02-27
Fix formatting for related objects field.


Twade968:
When a work has more than one related object, they run together on the show screen.

![Screen Shot 2023-02-27 at 9.41.35 AM.png](https://images.zenhubusercontent.com/62e44626d7978151c71e4583/1d7cfc7c-fd93-41a7-8e22-6e73ef9b1f17)

Completion Criteria:
- [ ] Each Citation is on a new line
- [ ] Check into if there is a real way a Data Dictionary item should have multiple definitions
- [ ] Automated Test to verify the structure exists

PR #1107, opened 2023-05-01
Correct spelling of options in individual and organizational contributor roles

### LINKED ISSUE #1069: Roles in Work Edit display as CamelCase to the user
Roles should not display in CamelCase to the user in the drop down list. We need to save them as such for DataCite but not for display purposes:

Screenshot 2023-04-17 at 4 15 55 PM

Make sure the display in the Work Show is also accurate.

@Twade968 also reported a similar issue with the related objects field (https://github.com/pulibrary/pdc_describe/issues/1013)

hectorcorrea:
Makes sure labels are displayed in human friendly spelling rather than CamelCase

Closes #1069


app/views/layouts/application.html.erb
```diff
@@ -56,13 +56,16 @@
       orcid_url: '<%= raw(ORCID_URL) %>',
       ror_url: '<%= raw(ROR_URL) %>',
       datacite: <%= raw(
-        ['ContributorType', 'RelationType', 'RelatedIdentifierType'].map { |name|
+        ['RelationType', 'RelatedIdentifierType'].map { |name|
```
hectorcorrea, line 59: We should also handle RelationType and RelatedIdentifierType in a similar way to ContributorType, but we'll do that via separate issue and PR (see https://github.com/pulibrary/pdc_describe/issues/1013)

app/views/layouts/application.html.erb
```diff
@@ -56,13 +56,16 @@
...
           [
             name,
             Datacite::Mapping.const_get(name).to_a.map(&:value)
           ]
         }.to_h.to_json
       ) %>,
+      dataciteContributorType: <%= raw(
+        Datacite::Mapping::ContributorType.to_a.map { |x| {key: x.key, value: x.value.titleize} }.to_json
+      ) %>
```
hectorcorrea, line 68: This is the big change in this PR. We now gather both the key and the human value (e.g. "CONTACT_PERSON" and "Contact Person") rather than always using "ContactPerson"

jrgriffiniii (approved):

PR #1112, opened 2023-05-02
Checking to make sure the ARK is unique

### LINKED ISSUE #1062: ARKs and DOIs must be unique
Screenshot 2023-04-13 at 11 16 36 AM

carolyncole:
Updating a few tests that created works with the same ARK
Also fixed a bug with unique to not count itself if it has been saved as a conflict with being unique

fixes #1062

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -131,7 +131,7 @@ class Work < ApplicationRecord
 
   validate do |work|
     if none?
-      work.validate_doi
+      work.validate_ids
     elsif draft?
       work.valid_to_draft
     else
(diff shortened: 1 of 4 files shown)


app/models/work.rb
```diff
@@ -171,12 +171,18 @@ def validate_doi
 
   def unique_doi
     other_record = Work.find_by_doi(doi)
+    return true if other_record == self
```
jrgriffiniii, line 174: 👍

app/models/work.rb
```diff
@@ -626,13 +632,24 @@ def data_cite_connection
...
       changed_value = metadata["ark"] != ark
       if first_save || changed_value
         errors.add(:base, "Invalid ARK provided for the Work: #{ark}") unless Ark.valid?(ark)
       end
     end
 
+    def unique_ark
+      return true if ark.blank?
+      other_record = Work.find_by_ark(ark)
+      return true if other_record == self
```
jrgriffiniii, line 646: 👍

Issue #1110, opened 2023-05-01
Describe Staging: Error Message


hhkitsune:
## Description

Received an error message in Describe. These were my actions leading to the message:
- successful log in
- navigated to the dash board
- opened a pending record and selected edit
- did not edit the content, but went back to the dashboard to look for a different record
- tried to open a "published" record and received the error message
- successfully went back to the dashboard
- tried to open other records, but received the error message each time

## Image shows the error message
<img width="1219" alt="Screen Shot 2023-05-01 at 5 15 38 PM" src="https://user-images.githubusercontent.com/50670293/235533604-81dd7d24-f20c-42e4-9061-163a8e51941b.png">

## Acceptance Criteria
- [ ] Diagnose the reason behind the error message
- [ ] Message will not appear and selections will work as expected

hhkitsune:
Update: Others in PRDS are experiencing this same error message.

hectorcorrea:
The error happens when editing any published record. The logs show a problem with a missing column in the `upload_snapshots` table:

```
F, [2023-05-03T14:00:54.961517 #710907] FATAL -- : [5f01e5db-6165-422e-b331-18ef6fdd8077]   
[5f01e5db-6165-422e-b331-18ef6fdd8077] ActiveRecord::StatementInvalid (PG::UndefinedColumn: ERROR:  column upload_snapshots.url does not exist
LINE 1: ...hots" WHERE "upload_snapshots"."work_id" = $1 AND "upload_sn...
                                                             ^
HINT:  Perhaps you meant to reference the column "upload_snapshots.uri".
):
[5f01e5db-6165-422e-b331-18ef6fdd8077]   
[5f01e5db-6165-422e-b331-18ef6fdd8077] app/models/work.rb:480:in `block in new_s3_files'
[5f01e5db-6165-422e-b331-18ef6fdd8077] app/models/work.rb:479:in `reject'
[5f01e5db-6165-422e-b331-18ef6fdd8077] app/models/work.rb:479:in `new_s3_files'
[5f01e5db-6165-422e-b331-18ef6fdd8077] app/models/work.rb:485:in `new_snapshots'
[5f01e5db-6165-422e-b331-18ef6fdd8077] app/models/work.rb:491:in `current_snapshots'
[5f01e5db-6165-422e-b331-18ef6fdd8077] app/models/work.rb:495:in `invalid_snapshots'
[...]

PR #1113, opened 2023-05-03
Add migration to rename UploadSnapshot uri to url

carolyncole:
refs #1110


hectorcorrea:
@carolyncole I am wondering if we should create a migration to drop the table and recreate it since the table in staging is so out of whack with the one in schema.rb (rather than patching field by field)

hectorcorrea (approved):

Issue #1129, opened 2023-05-08
Describe Staging: Error message when trying to approve


hhkitsune:
## Description

Describe staging provided an error message when we tried to approve a test submission.

## Image shows error message
![PDCfail](https://user-images.githubusercontent.com/50670293/236854754-8923fafb-2c2d-4903-9255-ff643734ab6e.png)

## Acceptance criteria
- [ ] Diagnose and fix the error
- [ ] Display a user friendly message rather than the default Rails error message
- [ ] We should be able to approve submissions in PDC staging -- to move from staging pre-curation to staging post curation -- in order to test the system.

hectorcorrea:
It looks like the serialize to DataCite does not handle unknown types correctly. The log shows the following error:

```
I, [2023-05-08T14:50:10.435691 #1147]  INFO -- : [a6f7301d-aefa-4be9-b974-b95adee2d5f0] Processing by WorksController#approve as HTML
I, [2023-05-08T14:50:10.435843 #1147]  INFO -- : [a6f7301d-aefa-4be9-b974-b95adee2d5f0]   Parameters: {"authenticity_token"=>"[FILTERED]", "id"=>"69"}
I, [2023-05-08T14:50:10.475358 #1147]  INFO -- : [a6f7301d-aefa-4be9-b974-b95adee2d5f0] Loading S3 objects. Bucket: pdc-describe-staging-precuration. Prefix: 10.80021/0hj6-9f47/69/. Elapsed: 0.024303241 seconds
I, [2023-05-08T14:50:10.480323 #1147]  INFO -- : [a6f7301d-aefa-4be9-b974-b95adee2d5f0] Completed 500 Internal Server Error in 44ms (ActiveRecord: 6.7ms | Allocations: 6772)
F, [2023-05-08T14:50:10.482097 #1147] FATAL -- : [a6f7301d-aefa-4be9-b974-b95adee2d5f0]   
[a6f7301d-aefa-4be9-b974-b95adee2d5f0] ArgumentError (Type cannot be nil):
[a6f7301d-aefa-4be9-b974-b95adee2d5f0]   
[a6f7301d-aefa-4be9-b974-b95adee2d5f0] app/models/pdc_serialization/datacite.rb:133:in `new'
[a6f7301d-aefa-4be9-b974-b95adee2d5f0] app/models/pdc_serialization/datacite.rb:133:in `block in individual_contributors_from_work_resource'
[...]

PR #1132, opened 2023-05-08
Rake task for DataCite validation 

hectorcorrea:
Adds rake task to help troubleshoot records with bad DataCite attributes.

Used to troubleshoot #1129

### DIFF
--- a/lib/tasks/works.rake
+++ b/lib/tasks/works.rake
@@ -82,4 +82,16 @@ namespace :works do
       puts "\t#{file_name}"
     end
   end
+
+  # Use this to figure out what works don't have the proper DataCite attributes
+  # (useful when troubleshooting bad data already in the system)
+  desc "Shows what works are valid for DataCite submission."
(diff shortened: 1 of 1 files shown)


lib/tasks/works.rake
```diff
@@ -82,4 +82,16 @@ namespace :works do
       puts "\t#{file_name}"
     end
   end
+
+  # Use this to figure out what works don't have the proper DataCite attributes
+  # (useful when troubleshooting bad data already in the system)
+  desc "Shows what works are valid for DataCite submission."
+  task :validate_datacite_attributes, [:verbose] => :environment do |_, args|
+    Work.all.each do |work|
+      _attributes = work.send(:doi_attribute_xml)
```
carolyncole, line 91: I believe the code below would trigger the same error.  I'm unsure why the doi_attribute_xml method exists.
```suggestion
      _attributes = work.resource.to_xml
```

carolyncole (changes requested):

carolyncole (approved):

PR #1136, opened 2023-05-09
Updated relative URLs

### LINKED ISSUE #1134: Use relative links for navigation menu
"About" and "How to Submit" URLs should be relative, not absolute.

kelynch:
Closes #1134

### DIFF
--- a/app/views/shared/_header.html.erb
+++ b/app/views/shared/_header.html.erb
@@ -3,8 +3,8 @@
     <%= link_to "PDC Describe", root_path, class: 'navbar-logo' %>
     <div class="menu">
       <nav id="menu" class="col-sm" role="menu">
-      <%= link_to "About", "/about", :class => "btn btn-primary", :role => "menuitem" %>
-      <%= link_to "How to Submit", "/how-to-submit", :class => "btn btn-primary", :role => "menuitem" %>
+      <%= link_to "About","#{root_url}about", :class => "btn btn-primary", :role => "menuitem" %>
+      <%= link_to "How to Submit", "#{root_url}how-to-submit", :class => "btn btn-primary", :role => "menuitem" %>
       </nav>
       <nav id="user-actions" class="col-sm">
         <% if current_user %>


app/views/shared/_header.html.erb
```diff
@@ -3,8 +3,8 @@
     <%= link_to "PDC Describe", root_path, class: 'navbar-logo' %>
     <div class="menu">
       <nav id="menu" class="col-sm" role="menu">
-      <%= link_to "About", "/about", :class => "btn btn-primary", :role => "menuitem" %>
-      <%= link_to "How to Submit", "/how-to-submit", :class => "btn btn-primary", :role => "menuitem" %>
+      <%= link_to "About","#{root_url}about", :class => "btn btn-primary", :role => "menuitem" %>
+      <%= link_to "How to Submit", "#{root_url}how-to-submit", :class => "btn btn-primary", :role => "menuitem" %>
```
hectorcorrea, line 7: I think the idiomatic way for this is:

```suggestion
<%= link_to "About",welcome_about_path, :class => "btn btn-primary", :role => "menuitem" %>
<%= link_to "How to Submit", welcome_how_to_submit_path, :class => "btn btn-primary", :role => "menuitem" %>
```
which uses the alias given to these pages (e.g. `welcome_about_path`) in `routes.rb`

hectorcorrea (dismissed):

spec/system/front_page_spec.rb
```diff
@@ -9,7 +9,7 @@
 
   it "has a header with links to helpful info" do
     visit "/"
-    expect(page).to have_link "About", href: "/about"
-    expect(page).to have_link "How to Submit", href: "/how-to-submit"
+    expect(page).to have_link "About"
+    expect(page).to have_link "How to Submit"
```
bess, line 13: This removes the check for the href. Seems like we should fix it instead, otherwise we won't know if it breaks.

bess (changes requested):
I'm going to poke at this for a minute and see if I can figure out what we should be testing for in those URLs.

bess (approved):

Issue #1030, opened 2023-03-08
Describe Form: Renaming "Collection" Label and Rename the "Research Data" Group


hhkitsune:
## Description

The group information currently labeled "collection" is confusing, given PPPL has collections beneath its group. We want to rename the "Collection" label "Group" and be consistent in calling all of the workflow pipelines added to Describe. Additionally, everything going into Describe is Research Data, so it's confusing to call the PRDS group's workflow pipline, "Research data." Let's rename this "Princeton Research Data Service."

## Example
This is what we want to experience for what is currently labeled "Collection" and what is currently entered as "Research data" 

**Group:** Princeton Research Data Service (PRDS)
**Group:** Princeton Plasma Physics Lab (PPPL)

## Acceptance Criteria
- [x] The "Collection" (aka "Community" from DataSpace in Discovery) label will be changed to "Group" in Describe
- [x] The "Research Data" collection will be changed to "Princeton Research Data Service (PRDS)"
- [x] Please consider changing "PPPL" to "Princeton Plasma Physics Lab (PPPL)" for consistency 
- [x] Rename the Collection Class to Group
- [x] Fix any broken tests  

This aspect of the change was proposed for Discovery to address consistency for PPPL. I'm not certain this change should be applied to Discovery. This ticket regards making changes to Describe. Changes to Discovery should be written to another ticket in Github/Discovery, if it's decided to make these changes. For example, we want the "Research Data" collection changed to "Princeton Research Data Service" under "Group" in Describe, because this labels the curatorial pipeline. It doesn't describe the collection, so I have doubts about making these changes in Discovery. 

~~- [ ] In PDC Discovery ([example](https://datacommons.princeton.edu/discovery/?f%5Bcollection_name_ssi%5D%5B%5D=NSTX-U&f%5Bcommunity_root_name_ssi%5D%5B%5D=Princeton+Plasma+Physics+Laboratory&f%5Bsubcommunity_name_ssi%5D%5B%5D=Spherical+Torus)), "Subcommunity" label will be changed to "Collection"~~
[...]

PR #1180, opened 2023-05-17
Removing the Collection model and moving all functions into Group

carolyncole:
Also renaming collection_id column in group options to facilitate the relationship moves

This does not yet rename the table


hectorcorrea (approved):

carolyncole:
refs #1030

PR #1181, opened 2023-05-17
Change GroupOption table to be group_options instead of collection_options

carolyncole:
Additionally renamed the default_collection_id in user to default_group_id


hectorcorrea (approved):

carolyncole:
refs #1030

PR #1184, opened 2023-05-17
Removing collection method from WorkDecorator

carolyncole:
Also moving some comments, ui references, and variables that were still named collection even though a group was being set


hectorcorrea (approved):

carolyncole:
refs #1030

PR #1189, opened 2023-05-17
Trigger the creation of the preservation once work has been approved

### LINKED ISSUE #512: Export object metadata for preservation
Export Datacite XML and object metadata as JSON to a backup location within the DOI directory for the files in S3, as follows:

## Example: 

```bash 
s3:/bucket/10.80021/ge21-2181/92/
    file1.png 
    dataset.zip 
    princeton_data_commons/
        datacite.xml
        metadata.json 
```

- [x] Keep the location configurable so we can easily change the location in the future
- [x] Export the datacite.xml which is the `work.resource.to_xml` in AWS
- [x] Export the metadata.json which is the `work.resource.to_json` in AWS
- [ ] This will occur when the curator approves the work
- [ ] Automated tests for the files getting created when the approve button is pressed

hectorcorrea:
Automatically triggers the creation of the preservation metadata once the last file of the work has been moved to the post-curation bucket.

Triggering the code at the right time was trickier than I expected because the work is not marked as "approved" until *after all the actions in the state machine have completed*. The code to create the preservation files is inside the transition, it's the last step, which makes sense, but the state not being completely saved threw me off for a while. 

I had to add a method to the work to force it to fetch post curation files (regardless of the state of the work in the database) because we want to include the file list as part of the preservation metadata and the files at this stage have all been moved to the post-curation bucket.

Closes #512

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -401,7 +401,18 @@ class Work < ApplicationRecord
 
     s3_resources
   end
-  alias post_curation_uploads post_curation_s3_resources
+
+  # Returns the files in post-curation for the work
+  def post_curation_uploads(force_post_curation: false)
+    if force_post_curation
+      # Always use the post-curation data regardless of the work's status
+      post_curation_s3_query_service = S3QueryService.new(self, false)
+      post_curation_s3_query_service.data_profile.fetch(:objects, [])
+    else
+      # Return the list based of files honoring the work status
+      post_curation_s3_resources
+    end
+  end
 
   def s3_files
     pre_curation_uploads_fast
@@ -434,11 +445,12 @@ class Work < ApplicationRecord
     nil
   end
 
-  def as_json(*)
+  def as_json(*args)
+    force_post_curation = args.any? {|arg| arg[:force_post_curation] == true }
     # Pre-curation files are not accessible externally,
     # so we are not interested in listing them in JSON.
     # (The items in pre_curation_uploads also have different properties.)
-    files = post_curation_uploads.map do |upload|
+    files = post_curation_uploads(force_post_curation:).map do |upload|
       {
         "filename": upload.filename,
         "size": upload.size,
--- a/app/services/work_preservation_service.rb
+++ b/app/services/work_preservation_service.rb
@@ -5,21 +5,18 @@
 # with our AWS credentials, but allows the bucket and path to be configurable.
 class WorkPreservationService
 
-  # @param work [Work] The work to preserve.
+  # @param work [Integer] The ID of the work to preserve.
   # @param bucket_name [String] The AWS S3 bucket name where the work will be preserved.
-  # @param path [String] The path where the work will be preserved, e.g.
-  def initialize(work:, bucket_name: nil, path: nil)
-    @work = work
-    # Defaults to the post curation bucket, e.g. "pdc-describe-staging-postcuration"
-    @bucket_name = bucket_name || @work.s3_query_service.bucket_name
-    # Defaults to the DOI prefix + DOI + work.id, e.g. "10.1234/xy123/10/"
-    @path = path || @work.s3_query_service.prefix
+  # @param path [String] The path where the work will be preserved.
+  def initialize(work_id:, bucket_name:, path:)
+    @work = Work.find(work_id)
+    @bucket_name = bucket_name
+    @path = path
   end
 
   # Creates and stores the preservation files for the work.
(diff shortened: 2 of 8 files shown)


app/models/work.rb
```diff
@@ -401,7 +401,18 @@ def post_curation_s3_resources
 
     s3_resources
   end
-  alias post_curation_uploads post_curation_s3_resources
+
+  # Returns the files in post-curation for the work
```
carolyncole, line 405: @hectorcorrea I would have expected post_curation_uploads to force the correct state by default.  It seems you have found and fixed a bug.  Would it be possible to change the default to true?

app/services/work_preservation_service.rb
```diff
@@ -5,21 +5,18 @@
 # with our AWS credentials, but allows the bucket and path to be configurable.
 class WorkPreservationService
 
-  # @param work [Work] The work to preserve.
+  # @param work [Integer] The ID of the work to preserve.
```
carolyncole, line 8:
```suggestion
  # @param work_id [Integer] The ID of the work to preserve.
```

spec/services/work_preservation_service_spec.rb
```diff
@@ -2,44 +2,22 @@
...
 
-    it "preserves a work to the default location" do
-      subject = described_class.new(work: approved_work)
+    it "preserves a work to the indicated location" do
+      subject = described_class.new(work_id: approved_work.id, bucket_name: bucket_name, path: path)
       expect(subject.preserve!).to eq "s3://example-bucket-post/#{preservation_directory}"
     end
   end
-
-  describe "preserve to custom location" do
```
carolyncole, line 30: Just curious why this was removed?  Was it intentional?

app/models/work.rb
```diff
@@ -401,7 +401,18 @@ def post_curation_s3_resources
...
+  def post_curation_uploads(force_post_curation: false)
```
carolyncole, line 406:
```suggestion
  def post_curation_uploads(force_post_curation: true)
```

carolyncole (changes requested):
@hectorcorrea I think you are seeing an impact of the jobs being inline in development.  Under normal circumstances the jobs would be queued and then the work saved, and then the jobs would be processed.

Maybe we need to force a save of the state before queuing the  jobs, so we do not need special handling just to make things work in development?

I also added just a few small comments in line.

spec/services/work_preservation_service_spec.rb
hectorcorrea: The service now always receives the location rather than default to something, so in a way it always uses a custom location. The test covers both cases, but I just realized the description might be misleading. I'll update it.

app/models/work.rb
hectorcorrea: Yeah, it could be but I didn't want to change the existing behavior on this PR. There are several refactors that we should do to the Work as it relates to the fetching of pre/post curation files, we have way too many methods that do the same/similar things on this class and I am not sure it is the best approach, but I didn't want to include that work on this PR.

app/models/work.rb
hectorcorrea: I didn't want to change the existing behavior on this PR since I am not sure what depends on the current behavior.

spec/services/work_preservation_service_spec.rb
hectorcorrea: Updated: https://github.com/pulibrary/pdc_describe/pull/1189/commits/b1f8f7735063846385f49162a89ce8912588a162

hectorcorrea:
> @hectorcorrea I think you are seeing an impact of the jobs being inline in development. Under normal circumstances the jobs would be queued and then the work saved, and then the jobs would be processed.
> Maybe we need to force a save of the state before queuing the jobs, so we do not need special handling just to make things work in development?

What complicated things for me was that the work did not have the "approved" state until after the `publish` has completed...which is not what I would have expected since that's defined as an `after` action in the AASM: https://github.com/pulibrary/pdc_describe/blob/main/app/models/work.rb#L41-L43

I am happy with the current approach. I think at lot of this will clear up when we clean up the code to fetch the pre/post curation files. I've opened a separate issue to tackle that: https://github.com/pulibrary/pdc_describe/issues/1192

app/models/work.rb
hectorcorrea: New issue for refactor: https://github.com/pulibrary/pdc_describe/issues/1192

carolyncole (approved):

PR #1199, opened 2023-05-22
Remove duplicate uploads for the same file in dspace and AWS

### LINKED ISSUE #1118: Readme is both in AWS and dspace
During the data migration, when the readme is in both aws and DataSpace we should do the following:

- [x] Check the MD5 checksum to see if they match.  If they match choose either one (who cares at that point)
- [x] Check the MD5  If they do not match:  Modify the name to include the location data_space_ and globus_ and save both.
 - [x] Only report the unique files in the "Number of files being migrated"
 - [x] Automated test to be certain the two paths are both working

carolyncole:
If the checksum matches only one file will be uploaded 
If the checksums are different the files will be renamed to show where the duplicates came from and both will be uploaded

Also changed the code that was cloning the s3_files to just put the new AWS filename in the filename_display instead of creating a second copy.  This allowed us to more easily rename the file if needed and should be faster.

fixes #1118

### DIFF
--- a/app/models/migration_upload_snapshot.rb
+++ b/app/models/migration_upload_snapshot.rb
@@ -6,7 +6,7 @@ class MigrationUploadSnapshot < UploadSnapshot
   end
 
   def store_files(s3_files, pre_existing_files: [])
-    self.files = s3_files.map { |file| { "filename" => file.filename, "checksum" => file.checksum, "migrate_status" => "started" } }
+    self.files = s3_files.map { |file| { "filename" => file.filename_display, "checksum" => file.checksum, "migrate_status" => "started" } }
     files.concat pre_existing_files if pre_existing_files.present?
   end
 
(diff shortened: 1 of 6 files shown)


app/models/migration_upload_snapshot.rb
```diff
@@ -6,7 +6,7 @@ def self.from_upload_snapshot(upload_snapshot)
   end
 
   def store_files(s3_files, pre_existing_files: [])
-    self.files = s3_files.map { |file| { "filename" => file.filename, "checksum" => file.checksum, "migrate_status" => "started" } }
+    self.files = s3_files.map { |file| { "filename" => file.filename_display, "checksum" => file.checksum, "migrate_status" => "started" } }
     files.concat pre_existing_files if pre_existing_files.present?
```
jrgriffiniii, line 10: 👍

app/services/pul_dspace_migrate.rb
```diff
@@ -31,28 +34,50 @@ def migration_message
...
+        idx = aws_file_names.index(s3_file.filename_display)
+        if idx.present?
+          check_matching_files(aws_files_only[idx], s3_file, files_to_remove)
+        end
+      end
+      @dspace_files = dspace_files - files_to_remove
+      dspace_files + aws_files_only
+    end
+
+    def dpsace_update_display_to_final_key(dspace_files)
```
jrgriffiniii, line 62: This might need to be `dpsace_update_display_to_final_key`

jrgriffiniii (approved):

Issue #959, opened 2023-02-06
Describe Form: Add Affiliation (and ROR) to the Creator Property Set


hhkitsune:
### General description
The creator property set needs to have the affiliation sub-property ([affiliation subproperty 2.5 in Datacite 4.4](https://support.datacite.org/docs/datacite-metadata-schema-v44-mandatory-properties#25-affiliation)) added. It makes sense to leverage the ability to enter ROR information, as we have for some other properties. However, we also want the ability to enter free text for the affiliation in case the depositor doesn't know the ROR ID to enter. (note: adding ROR will add additional sub-properties to the creator set which become mandatory when the identifier is used.)

### Placement within the Creator Property set
In the form's display, the ROR/Affiliation field should be placed last in the Creator Property set. Here is the order:
1. ORCID ID 
2. Given Name
3. Family Name
4. Affiliation 
5. ROR

### Image shows the order of appearance for the sub-properties
![Screen Shot 2023-02-06 at 1 18 40 PM](https://user-images.githubusercontent.com/50670293/217053770-545e0870-4589-48b9-b92f-c639e9b52b6e.jpeg)

## Display Information for Discovery
I'm adding this as more of a note to think about for the application of our request in this ticket. Please see these guidelines for how to display ROR: https://ror.readme.io/docs/ror-id-display-guidelines-and-logos

## Acceptance Criteria
- [ ] Move the asterisk denoting a required field to be next to Given Name and Family Name only (as opposed to next to Creator)
- [x] The ability to add a free text affiliation is added to the Creator Property set.
- [x] The ability to add a ROR ID is added to the Creator Property set.
- [x] The field for Affiliation will be placed second to last in the set, as shown in the image.
- [x] The field for ROR will be placed at the end of the set, as shown in the image.
- [x] Creator affiliation appears in the DataCite record
- [x] Should work the same as in other places where when they fill in the ROR the Affiliation will get filled in
         This will work the same as The ROR and Funder name for the Funding Reference field 
- [x] This should be stored in the JSONB metadata field along with the rest of the creator information
- [ ] apply https://ror.readme.io/docs/ror-id-display-guidelines-and-logos to the display of the ROR information for both the Funding Reference, the Additional Organizational Contributors and the Creator 
- [ ] Make sure the information is available in the indexing fields for PDC Discovery ( Likely to be present)

refs https://github.com/pulibrary/pdc_discovery/issues/417

hhkitsune:
[...]

PR #1224, opened 2023-05-30
Moving creator add and sort to match with funder mechanism

carolyncole:
The funder mechanism uses more generic javascript and html.erb for setting up the table. 
I believe the funder way will be more easily updatable in the future. 
This work will set up allowing RORs to be added for the creator refs #959


app/javascript/entrypoints/pdc/edit_required_fields.es6
```diff
@@ -1,4 +1,7 @@
-/* eslint class-methods-use-this: ["error", { "exceptMethods": ["isEmptyRow"] }] */
+/* eslint class-methods-use-this: ["error", {
+  "exceptMethods": ["hasCreators", "findEmptyCreator"] }] */
+
+import TableRow from './trable_row.es6';
```
hectorcorrea, line 4: Shouldn't this be `table_row.es6`?

app/javascript/entrypoints/pdc/edit_table_actions.es6
```diff
@@ -0,0 +1,73 @@
+/* eslint class-methods-use-this: ["error", {
+  "exceptMethods": ["create_new_row", "delete_row", "find_empty_row"] }] */
+
+import TableRow from './trable_row.es6';
```
hectorcorrea, line 4: Same typo here (?)

app/views/works/_required_creators_table.html.erb
```diff
@@ -23,26 +23,38 @@
         <th>&nbsp;</th> <!-- delete icon-->
       </tr>
     </thead>
-    <tbody class="sortable"></tbody>
+    <tbody class="sortable">
+      <% creators = if @work.nil? || @work.resource.creators.empty? 
+                      [nil]
```
hectorcorrea, line 28: I would have expected this to be `[]` rather that `[nil]`. Am I reading it wrong?

hectorcorrea (approved):

PR #1229, opened 2023-05-31
Removing ActiveStorage Integration

### LINKED ISSUE #1144: Remove ActiveStorage layer between Uploads and AWS S3
We want to simplify the code that works with uploads `precuration_uploads` and `pre_curation_uploads_fast` exist.  The first is utilizing ActiveStorage and the second interfaces directly with AWS.

We moved away from ActiveStorage when  we realized that it was very slow for large numbers of files.  What we want is for the uploads to act like the readme does and directly upload via the S3QueryService.

- [ ] Move all configuration from the active storage `storage.yml` into the s3 configuration file and remove `active_storage_configuration` from the S3QueryService
- [ ] Remove precuration_uploads from the work and update the code to upload via the S3QueryService.upload_file (The readme does this already for an example)
- [ ] Add/ clean up tests as needed

jrgriffiniii:
Resolves #1144

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -112,6 +112,10 @@ class Work < ApplicationRecord
     def publish_test_doi?
       (Rails.env.development? || Rails.env.test?) && Rails.configuration.datacite.user.blank?
     end
+
+    def presenter_class
+      WorkPresenter
+    end
   end
 
   include Rails.application.routes.url_helpers
@@ -122,10 +126,7 @@ class Work < ApplicationRecord
   end
 
   after_save do |work|
-    unless Rails.env.test?
-      save_pre_curation_uploads
-      s3_query_service.client_s3_files.sort_by(&:filename)
-    end
+    save_pre_curation_uploads
 
     if work.approved?
       work.reload
@@ -207,7 +208,7 @@ class Work < ApplicationRecord
     unless user.has_role? :group_admin, group
       errors.add :base, "Unauthorized to Approve"
     end
-    if pre_curation_uploads_fast.empty? && post_curation_uploads.empty?
+    if pre_curation_uploads.empty? && post_curation_uploads.empty?
       errors.add :base, "Uploads must be present for a work to be approved"
     end
     errors.count == 0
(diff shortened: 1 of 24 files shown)


app/services/s3_query_service.rb
```diff
@@ -122,14 +122,17 @@ def find_s3_file(filename:)
...
   def client_s3_files(reload: false, bucket_name: self.bucket_name, prefix: self.prefix, ignore_directories: true)
+    return [] unless model.persisted?
+
     @client_s3_files = nil if reload # force a reload
     @client_s3_files ||= begin
       start = Time.zone.now
       resp = client.list_objects_v2({ bucket: bucket_name, max_keys: 1000, prefix: prefix })
       resp_hash = resp.to_h
       objects = parse_objects(resp_hash, ignore_directories: ignore_directories)
-      objects += parse_continuation(resp_hash, bucket_name: bucket_name, prefix: prefix, ignore_directories: ignore_directories)
```
hectorcorrea, line 131: Why was this line removed? I seem to remember that this is what fetches when there are more than 1000 files. Are we doing this differently now?

app/models/work.rb
```diff
@@ -377,6 +380,53 @@ def current_transition
...
+    alias append <<
+  end
+
+  def pre_curation_uploads_fast
+    s3_resources = s3_query_service.client_s3_files.sort_by(&:filename)
+    UploadsProxy.new(model: self, values: s3_resources)
+  end
+  alias pre_curation_uploads pre_curation_uploads_fast
+
+  # Deprecate
```
jrgriffiniii, line 420: This still needs to be removed, I apologize.

app/models/work.rb
```diff
@@ -377,54 +382,57 @@ def current_transition
...
+    alias append <<
+  end
+
+  # Accesses the S3 file uploads for Works in pre-curation/preapproved states
+  # @return [Array<S3File>]
+  def pre_curation_uploads
+    values = s3_query_service.client_s3_files.sort_by(&:filename)
 
-    save_new_attachments(new_attachments: new_attachments)
+    UploadsProxy.new(model: self, values: values)
```
jrgriffiniii, line 425: This is breaking the test suite (due to an absence of mocks for HTTP responses)

Issue #1228, opened 2023-05-23
Describe Form: New Property for "Subcommunity"


hhkitsune:

kelynch:
Using estimate from #1031

Issue #1227, opened 2023-05-22
Describe Form: New Property for "Community"


hhkitsune:
## Description

"Community" is currently a facet in Discovery that is reflecting from DataSpace. We would like to keep this faceting when PDC goes live and offer this option for future submissions. We do not currently have a way to designate that a record belongs to a specific community in Describe. We want to add a property named, "Community" that will be added to the Describe form for for every group, under the "Additional Metadata" Tab. This property will have a controlled vocabulary that includes all of the current listings under the "Community" facet in Discovery, but will have the flexibility to add to or change the listings as needed. This property will allow multiple selections for a single record. For example, a record may belong to both "Music and Arts" and "Digital Humanities." This necessitates the field for the "Community" property to have a multi-selection tool of some kind.

(note: This is a localized property that is not shared with Datacite.)

### List of communities to add to the controlled list
- Princeton Plasma Physics Laboratory
- Princeton Neuroscience Institute
- Department of Geosciences
- Mechanical and Aerospace Engineering
- Astrophysical Sciences
- Civil and Environmental Engineering
- Chemical and Biological Engineering
- Digital Humanities
- Music and Arts
- Princeton School of Public and International Affairs

### PPPL Specific consideration
Selecting "Princeton Plasma Physics Laboratory" should trigger "Subcommunity".

### Acceptance Criteria
- [ ] A localized property for "Community" will be added to the Describe form under the *Additional Metadata* tab
- [ ] The property is local and will not be included with information sent to Datacite
- [ ] The property will reflect information to the Discovery Interface that corresponds to the "Community" Tab
- [ ] The entry field for the "Community" property will allow selection for multiple communities for one record
- [ ] The entry field for the "Community" property will contain a controlled list (see info above) that we may add to or change as needed in the future.

PR #1235, opened 2023-06-01
Adds support for communities and subcommunities

### LINKED ISSUE #1031: Describe Form: Add "Collection" to the deposited work for PPPL Group
## Description
The UI will add a new set of form elements when the New Group Concept in #1030 is changed to PPPL.
We need a location to store this information AND the visibility of this on the form (only when the group is PPPL)

Ticket #1030 renamed "Collection" to Group, when describing a group's workflow pipeline. This ticket adds a new label and drop-down entry field for describing a collection, such as those PPPL has. The drop-down will contain a controlled list for PPPL collections. We will not allow nested collections and this may need to be discussed further with the primary stakeholder to ensure the collections are optimally defined. The new "Collection" field will also allow depositors to add more than one. Adding at least one collection will be required for PPPL depositors. The collection field will not show for depositors to the PRDS group's workflow pipeline

## Acceptance Criteria
- [ ] The label "Collection" is added to the required metadata page with a drop-down entry field (if PPPL is the Group).
- [ ] The drop down entry field will contain a controlled list of PPPL Collections. See comments below.
- [ ] The depositor may optionally add another PPPL collection which their submission may belong to. This will happen using the same kind of feature shown for the Funding Reference, which allows users to add another.
- [ ] The depositor must select at least one collection when making their deposit.  The form should not allow the user to continue to the Readme is a collection is not set.
- [ ] This will be stored in the JSONB metadata field so that class needs to be updated to retrieve and store it.
- [ ] Automated test for this function
- [ ] Make sure the indexing information created for PDC Discovery has the new collection

See https://github.com/pulibrary/pdc_discovery/issues/416 for indexing of this new information in PDC Discovery

hectorcorrea:
Allows a user to select communities and subcommunities for their datasets.

The list of communities available to select depends on the group that the dataset is linked to. This means that PRDS submissions have a different list of communities available to them than PPPL submissions. In fact, PPPL only has one community available to them, but PPPL submissions have an extra field where users can select one or more subcommunities for their dataset.

More notes and discussion about the implementation available here: https://docs.google.com/document/d/1IRy2C_NFpODDnDTPKXQMZZx0NodXb-aVhIriGax3AwU/edit#

Below are a couple of screenshots showing how the edit screen looks for a PRDS submission vs a PPPL submission:

prds

pppl

Closes #1031 #1227 #1228


app/models/group.rb
```diff
@@ -153,4 +153,48 @@ def delete_permission(current_user, removed_user)
       errors.add(:delete_permission, "Unauthorized")
     end
   end
+
```
hectorcorrea, line 156: Eventually we might want to make communities and subcommunities configurable somehow, but for now we are just trying to see if this new design works for PPPL.

carolyncole (approved):

PR #1253, opened 2023-06-09
Optimizes the file list generation

### LINKED ISSUE #1156: Error when loading works with large number of files
Loading a work that has a lot of files sometimes gives a JavaScript error as shown in the following screenshot.

Notice that the AJAX call to `/file-list` took a long time (I think 2 minutes) before it returned an error rather than the expected JSON response. You can also see that the response is what the load balancer shows rather than an "Rails" error message.

Screenshot 2023-05-12 at 2 29 10 PM

We should figure out why these calls are taking so long (we already optimized this code) and we should handle this error more gracefully. 

Note: This error happened after the load issues reported in https://github.com/pulibrary/pdc_describe/issues/1154 and after Francis/Alicia biggie-size the servers -- the servers were not overloaded at the time of this issue.

hectorcorrea:
Optimize the code that produces the JSON file list for a work to prevent it from nesting the work itself for everyfile on the list.

Closes #1156

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -491,8 +491,6 @@ class Work < ApplicationRecord
     }
   end
 
-
-
   def pre_curation_uploads_count
     s3_query_service.file_count
   end
@@ -513,8 +511,6 @@ class Work < ApplicationRecord
   # Build or find persisted UploadSnapshot models for this Work
   # @return [UploadSnapshot]
   def reload_snapshots
-    start = Time.zone.now
-    Rails.logger.info "SNAPSHOTS: reload_snapshots started"
     work_changes = []
     s3_files = pre_curation_uploads_fast
     s3_filenames = s3_files.map(&:filename)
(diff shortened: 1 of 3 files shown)


app/models/work.rb
```diff
@@ -487,16 +508,6 @@ def past_snapshots
...
-  # Remove the snapshots for S3 file resources which have been deleted
-  def valid_snapshots
-    s3_resource_urls = s3_files.map(&:url)
-    s3_resource_filenames = s3_files.map(&:filename)
-
-    past_snapshots.select do |snapshot|
-      s3_resource_urls.include?(snapshot.url) || s3_resource_filenames.include?(snapshot.filename)
-    end
-  end
-
```
hectorcorrea, line 499: We don't use this method. I noticed that is extra while troubleshooting this issue since we initially thought the problem was with the "snapshots code"

bess (approved):

PR #1277, opened 2023-06-21
Fixes issue fetching ORCID values

### LINKED ISSUE #1271: ORCID search fails for some ORCID IDs
We noticed that the ORCID search does not return the user information for the following ORCIDs:

* 0000-0002-8201-2528 (Alexander	Nagy)
* 0000-0002-7348-7142 (Dang	Cai)

There are some JavaScript errors for these ids. The code works OK for most other IDs

See also #1161

hectorcorrea:
Forces the request to the ORCID API to pass `application/json` on the HTTP headers. This fixes the problem where a few requests return a response that jQuery cannot parse -- even though the majority of the calls work OK with the previous configuration using JSONP via `dataType: 'jsonp'`

Closes #1271

### DIFF
--- a/app/javascript/entrypoints/pdc/work_orcid.es6
+++ b/app/javascript/entrypoints/pdc/work_orcid.es6
@@ -22,14 +22,14 @@ export default class WorkOrcid {
     if (this.isOrcidFormat(orcidValue)) {
       $.ajax({
         url: `${pdc.orcid_url}/${orcidValue}`,
-        beforeSend: function( xhr ) {
+        beforeSend(xhr) {
           // The ORCID API is very finicky and does not reliably work if the
           // "`Accept: application/json" request header is not present.
           //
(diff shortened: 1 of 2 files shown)


app/javascript/entrypoints/pdc/work_orcid.es6
```diff
@@ -22,16 +22,23 @@ export default class WorkOrcid {
     if (this.isOrcidFormat(orcidValue)) {
       $.ajax({
         url: `${pdc.orcid_url}/${orcidValue}`,
-        dataType: 'jsonp',
+        beforeSend(xhr) {
+          // The ORCID API is very finicky and does not reliably work if the
+          // "`Accept: application/json" request header is not present.
```
jrgriffiniii, line 27: 😞

jrgriffiniii (approved):

PR #1283, opened 2023-06-22
Display basic information about users in the Users List page

hectorcorrea:
user-list

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -8,7 +8,7 @@ class UsersController < ApplicationController
   before_action :authenticate_user!
 
   def index
-    @users = User.all.sort_by(&:family_name)
+    @users = User.all.sort_by { |user| user.family_name || "" }
   end
 
   # GET /users/1
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -4,6 +4,10 @@ class UsersController < ApplicationController
   GROUP_MESSAGING_DISABLED = "0"
   GROUP_MESSAGING_ENABLED = "1"
 
+  # Notice that `set_user` sets the value of the user that we are viewing or editing
+  # while `authenticate_user` sets the value of the current logged in user.
+  # These values can be different (e.g. when an admin users is editing the information
+  # of another user)
   before_action :set_user, only: %i[show edit update]
   before_action :authenticate_user!
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -5,9 +5,10 @@ class UsersController < ApplicationController
   GROUP_MESSAGING_ENABLED = "1"
 
   before_action :set_user, only: %i[show edit update]
+  before_action :authenticate_user!
 
   def index
-    @users = User.all
+    @users = User.all.sort_by(&:family_name)
   end
 
   # GET /users/1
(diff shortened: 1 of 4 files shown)


spec/requests/users_spec.rb
```diff
@@ -28,17 +28,6 @@
...
-  # via CAS authentication.
-  describe "GET /new" do
-    it "redirects to CAS" do
-      get "/users/new"
-      expect(response.code).to eq "302"
-      redirect_location = response.header["Location"]
-      expect(redirect_location).to eq "http://www.example.com/sign_in"
-    end
-  end
-
```
hectorcorrea, line 41: I don't think we need this test. The authenticate method will force the process to go through CAS. 

Is it possible that we have a user with a netid of `new`?

app/controllers/users_controller.rb
```diff
@@ -5,9 +5,10 @@ class UsersController < ApplicationController
   GROUP_MESSAGING_ENABLED = "1"
 
   before_action :set_user, only: %i[show edit update]
+  before_action :authenticate_user!
```
hectorcorrea, line 8: Notice that we now force uses to be authenticated before getting to this page.

carolyncole, line 8: @hectorcorrea This line is redundant, and I think it will add confusion https://github.com/pulibrary/pdc_describe/blob/aa589ca8600480114ac307bf5155334433ea229e/app/controllers/application_controller.rb#L10

Can you remove it from the PR?

carolyncole (changes requested):
As long as you remove the users method as requested this looks great

app/controllers/users_controller.rb
hectorcorrea: @carolyncole good point. I added a comment to the code to indicate why we need both.

app/controllers/users_controller.rb
```diff
@@ -4,10 +4,15 @@ class UsersController < ApplicationController
   GROUP_MESSAGING_DISABLED = "0"
   GROUP_MESSAGING_ENABLED = "1"
 
+  # Notice that `set_user` sets the value of the user that we are viewing or editing
+  # while `authenticate_user` sets the value of the current logged in user.
+  # These values can be different (e.g. when an admin users is editing the information
+  # of another user)
```
hectorcorrea, line 10: @carolyncole this is the comment to clarify why we need both.

app/views/users/index.html.erb
```diff
@@ -1,10 +1,30 @@
...
+    </tr>
+  </thead>
+  <tbody>
+    <% @users.each do |user| %>
+    <tr class="user-list-data-row">
+      <td>
+        <%= link_to(user.uid, user_path(user)) %>
+        <% if user.uid == current_user.uid %>
+          &nbsp;(you)
+        <% end %>
```
hectorcorrea, line 20: This is why I want to have access to `current_user`. I want to indicate the user which one is _their_ record, in this case we display the label "(you)" right next to their name.

app/controllers/users_controller.rb
carolyncole: @hectorcorrea is this a timing thing?  Aka the authenticate_user! has to happen after the set_user?

hectorcorrea: @carolyncole I don't think so, I suspect it could go before it too. Would that make more sense?

carolyncole: From slack: We are now suspectingthere may be a timing issue at play and we will leave it in palce here

carolyncole (approved):

PR #1284, opened 2023-06-22
Use Base64 encoded MD5 checksum as required by updated AWS S3 API

### LINKED ISSUE #1273: Data migration bug: The Content-MD5 you specified was invalid
Automated data migration is broken. When you push the button it gives the attached error, but no error is raised to Honeybadger. 

Screenshot 2023-06-20 at 2 00 10 PM

bess:
Fixes #1273 

Co-authored-by: Carolyn Cole <carolyncole@users.noreply.github.com>

### DIFF
--- a/app/services/pul_dspace_connector.rb
+++ b/app/services/pul_dspace_connector.rb
@@ -35,7 +35,7 @@ class PULDspaceConnector
     bitstreams.map do |bitstream|
       filename = download_bitstream(bitstream["retrieveLink"], bitstream["name"])
       if checksum_file(filename, bitstream)
-        S3File.new(filename: filename, checksum: bitstream["checkSum"]["value"], last_modified: DateTime.now, size: -1, work: work)
+        S3File.new(filename: filename, checksum: bitstream["checkSum"]["base64"], last_modified: DateTime.now, size: -1, work: work)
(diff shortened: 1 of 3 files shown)


app/services/pul_dspace_connector.rb
```diff
@@ -87,9 +87,14 @@ def download_file(url, filename)
...
 
+    # rubocop:disable Metrics/MethodLength
     def checksum_file(filename, bitstream)
       checksum_class = Digest.const_get(bitstream["checkSum"]["checkSumAlgorithm"])
-      if checksum_class.file(filename).hexdigest != bitstream["checkSum"]["value"]
+      checksum = checksum_class.file(filename)
+      hexdigest = checksum.hexdigest
+      base64 = checksum.base64digest
+      bitstream["checkSum"]["base64"] = base64
+      if hexdigest != bitstream["checkSum"]["value"]
```
hectorcorrea, line 97: Just double-checking that this was indeed meant to be `value` and not `base64`

bess: @hectorcorrea Yes, that's correct as-is. DSpace needs the hexdigest and S3 needs the base64digest.

PR #1292, opened 2023-06-27
Include provenance information in preservation files

### LINKED ISSUE #1214: Export provenance log for preservation
In issue #512 we exported the metadata for a dataset for preservation purposes. This ticket is a continuation of that work to also export the provenance log and the messages associated with the dataset.

We could export this data as separate files, there is not requirement to included it all in metadata.json

hectorcorrea:
Added the provenance information to the files that we create during preservation. This information includes the user entered messages and also the history of changes to the metadata on the record.

I also added an option to save the preservation files locally (rather than always saving them on S3). This was not part of the original request but it is really handy to have this option and use it via the original rake task:

```
$ bundle exec rake works:preserve[1,localhost,preservation_test/1]
file:///Users/correah/src/pdc_describe/preservation_test/1/princeton_data_commons/
```

Closes #1214

### DIFF
--- a/app/services/work_preservation_service.rb
+++ b/app/services/work_preservation_service.rb
@@ -7,6 +7,7 @@ class WorkPreservationService
 
   # @param work_id [Integer] The ID of the work to preserve.
   # @param bucket_name [String] The AWS S3 bucket name where the work will be preserved.
+  #    If the bucket name is "localhost" the preservation files will be saved to disk on the localhost.
   # @param path [String] The path where the work will be preserved.
   def initialize(work_id:, bucket_name:, path:)
     @work = Work.find(work_id)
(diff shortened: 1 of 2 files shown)


app/services/work_preservation_service.rb
```diff
@@ -7,6 +7,7 @@ class WorkPreservationService
 
   # @param work_id [Integer] The ID of the work to preserve.
   # @param bucket_name [String] The AWS S3 bucket name where the work will be preserved.
+  #    If the bucket name is "localhost" the preservation files will be saved to disk on the localhost.
```
jrgriffiniii, line 10: 👍

app/services/work_preservation_service.rb
```diff
@@ -48,7 +62,19 @@ def s3_client
     end
 
     def create_preservation_directory
-      s3_client.put_object({ bucket: @bucket_name, key: preservation_directory.to_s, content_length: 0 })
+      if is_local?
```
jrgriffiniii, line 65: 👍

jrgriffiniii (approved):

Issue #1293, opened 2023-06-28
Creator field display changes & Help Text


kelynch:
## Description
(Kate; 6/28/23) Creators field display changes per meeting with Hannah today.

This ticket outlines visual changes to the Creators property set in the Describe form. We are making changes to better guide users to perform the mandatory steps without feeling anxious if they do not have additional information for the other options, such as the ROR. The fields are moved so the required fields (i.e., given and family name) come first. Larger asterisks will show next to the "Creators," "Given name," and "Family name" fields. We are also disambiguating the button for the help text by changing the "?" to "See more" and taking some of the essential help text out from behind this feature. The "See more" button will be used to hide lengthy help, rather than hiding all help text.

## Wireframe shows visual changes to the Creators property set

![Screen Shot 2023-06-28 at 4 41 22 PM](https://github.com/pulibrary/pdc_describe/assets/50670293/80237b40-f259-43c6-b209-2b154b60b778)

## Acceptance Criteria
- [ ] Larger asterisks for required fields (~24px use multiple of em)
- [x] Field order changing as follows (required fields marked with "*")
  - Creator * 
    - Given name * 
    - Family name * 
    - ORCiD
    - Affiliation
    - ROR

hhkitsune:
My apologies for leaving the "ORCID" field out of the wireframe. As Kate had mentioned, it should go after the "Family Name" field.

carolyncole:
Finally got https://github.com/pulibrary/pdc_describe/pull/1322 to pass

PR #1322, opened 2023-07-13
Added a class to the required star to make it more visible

carolyncole:
We made the start 1.5 em since that seemed to match with the change in the ticket.

Additionally fixes a bug in the work_spec where the new ROR was not included in the bi directional second drag, so the test was dragging the row forward and then backward.  Why making the star larger surfaced this issue, I am unsure...

refs #1293

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -32,6 +32,10 @@
   color: #d71616;
 }
 
+.required-field-star {
+  font-size: 1.5em;
+}
+
 .wizard-area {
   max-width: 960px; /* does this value need bang important too? */
 }
--- a/app/views/works/_curator_controlled_form.html.erb
+++ b/app/views/works/_curator_controlled_form.html.erb
@@ -7,13 +7,13 @@
 
 <!-- Publisher field -->
 <div class="field">
-  Publisher:<span class="required-field">*</span><br>
+  Publisher:<span class="required-field required-field-star">*</span><br>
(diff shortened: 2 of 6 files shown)


hectorcorrea:
Go home CircleCI --  you are drunk. Re-running the test to see if that fixes it.

hectorcorrea (dismissed):

hectorcorrea:
The test failure seems to be a real one in this case (?)
```
expected: "Sally Smith National Science Foundation https://ror.org/021nxhr62  Quyen Jacobi"
got: "Quyen Jacobi    Sally Smith National Science Foundation https://ror.org/021nxhr62"

```

carolyncole:
> The test failure seems to be a real one in this case (?)
> 
> ```
> expected: "Sally Smith National Science Foundation https://ror.org/021nxhr62  Quyen Jacobi"
> got: "Quyen Jacobi    Sally Smith National Science Foundation https://ror.org/021nxhr62"
> ```

It does not fail locally for me, but does repetitively in CI.  🤞 rebasing on flapping test  branch will solve it.

carolyncole:
OMG!  Ok, so after a bit of digging it is the change in the `app/assets/stylesheets/application.scss` that is breaking the test.  I removed that change and the tests went green...  I'll do some more digging

carolyncole:
Ok, so after digging into this further making the stars bugger in the table header are messing with the drag and drop in some way.

hectorcorrea (approved):

PR #1328, opened 2023-07-13
Removing save of pre_curation_uploads since we no longer utilize them

carolyncole:
We are no storing our files via AttachFileToWorkJob


hectorcorrea (approved):

hectorcorrea:
yippie!

PR #1335, opened 2023-07-14
Extending S3QueryService#client invocations of #get_object with #get_object_attributes for S3 Object metadata

### LINKED ISSUE #1075: AWS get_object gets the data with IO, should utilize get_object_attributes for metadata
For example to get the metadata only you could call:
```
client.get_object_attributes({bucket: bucket_name,
                                   key: resp.key, object_attributes: ["ETag","Checksum","ObjectSize","ObjectParts"]})
```
which returns
```
#<struct Aws::S3::Types::GetObjectAttributesOutput delete_marker=nil, last_modified=2023-04-19 12:45:04 +0000, version_id="5Y8b5cZyIsUlvN38NkVZJYjAMImL023L", request_charged=nil, etag="10d9365a047b19a73d192dcac50b75f8-1", checksum=nil, object_parts=#<struct Aws::S3::Types::GetObjectAttributesParts total_parts_count=1, part_number_marker=nil, next_part_number_marker=nil, max_parts=nil, is_truncated=nil, parts=[]>, storage_class=nil, object_size=167772848>
```

jrgriffiniii:
Resolves #1075

### DIFF
--- a/app/services/s3_query_service.rb
+++ b/app/services/s3_query_service.rb
@@ -121,8 +121,6 @@ class S3QueryService
   end
 
   def get_s3_object(key:)
-    object_attributes = get_s3_object_attributes(key: key)
-
     response = client.get_object({
       bucket: bucket_name,
       key: key
@@ -130,7 +128,7 @@ class S3QueryService
     object = response.to_h
     return if object.empty?
 
-    object.merge(object_attributes)
+    object
   end
 
   def build_s3_object_key(filename:)
--- a/app/services/s3_query_service.rb
+++ b/app/services/s3_query_service.rb
@@ -75,7 +75,7 @@ class S3QueryService
   # Public signed URL to fetch this file from the S3 (valid for a limited time)
   def file_url(key)
     signer = Aws::S3::Presigner.new(client: client)
-    signer.presigned_url(:get_object, bucket: bucket_name, key: key)
+    signer.presigned_url(:get_object_attributes, bucket: bucket_name, key: key)
   end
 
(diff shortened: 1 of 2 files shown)


app/services/s3_query_service.rb
```diff
@@ -74,7 +74,7 @@ def s3_address
   # Public signed URL to fetch this file from the S3 (valid for a limited time)
   def file_url(key)
     signer = Aws::S3::Presigner.new(client: client)
-    signer.presigned_url(:get_object, bucket: bucket_name, key: key)
+    signer.presigned_url(:get_object_attributes, bucket: bucket_name, key: key)
```
carolyncole, line 77: Since this is the download link, I think we want to leave it as get_object

app/services/s3_query_service.rb
```diff
@@ -99,15 +99,37 @@ def client
...
+      key: key,
+      object_attributes: self.class.object_attributes
+    })
+    response.to_h
+  end
+
   def get_s3_object(key:)
+    object_attributes = get_s3_object_attributes(key: key)
+
     response = client.get_object({
```
carolyncole, line 125: This was the line I thought should change to get_object_attributes. It is only used here:https://github.com/pulibrary/pdc_describe/blob/7bb589a62ec74204f59431f934d909a08a492c37/app/services/s3_query_service.rb#L116 Which is just utilized for creating an S3File which is all about the attributes not the file data

carolyncole (changes requested):

carolyncole (approved):

PR #1359, opened 2023-07-20
Added first pass object recreation documentation

### LINKED ISSUE #516: Test preservation backup of a single object
Test restoring a single object's files and metadata from preservation backup (references #512)

- [x] With the data in #512 reassemble one object
- [x] Document the process in pdc_describe/docs

kelynch:
Closes #516


docs/preservation_recreate.md
```diff
@@ -0,0 +1,39 @@
...
+    * datacite.xml
+    * metadata.json 
+    * provenance.json
+```
+
+## Recreating an object from S3
+
+1. Locate the DOI folders in the appropriate bucket is S3.
+1. Download all of the files within the folder associated with the object, including the contents of the `princeton_data_commons` folder, which should contain three files as noted in the example directory structure aboveabove.
+1. Repopulate the metadata field values in the PDC Describe interface using the contents of `metadata.json`.  This file contains non-markdown formatting information such as line breaks that do not belong in the `datacite.xml` file.
```
hectorcorrea, line 32: We should add a note here to indicate that if the `metadata.json` was generated with a version of the Work that is not compatible with the current version in the system (e.g. `metadata.json` was produced 10 months ago and PDC Describe has new fields and perhaps some of them mandatory) then the restore process won't be an easy process. Doable for sure, but not easy.

hectorcorrea (approved):

PR #1387, opened 2023-07-28
Stores preservation files in an S3 bucket with a different availability region

### LINKED ISSUE #1380: Save preservation files in a separate bucket
Save preservation files (`datacite.xml`, `metadata.json`, `provenance.log`) in a separate bucket so that we don't make them available to end users through Globus.

hectorcorrea:
Updated the `WorkPreservationService` to store the preservation files on a new S3 bucket and make sure this bucket can be in a different S3 availability region. For example, we store post curation data on `pdc-describe-postcuration-production` bucket in the `us-east-1` region but our preservation files in `pdc-describe-preservation-production` in the `us-west-1` region.

Closes #1380

~~This PR is ready for review but I marked it as `DRAFT` so we don't accidentally deploy it before we reprovision our staging and production servers with the new ENV values for the preservation bucket (https://github.com/pulibrary/princeton_ansible/pull/4096)~~ Staging and prod environments have been updated.


app/services/s3_query_service.rb
```diff
@@ -19,31 +19,44 @@ def self.post_curation_config
     configuration.post_curation
   end
 
+  def self.preservation_config
+    configuration.preservation
+  end
+
```
hectorcorrea, line 25: This is really the key of this PR. I added a new configuration setting that allows the `S3QueryService` to use a new bucket + region settings that are defined for preservation uses in s3.yml.

spec/jobs/approved_file_copy_job_spec.rb
```diff
@@ -4,41 +4,64 @@
...
+      described_class.perform_later(work_id: work.id, source_bucket: "example-bucket", source_key: "10.34770/ackh-7y71/1/test_key", target_bucket: "example-bucket-post",
+                                    target_key: "10.34770/ackh-7y71/1/test_key", size: 200)
+    end
+    let(:fake_s3_service) { stub_s3 }
+    let(:work) { FactoryBot.create :approved_work }
 
-  context "the copy fails" do
     before do
-      fake_completion = instance_double(Seahorse::Client::Response, "successful?": false)
+      fake_completion = instance_double(Seahorse::Client::Response, "successful?": true)
```
hectorcorrea, line 16: This diff is misleading. I split the test in two in order to pass different buckets, but the the way the diff renders it looks like I changed the last value in the call to `instance_double()` from `false` to `true` but I didn't, if you look at like 15 in the diff you see that it was `true` before.

app/services/s3_query_service.rb
```diff
@@ -19,31 +19,44 @@ def self.post_curation_config
...
 
   ##
   # @param [Work] model
-  # @param [Boolean] pre_curation
+  # @param [Boolean] pre_curation True to use pre_curation S3 buckets, false to use post_curation S3 buckets.
+  # @param [Boolean] preservation True to use our preservation buckets in AWS S3
+  #                               (when true it takes precedence over the pre_curation parameter)
   # @example S3QueryService.new(Work.find(1), true)
-  def initialize(model, pre_curation = true)
+  def initialize(model, pre_curation = true, preservation = nil)
```
carolyncole, line 34: I wonder instead of two flags if it would make more sense to pass a mode parameter that could have the value or precuration, postcuration, or preservation.  It just seems weird for precuration to be true, but still be false if preservation is true.

```suggestion
  def initialize(model, mode = "pre_curation")
```

app/services/s3_query_service.rb
```diff
@@ -19,31 +19,44 @@ def self.post_curation_config
...
     @model = model
     @doi = model.doi
     @pre_curation = pre_curation
+    @preservation = (preservation == true)
```
carolyncole, line 38:
```suggestion
    @preservation = mode == "preservation"
```

app/services/s3_query_service.rb
```diff
@@ -19,31 +19,44 @@ def self.post_curation_config
...
+      self.class.preservation_config
+    elsif post_curation?
+      self.class.post_curation_config
+    else
+      self.class.pre_curation_config
+    end
   end
 
   def pre_curation?
+    return false if @preservation
```
carolyncole, line 54: See comment above about making this a mode check.  This change would not be needed
```suggestion
```

app/services/s3_query_service.rb
```diff
@@ -19,31 +19,44 @@ def self.post_curation_config
...
     @pre_curation
   end
 
   def post_curation?
+    return false if @preservation
```
carolyncole, line 59: And this line too
```suggestion
```

app/services/work_preservation_service.rb
```diff
@@ -73,15 +74,19 @@ def preservation_directory
       Pathname.new(@path).join("#{PRESERVATION_DIRECTORY}/")
     end
 
-    def s3_client
-      @work.s3_query_service.client
+    def s3_query_service
+      @s3_query_service ||= S3QueryService.new(@work, false, true)
```
carolyncole, line 78: Then this would be a mode
```suggestion
      @s3_query_service ||= S3QueryService.new(@work, "preservation")
```

carolyncole (changes requested):
@hectorcorrea I would prefer to change the initialize.  I believe this line would also have to change for the mode https://github.com/pulibrary/pdc_describe/blob/3ae84e5ff8acb139a758c50e9619cc5cd56f087b/app/jobs/approved_file_move_job.rb#L7. The logic just seems harder with two flags vs a mode...

app/services/s3_query_service.rb
```diff
@@ -19,32 +23,43 @@ def self.post_curation_config
...
 
   ##
   # @param [Work] model
-  # @param [Boolean] pre_curation
-  # @example S3QueryService.new(Work.find(1), true)
-  def initialize(model, pre_curation = true)
+  # @param [String] mode Valid values are "precuration", "postcuration", "preservation".
+  #                      This value controlls the AWS S3 bucket used to access the files.
+  # @example S3QueryService.new(Work.find(1), "precuration")
+  def initialize(model, mode = "precuration")
```
hectorcorrea, line 37: @carolyncole this is the new `mode` parameter as you suggested.

app/services/s3_query_service.rb
```diff
@@ -19,32 +23,43 @@ def self.post_curation_config
...
   end
 
   def pre_curation?
-    @pre_curation
+    @mode == PRECURATION
   end
 
   def post_curation?
-    !pre_curation?
+    @mode == POSTCURATION
```
carolyncole, line 62: ❤️

carolyncole (approved):
The mode looks great @hectorcorrea

PR #1395, opened 2023-08-04
Utilizing wget so that the download can restart when the dspace connection fails

carolyncole:
Also change to run through the entire download upload delete cycle for one file at a time. This should allow for better restart capabilities and also keep disk usage down.


carolyncole:
The snapshots look weird because of this issue: https://github.com/pulibrary/pdc_describe/issues/1396

app/services/pul_dspace_connector.rb
```diff
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 class PULDspaceConnector
-  attr_reader :work, :ark
+  attr_reader :work, :ark, :download_base
 
   def initialize(work)
     @work = work
     @ark = work.ark&.gsub("ark:/", "")
+    @download_base = "#{Rails.configuration.dspace.base_url.gsub('rest/', '')}bitstream/#{ark}"
```
jrgriffiniii, line 8: 👍

jrgriffiniii (approved):

PR #1405, opened 2023-08-09
Handles encoded checksums when comparing upload snapshots

### LINKED ISSUE #1396: AWS is sometimes returning MD5 checksums and sometimes it is returning base64 encoded MD5 checksums
This is happening in the code that checks if the files in s3 have changed in the background.

From what I can tell the checksum is sometimes just plain MD5 and sometimes it is base64.  I'm not sure the rime or reason for this, but when we compare the checksums we should compare the lengths and if they are different try unpacking or packing the string.
```
checksum = "98691a716ece23a77735f37b5a421253"
base64 = [[checksum].pack("H*")].pack("m0"). # "mGkacW7OI6d3NfN7WkISUw=="
base64.unpack("m0").first.unpack('H*').first # "98691a716ece23a77735f37b5a421253"
```

This is making files look like they have been replaced when really they have not been...

hectorcorrea:
Closes #1396

### DIFF
--- a/app/models/upload_snapshot.rb
+++ b/app/models/upload_snapshot.rb
@@ -22,7 +22,7 @@ class UploadSnapshot < ApplicationRecord
   end
 
   def index(s3_file)
-    files.index { |file| file["filename"] == s3_file.filename && file["checksum"] == s3_file.checksum }
+    files.index { |file| file["filename"] == s3_file.filename && checksum_compare(file["checksum"], s3_file.checksum) }
   end
 
   def match?(s3_file)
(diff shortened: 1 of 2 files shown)


jrgriffiniii (approved):

app/models/upload_snapshot.rb
```diff
@@ -42,4 +42,24 @@ def self.find_by_filename(work_id:, filename:)
...
+    # a plain MD5 value and the other has been encoded with base64.
+    # See also
+    #   https://ruby-doc.org/core-2.7.0/Array.html#method-i-pack
+    #   https://ruby-doc.org/core-2.7.0/String.html#method-i-unpack
+    def checksum_compare(checksum1, checksum2)
+      if checksum1 == checksum2
+        true
+      elsif checksum1.length < checksum2.length
+        # Decode the first one and then compare
+        checksum1.unpack("m0").first.unpack("H*").first == checksum2
```
jrgriffiniii, line 56: 👍

app/models/upload_snapshot.rb
```diff
@@ -42,4 +42,24 @@ def self.find_by_filename(work_id:, filename:)
...
+      else
+        # Decode the second one and then compare
+        checksum1 == checksum2.unpack("m0").first.unpack("H*").first
```
jrgriffiniii, line 59: 👍

Issue #1383, opened 2023-07-27
Allow PPPL users to configure notifications at the subcommunity level


carolyncole:
PPPL users would like to be able to just turn on messages for the subcommunities they care about.

Build on #1382 to allow each subcommunity to be a different kind of message type. Allow the user to choose the sub communities they care about, but by default they will all be turned on.

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-Workcycles-61a4f1a12a399b001730f65a/issues/pulibrary/pdc_describe/1383?planning-poker) with Zenhub @bess @hectorcorrea @jrgriffiniii

PR #1414, opened 2023-08-10
Updating the database to store if a user enables or disables group messaging

carolyncole:
This allows us to default to true if nothing is found. 

Also added the sub-community field to the database so we can turn the group options on and off for a sub-community too in a later PR

refs #1383

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -3,12 +3,8 @@
 class Group < ApplicationRecord
   resourcify
 
-  def self.option_class
-    GroupOption
-  end
-
   # GroupOptions model extensible options set for Groups and Users
-  has_many :group_options, dependent: :destroy
+  has_many :group_options, dependent: :destroy, class_name: "GroupOption"
   has_many :group_messaging_options, -> { where(option_type: GroupOption::EMAIL_MESSAGES) }, class_name: "GroupOption", dependent: :destroy
 
   has_many :users_with_options, through: :group_options, source: :user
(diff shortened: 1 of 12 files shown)


app/models/group.rb
```diff
@@ -46,28 +42,32 @@ def submitters
...
-
-    found.present?
+    group_option = group_messaging_options.find_by(user: user, group: self)
+    group_option ||= GroupOption.new(enabled: true)
+    group_option.enabled
   end
 
   def self.create_defaults
-    return if count > 0
+    return if where(code: "RD").count > 0
```
carolyncole, line 70: This is a bug that I found when the test creates a group prior to the defaults being created they never get created

app/models/group.rb
```diff
@@ -155,21 +155,13 @@ def delete_permission(current_user, removed_user)
...
-      values << "Princeton Neuroscience Institute"
-      values << "Department of Geosciences"
-      values << "Mechanical and Aerospace Engineering"
-      values << "Astrophysical Sciences"
-      values << "Civil and Environmental Engineering"
-      values << "Chemical and Biological Engineering"
-      values << "Digital Humanities"
-      values << "Music and Arts"
-      values << "Princeton School of Public and International Affairs"
+      ["Princeton Neuroscience Institute", "Department of Geosciences", "Mechanical and Aerospace Engineering",
```
carolyncole, line 161: This was just to make rubocop happy about the class length.  I believe it is still readable.

app/models/user.rb
```diff
@@ -222,36 +222,13 @@ def pending_notifications_count
...
-    enable_messages_from(group: default_group)
+    default_group.enable_messages_for(user: self)
   end
 
   # Returns true if the user has notification e-mails enabled
   # @return [Boolean]
   def email_messages_enabled?
     email_messages_enabled
   end
-
```
carolyncole, line 233: We have the logic duplicated, which I only found out because the tests failed.  Since User is so large I though utilizing the code in Group made sense.

app/models/group.rb
```diff
@@ -46,28 +42,32 @@ def submitters
   # Permit a User to receive notification messages for members of this Group
   # @param user [User]
   def enable_messages_for(user:)
-    raise(ArgumentError, "User #{user.uid} is not an administrator for this group #{title}") unless user.can_admin?(self)
-    group_messaging_options << self.class.option_class.new(option_type: self.class.option_class::EMAIL_MESSAGES, group: self, user: user)
+    raise(ArgumentError, "User #{user.uid} is not an administrator or submitter for this group #{title}") unless user.can_admin?(self) || user.can_submit?(self)
+    group = GroupOption.find_or_initialize_by(option_type: GroupOption::EMAIL_MESSAGES, user: user, group: self)
```
jrgriffiniii, line 46: 👍

jrgriffiniii (approved):

PR #1426, opened 2023-08-11
Ensuring that the S3 file object checksum is updated from AWS within DSpaceBitstreamCopyJob

### LINKED ISSUE #1413: Multipart Uploads checksum should be updated to what AWS returns in the migration snapshot
When we upload items to AWS that are larger than 5GB the etag is no longer a checksum, but a cecksum of the checksums of all the parts.  This means that the checksum of the whole file we store in the MigrationUploadSnapshot is not the same as what AWS returns to us at the end of the upload.

When we update the MigrationUploadSnapshot [here](https://github.com/pulibrary/pdc_describe/blob/main/app/jobs/dspace_bitstream_copy_job.rb#L29) we should update the checksum also.

What AWS returns when queried for the etag
```
 {"checksum"=>"87ade20a34dd1f99b1e37f4e754c0044-2",
    "filename"=>"10.34770/qyrs-vg25/50/2_23_2017_2.archive.tar.gz"},
```

What we calculate and get from dspace
```
{"checksum"=>"UtfdHmVieJjXvmUz/UTwAw==",
    "filename"=>"10.34770/qyrs-vg25/50/2_23_2017_2.archive.tar.gz",
    "migrate_status"=>"complete"},
```

See https://pdc-describe-prod.princeton.edu/describe/works/50 for an example of this happening
Screenshot 2023-08-10 at 8 31 32 AM

jrgriffiniii:
Resolves #1413

### DIFF
--- a/app/jobs/dspace_bitstream_copy_job.rb
+++ b/app/jobs/dspace_bitstream_copy_job.rb
@@ -26,6 +26,7 @@ class DspaceBitstreamCopyJob < ApplicationJob
         if result[:error].present?
           migration_snapshot.mark_error(result[:file], result[:error])
         else
+          # update the checksum here
           migration_snapshot.mark_complete(result[:file])
         end
       end
--- a/app/models/migration_upload_snapshot.rb
+++ b/app/models/migration_upload_snapshot.rb
@@ -18,6 +18,13 @@ class MigrationUploadSnapshot < UploadSnapshot
     if index.present?
       files[index]["migrate_status"] = "complete"
       files[index].delete("migrate_error")
+
+      # Retrieve the checksum from AWS, as this often differs from what is migrated from DSpace
+      # Please see https://github.com/pulibrary/pdc_describe/issues/1413
+      updated = s3_file.s3_query_service.get_s3_object_attributes(key: s3_file.filename)
+      updated_checksum = updated[:etag]
+
+      files[index]["checksum"] = updated_checksum
       finalize_migration if migration_complete?
     end
   end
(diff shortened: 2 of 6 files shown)


app/jobs/dspace_bitstream_copy_job.rb
```diff
@@ -26,6 +26,7 @@ def migrate_file(dspace_file, migration_snapshot_id)
         if result[:error].present?
           migration_snapshot.mark_error(result[:file], result[:error])
         else
+          # update the checksum here
```
carolyncole, line 29: Looks like this comment is unneeded
```suggestion
```

app/models/migration_upload_snapshot.rb
```diff
@@ -18,6 +18,13 @@ def mark_complete(s3_file)
     if index.present?
       files[index]["migrate_status"] = "complete"
       files[index].delete("migrate_error")
+
```
carolyncole, line 21: At some point we may want to change this to utilize the last_response from s3_query_service, but this works, so lets go with it.

carolyncole (approved):

app/models/migration_upload_snapshot.rb
```diff
@@ -18,6 +18,13 @@ def mark_complete(s3_file)
...
+      # Retrieve the checksum from AWS, as this often differs from what is migrated from DSpace
+      # Please see https://github.com/pulibrary/pdc_describe/issues/1413
+      updated = s3_file.s3_query_service.get_s3_object_attributes(key: s3_file.filename)
```
carolyncole, line 24: You could pass in the returned etag instead of querying again.  That would fix the failing tests

```
   migration_snapshot.mark_complete(result[:file],result[:etag].delete('"'))
```

carolyncole (changes requested):

carolyncole (approved):
Since it is green, lets go with it!

PR #1435, opened 2023-08-15
Adding ApprovedUploadSnapshot so we can show the files moving from pre-curation to post-curation

### LINKED ISSUE #1417: How should the provenance log look for files moved from pre-curation to post-curation?
See https://pdc-describe-prod.princeton.edu/describe/works/64

It shows 25 files deleted and 25 files added, which is a bit confusing...

carolyncole:
fixes #1417

New display.  We did not yet get any feedback, but this is better than 15 files deleted and 15 files added with no explaination.

Screenshot 2023-08-16 at 8 04 24 AM

### DIFF
--- a/app/jobs/approved_file_move_job.rb
+++ b/app/jobs/approved_file_move_job.rb
@@ -2,8 +2,9 @@
 class ApprovedFileMoveJob < ApplicationJob
   queue_as :default
 
-  def perform(work_id:, source_bucket:, source_key:, target_bucket:, target_key:, size:)
+  def perform(work_id:, source_bucket:, source_key:, target_bucket:, target_key:, size:, snapshot_id:)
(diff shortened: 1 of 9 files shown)


app/jobs/approved_file_move_job.rb
```diff
@@ -15,6 +16,15 @@ def perform(work_id:, source_bucket:, source_key:, target_bucket:, target_key:,
     unless status
       raise "File check was not valid #{source_key} to #{target_bucket}/#{target_key} Response #{status.to_json}"
     end
+    etag = if resp.respond_to? :copy_object_result
```
jrgriffiniii, line 19: I am assuming that this is limitation of the S3 Client?

jrgriffiniii (approved):

PR #1436, opened 2023-08-15
Strip https://... from DOI

### LINKED ISSUE #1429: Save DOI without URL
When entering a DOI in the Additional tab the system should strip the `https://doi.org/` portion so that `https://doi.org/10.34770/8kyd-5v18` is stored as `10.34770/8kyd-5v18`. The ARK field already does the equivalent functionality.

Screenshot 2023-08-14 at 1 55 16 PM

hectorcorrea:
Strip `https://...` from DOI input textbox, like we do for the ARK

Closes #1429

### DIFF
--- a/app/views/works/_required_form.html.erb
+++ b/app/views/works/_required_form.html.erb
@@ -8,7 +8,7 @@
   <% if @wizard_mode || !doi_mutable? %>
     <input id="doi_read_only" name="doi_read_only" value="<%= @work.resource.doi %>" readonly/>
   <% else %>
-    <p id="doi">DOI: <%= @work.resource.doi %></p>
+    <p id="doi_text">DOI: <%= @work.resource.doi %></p>
   <% end %>
 </div>
 
--- a/spec/system/authz_admin_spec.rb
+++ b/spec/system/authz_admin_spec.rb
@@ -26,7 +26,7 @@ RSpec.describe "Authz for curators", type: :system, js: true do
         login_as research_data_moderator
         visit edit_work_path(work)
         expect(page).to have_content("Editing Dataset")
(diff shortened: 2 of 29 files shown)


app/views/works/_required_form.html.erb
```diff
@@ -8,7 +8,7 @@
   <% if @wizard_mode || !doi_mutable? %>
     <input id="doi_read_only" name="doi_read_only" value="<%= @work.resource.doi %>" readonly/>
   <% else %>
-    <p id="doi">DOI: <%= @work.resource.doi %></p>
+    <p id="doi_text">DOI: <%= @work.resource.doi %></p>
```
hectorcorrea, line 11: We should not have two HTML elements with the same id, so I made the paragraph have a different id so that the JavaScript applies to the input box, not to the paragraph.

carolyncole (approved):

carolyncole:
@hectorcorrea the tests are failing...

spec/system/authz_admin_spec.rb
```diff
@@ -26,7 +26,7 @@
         login_as research_data_moderator
         visit edit_work_path(work)
         expect(page).to have_content("Editing Dataset")
-        fill_in "title_main", with: "New Title", fill_options: { clear: :backspace }
+        fill_in "title_main", with: "New Title", fill_options: { clear: :true }
```
hectorcorrea, line 29: This seems to be needed now that we are using `selenium` instead of `selenium-chrome' (see https://github.com/pulibrary/pdc_describe/pull/1441)

hectorcorrea:
@carolyncole thanks for the heads, one failing test was my bad the other was due to the new selenium configuration.

carolyncole (approved):

PR #1444, opened 2023-08-16
Adding group to the Migrate Dataset menu link

### LINKED ISSUE #1439: Improve workflow for PPPL subcommunities
As a staff member with permission to deposit in both Research Data and PPPL Groups, I should be able to select the PPPL group early in the workflow so that I have access to the PPPL subcommunities the first time I'm describing a record. Instead, the current workflow requires me to migrate or save the record first, then go back to select the PPPL subcommunities. 

## Acceptance criteria
- [ ] Add a menu item for "Migrate PPPL Dataset" that sets the Group to PPPL from the start
- [ ] There is also a menu item for "Migrate PRDS Dataset"

carolyncole:
This allows the migrator to determine which group the item is migrating to before starting the form. Then there is not a need to change the group in the form because it is already set.

fixes #1439

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -32,7 +32,8 @@ class WorksController < ApplicationController
 
   # Renders the "step 0" information page before creating a new dataset
   def new
-    @work = Work.new(created_by_user_id: current_user.id, group: current_user.default_group)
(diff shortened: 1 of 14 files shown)


bess (approved):
🚀

PR #1447, opened 2023-08-16
Don't mark as backdated a note added today

### LINKED ISSUE #1430: Add provenance note - backdate
When users enter a provenance note that has today's date, the system should not mark it as "backdated"

Screenshot 2023-08-14 at 1 59 06 PM

Notice that the system does the correct thing if the user does not enter a date

hectorcorrea:
Closes #1430

### DIFF
--- a/app/models/work_activity.rb
+++ b/app/models/work_activity.rb
@@ -118,6 +118,7 @@ class WorkActivity < ApplicationRecord
 
     UNKNOWN_USER = "Unknown user outside the system"
     DATE_TIME_FORMAT = "%B %d, %Y %H:%M"
+    DATE_FORMAT = "%B %d, %Y"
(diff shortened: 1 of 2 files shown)


app/models/work_activity.rb
```diff
@@ -137,7 +138,9 @@ def created_sortable_html
     def created_updated_html
       created = @work_activity.created_at.time.strftime(DATE_TIME_FORMAT)
       updated = @work_activity.updated_at.time.strftime(DATE_TIME_FORMAT)
-      if created == updated
+      created_date = @work_activity.created_at.time.strftime(DATE_FORMAT)
+      updated_date = @work_activity.updated_at.time.strftime(DATE_FORMAT)
+      if created_date == updated_date
```
hectorcorrea, line 143: Compare dates instead of comparing timestamps

bess (approved):

Issue #1424, opened 2023-08-11
Describe Form: Change PPPL's Default Publisher


hhkitsune:
# Description

In DataSpace, PPPL's publisher was: Princeton Plasma Physics Laboratory, Princeton University. (See: [sample record](https://dataspace.princeton.edu/handle/88435/dsp01ft848t262)) The default in PDC Describe is: Princeton University. We would like to change the default for the PPPL Group pipeline only. All submissions to the PPPL Group pipline should have this new default. All records from the PRDS Group pipeline should remain with the default: Princeton University.

## PPPL Default Publisher
Princeton Plasma Physics Laboratory, Princeton University

## Acceptance Criteria
- [x] The default publisher will change for the PPPL Group only (Please see above for the new default publisher.)
- [x] The default publisher for PRDS Group will remain unchanged.

hectorcorrea:
With PR #1453 the publisher defaults correctly:
* for records migrated via the "Migrate PPPL Dataset" button and 
* for new submissions created by users with only access to the PPPL group

New submissions for users with access to *more than one group* will still have "Princeton University" as the default publisher.

PR #1453, opened 2023-08-17
Default publisher for PPPL records

hectorcorrea:
Defaults the publisher for PPPL *migrated* records to "Princeton Plasma Physics Laboratory"

Moves #1424 forward.


app/models/group.rb
```diff
@@ -186,7 +187,16 @@ def subcommunities
   end
   # rubocop:enable Metrics/MethodLength
 
+  def publisher
```
jrgriffiniii, line 190: 👍

jrgriffiniii (approved):

astrochun:
Actually I just noticed that this wasn't "Princeton Plasma Physics Laboratory, Princeton University" as we did in DataSpace. Was there a reason @hectorcorrea?

Issue #1496, opened 2023-09-05
Change Plasma Science and Technology sub-community to "Discovery Plasma Science"


astrochun:
Because of some reorganization, PS&T is now DPS. To avoid confusion to researchers, this sub-community should be renamed. We should also make sure this is reflected in the metadata for existing records.

- [x] Changed PS&T to "Discovery Plasma Science"
- [x] Check that meta data for DPS is correct for existing records

## Moved to a separate ticket:
Also, we need to create a new sub-community called "Applied Materials and Sustainability Sciences"
- [ ] ~[Create "Applied Materials and Sustainability Sciences"](https://github.com/pulibrary/pdc_describe/issues/1498)~

bess:
Output of job in production:

> Updating work 85
> Updating work 48
> Updating work 45
> Updating work 66
> Updating work 40
> Updating work 77
> Updating work 64
> Updating work 84
> Updating work 103
> Updated subcommunities metadata, recorded user bs3097 in WorkActivity

![Screenshot 2023-09-13 at 8 30 57 AM](https://github.com/pulibrary/pdc_describe/assets/65608/49c8c1be-43f2-43d0-ab97-f54b8de99ebe)

PR #1505, opened 2023-09-11
PPPL Department name change from PS&T to Discovery Plasma Science

bess:
Ref #1496 

Screenshot 2023-09-12 at 8 44 20 AM
Screenshot 2023-09-12 at 8 40 43 AM
Screenshot 2023-09-12 at 8 39 20 AM

### DIFF
--- a/lib/tasks/metadata.rake
+++ b/lib/tasks/metadata.rake
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+require "byebug"
+
+namespace :metadata do
+  # command line syntax: bundle exec rake metadata:update_pppl_subcommunities\["netid"\]
+  desc "Update renamed PPPL subcommunities"
+  task :update_pppl_subcommunities, [:netid] => [:environment] do |_, args|
+    netid = args[:netid]
+    user = User.find_by(uid: netid)
+    raise("No user found for id #{netid}") unless user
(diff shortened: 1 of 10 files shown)


lib/tasks/metadata.rake
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require "byebug"
```
carolyncole, line 3: I'm guessing this is not needed
```suggestion
```

carolyncole (dismissed):
Small update if you want to, but looks good to me otherwise...

bess:
@astrochun When we run the rake task (documented here: https://github.com/pulibrary/pdc_describe/blob/70da0f069b120110f3ccfaccf39837c6b538ac11/docs/bulk_metadata_updates.md) that will update all of the records in the system, regardless of whether they are in review or have already been published.

astrochun:
> @astrochun When we run the rake task (documented here: https://github.com/pulibrary/pdc_describe/blob/70da0f069b120110f3ccfaccf39837c6b538ac11/docs/bulk_metadata_updates.md) that will update all of the records in the system, regardless of whether they are in review or have already been published.

Great!

lib/tasks/metadata.rake
bess: 👍🏻 Fixed

carolyncole (approved):

Issue #1491, opened 2023-09-05
Embargoes for files


kelynch:
Enable objects in PDC Describe to have their descriptive metadata indexed into PDC Discovery when approved, but not their files if those files are embargoed.  File embargoes at the file level are handled and enforced by curators, with embargoed files stored somewhere separate from PDC Describe infrastructure/storage while they are embargoed.  

Objects in PDC Describe would still need to have a README associated with it in order to be approved, but all files to be embargoed are stored outside of PDC Describe, and then after the embargo period is done, they are attached to the object via PDC Describe/Globus manually by the curator. 

## Potential method: 
- [ ] Add a date field to "embargo files" on an object in PDC Describe during curation (in "Curator Controlled" metadata)
- [ ] Update PDC Discovery to skip indexing on all files attached to an object when the object's metadata has an embargo date set with a value that is a date in the future.  PDC Discovery should index files on objects that have either no embargo date set, or an embargo date set to the past or the current day.
- [ ] Display a message in PDC Discovery to indicate that files are embargoed, including the embargo date, eg "File(s) associated with this object are embargoed until 2024/01/01."
- [ ] Write an end-to-end system spec in PDC Describe demonstrating what it looks like to create an item with an embargo, so we can ensure all parts of the system continue to work together

jrgriffiniii:
I am nearly complete with extending the `Work` model to support embargo dates, however, my only delay has been some ongoing troubleshooting with Sass preprocessing for improving the styling for the form:

![image](https://github.com/pulibrary/pdc_describe/assets/1443986/ef0dfea6-90a6-4b58-a203-9fdb67581517)

jrgriffiniii:
After the undertaking of testing within the `staging` environments for PDC Describe and PDC Discovery, the following two pull requests are all which remain from fully-resolving this issue:

- https://github.com/pulibrary/pdc_discovery/pull/512
- https://github.com/pulibrary/pdc_describe/pull/1544

PR #1511, opened 2023-09-12
Extending the Work model and form to support setting embargo dates

jrgriffiniii:
Advances #1491. Additional pull requests shall be required in order to ensure the following:

- Attempting to access Works under active embargo without proper authorization will result in a 301 redirect response for unauthenticated users
- Attempting to access Works under active embargo without proper authorization will result in a 403 error for authenticated users
- Implementing support for lifting embargoes synchronously

### DIFF
--- a/spec/requests/works_spec.rb
+++ b/spec/requests/works_spec.rb
@@ -65,6 +65,11 @@ RSpec.describe "/works", type: :request do
         stub_s3
       end
 
+      context "when an embargo date is specified" do
+        it "updates the embargo date" do
+        end
+      end
+
       context "when files are added to the Work" do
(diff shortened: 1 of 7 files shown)


spec/requests/works_spec.rb
```diff
@@ -65,6 +65,11 @@
         stub_s3
       end
 
+      context "when an embargo date is specified" do
+        it "updates the embargo date" do
```
bess, line 69: There's no test here?

bess (changes requested):
Looks like there's an empty test? Was that just left in by accident perhaps?

jrgriffiniii:
I am very sorry, I have fixed this and implemented authorization support with https://github.com/pulibrary/pdc_describe/pull/1515

PR #1515, opened 2023-09-13
Extending the Work model to support embargoes and implementing authorization policies

jrgriffiniii:
Advances #1491 and deprecates https://github.com/pulibrary/pdc_describe/pull/1511

### DIFF
--- a/app/services/form_to_resource_service.rb
+++ b/app/services/form_to_resource_service.rb
@@ -7,6 +7,9 @@ class FormToResourceService
     #  @param [Work] work params will be applied to. Utilizes the work for old values if needed.
     #
     # @return [PDCMetadata::Resource] Fully formed resource containing updates from the user
+    #
+    # rubocop:disable Metrics/MethodLength
+    # rubocop:disable Metrics/AbcSize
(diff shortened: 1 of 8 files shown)


app/services/form_to_resource_service.rb
```diff
@@ -7,6 +7,9 @@ class << self
     #  @param [Work] work params will be applied to. Utilizes the work for old values if needed.
     #
     # @return [PDCMetadata::Resource] Fully formed resource containing updates from the user
+    #
```
carolyncole, line 10: @jrgriffiniii there is no change to the method, why are these new disables required?

carolyncole (changes requested):
Looks good.  I'm unsure why additional disables of rubocop were required for a method that did not change.  I'm happy to approve once those are removed!

app/services/form_to_resource_service.rb
jrgriffiniii: I am very sorry, these must have been tracked from an earlier commit. I have removed these.

carolyncole (approved):

PR #1540, opened 2023-09-26
Upgrade rubocop rules

bess:


carolyncole (approved):

app/models/ark.rb
```diff
@@ -60,7 +60,7 @@ def self.valid?(ezid)
   end
 
   def self.valid_shoulder?(ezid)
-    !ezid.include?(self::EZID_TEST_SHOULDER)
+    ezid.exclude?(self::EZID_TEST_SHOULDER)
```
carolyncole, line 63: Super cool!  I've thought ruby should have exclude?...

PR #1543, opened 2023-09-28
 add provennance note test

### LINKED ISSUE #1520: Create a Test for the Provenance Note Edit screen
Step 3 to complete https://github.com/pulibrary/pdc_describe/issues/1336
Make a system spec test file in the spec/system/ called provenance_note_spec.rb
Use the code below as a template 
```
# frozen_string_literal: true
require "rails_helper"

RSpec.describe "Adding a Provenance note", type: :system, js: true do
  let(:work) { FactoryBot.create(:draft_work, created_by_user_id: user) }

  context "A user that does not have permissions to edit the provenance" do
    let(:user) { FactoryBot.create(:user) }
    it "does not show the form" do
      pending "This should test that the page does not have the form"
    end
  end

  context "A user that does has permissions to edit the provenance" do
    let(:user) { FactoryBot.create(:user, uid: config.provenance_message_writers.first) }
    it "does show the form" do
      pending "This should test that the page does have the form"
    end
  end
end
```
Then remove the two pending lines and put in actual tests like `expect(page).not_to have_content(...)` and `expect(page).to have_content(...)`

FavourNwogbo21:
fixes #1520

### DIFF
--- a/spec/system/provenance_note_spec.rb
+++ b/spec/system/provenance_note_spec.rb
@@ -2,7 +2,6 @@
 require "rails_helper"
 
 RSpec.describe "Adding a Provenance note", type: :system, js: true do
-
   context "A user that does not have permissions to edit the provenance" do
     let(:any_user) { FactoryBot.create(:user) }
     let(:work) { FactoryBot.create(:draft_work, created_by_user_id: any_user.id) }
--- a/spec/system/provenance_note_spec.rb
+++ b/spec/system/provenance_note_spec.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.describe "Adding a Provenance note", type: :system, js: true do
+  let(:work) { FactoryBot.create(:draft_work, created_by_user_id: user) }
+
+  context "A user that does not have permissions to edit the provenance" do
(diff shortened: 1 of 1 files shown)


spec/system/provenance_note_spec.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.describe "Adding a Provenance note", type: :system, js: true do
```
jrgriffiniii, line 4: 💯

spec/system/provenance_note_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    it "does not show the form" do
+      login_as any_user
+      visit work_path(work)
+      expect(page).to have_content(work.title)
+      expect(page).not_to have_button("Add Provenance Note")
+      expect(page).not_to have_form(add_provenance_note_path(work), :post)
+    end
+  end
+
+  context "A user that does has permissions to edit the provenance" do
```
jrgriffiniii, line 17: 💯

jrgriffiniii (approved):
Excellent work @FavourNwogbo21! Thank you very much for this!

PR #1549, opened 2023-10-03
Upgrade postgres

bess:
Since we're upgrading to postgres 15 in production, let's match that as closely as possible in dev and test

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -1,5 +1,5 @@
 name: pdc-describe
 services:
   database:
-    type: postgres:13
+    type: postgres:14
     portforward: true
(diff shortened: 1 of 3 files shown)


.lando.yml
```diff
@@ -1,5 +1,5 @@
 name: pdc-describe
 services:
   database:
-    type: postgres:13
+    type: postgres:14
```
carolyncole, line 4: The lates lando release is 14 https://docs.lando.dev/postgres/

carolyncole (commented):
Looks good!  Leaving not to self that 14 was chosen because that is the lates for now...

carolyncole (approved):

Issue #147, opened 2022-05-27
Move webmock to "test"group in gemfile


kelynch:
Webmock is currently in the all environment groups of the gemfile, and should be moved out of production for security.

It is pulled in as part of the [sample_data.rake](https://github.com/pulibrary/pdc_describe/blob/main/lib/tasks/sample_data.rake) file.

bess:
The reason webmock is in the main gem group is to enable the creation of sample data on production. That was a deliberate decision so that we could create our fixture objects on the server for demo and testing purposes. Do we know that webmock actually has a security issue? Or is this a general “don’t put things in production if you don’t actually need them there” concern? I’m really glad to talk about different choices we might make, but I’d like to suggest that the ability to easily create our fixture objects on the server is a valuable feature, especially during development. Might we compromise and say we’ll remove it from the main group when we make the site available outside VPN?

PR #1559, opened 2023-10-04
Move testing gems to test environment

bess:
These were in production early in the project because we were using them to generate test data. No longer needed except in test.

Fixes https://github.com/pulibrary/pdc_describe/issues/147


carolyncole (approved):
Thank you!

Issue #1580, opened 2023-10-20
Zero bytes files that are not migrated from pre-curation to post-curation should be noted in the Provenance Log


hectorcorrea:
Anya reported that this work https://datacommons.princeton.edu/describe/works/139 lost its files once it was approved. The problem seems that the files were not moved from pre-curation to post-curation. The files are still in the pre-curation bucket in AWS:

![wortk_139](https://github.com/pulibrary/pdc_describe/assets/568286/5cb3906e-1eee-48f5-8a01-25cf1042eb44)

Is it possible that the zero byte README is causing this problem?

This issue might be related to https://github.com/pulibrary/pdc_describe/issues/1574

hectorcorrea:
The job to move the files had failed. Retrying the job manually (as documented in #1574) fixed the problem with the `ARK_DATA.zip` file, but not with the `README.txt` presumably because that file is zero bytes.

carolyncole:
I requested some clarification on this in [slack](https://pulibrary.slack.com/archives/C01555TP6HE/p1698333567439799)

carolyncole:
From the discussion on slack (see above comment) we should be adding the zero length files not moving into the provenance log and not move the files.

carolyncole:
Doing a bit of digging.  Directories are zero bytes and have a file name ending with `/` (`10.34770/zztk-f783/35/bitKlavierGrand_Binaural_44k16b/`) Files that are zero bytes do not have a filename that end with `/`(`10.11578/1562101/137/README.txt`)

PR #1589, opened 2023-10-26
Adding a method to discover empty files to the s3query service

carolyncole:
refs #1580

Should allow us to see if there are any empty files when we transfer files on approval.


spec/services/s3_query_service_spec.rb
```diff
@@ -709,6 +709,38 @@
     end
   end
 
+  describe "#client_s3_empyt_files" do
```
bess, line 712:
```suggestion
  describe "#client_s3_empty_files" do
```

bess (approved):

bess (approved):

PR #1593, opened 2023-11-01
Use build instead of create to avoid ARK collision

### LINKED ISSUE #1592: Order dependent test failure
Running the "services" spec folder by itself (`rspec spec/services`) results in a failure.

Screenshot 2023-11-01 at 1 56 45 PM

bess:
This will also make the test run faster because it won't write to the database

Fixes #1592


carolyncole (approved):
Should also be faster if we do not need to save them

Issue #1613, opened 2023-11-21
Handle the Individual Contributor not having a type (role in UI) better and validate the datacite before apporval


carolyncole:
To recreate this generate a work that has an Additional Individual Contributor with no role filled in, then try to approve the work. 
![Screenshot 2023-11-21 at 1 42 17 PM](https://github.com/pulibrary/pdc_describe/assets/1599081/e2a3fb6c-dde7-4024-81fe-51c2e384336e)

The error message is very hard to understand.  Datacite rejects the update with
```
/opt/pdc_describe/shared/bundle/ruby/3.1.0/gems/datacite-mapping-0.4.1/lib/datacite/mapping/contributor.rb:120:in `type=': Type cannot be nil (ArgumentError)                                                                            
```

That bubbles up to the controller and gets shown to the user as `"We apologize, an error was encountered: Type cannot be nil (ArgumentError). Please contact the PDC Describe administrators.  

https://datacommons.princeton.edu/describe/works/203/datacite/validate - currently errors becuase there is no type
https://datacommons.princeton.edu/describe/works/162/datacite/validate does not error completly, but shows an issue
![Screenshot 2023-11-21 at 2 20 46 PM](https://github.com/pulibrary/pdc_describe/assets/1599081/6dc03610-f9be-44da-ac2c-c3f7e0f75fd9)

We should be running the XML validation on the datacite record as part of our pre approvals validation
![Screenshot 2023-11-21 at 2 22 31 PM](https://github.com/pulibrary/pdc_describe/assets/1599081/46083205-5e2b-4fb0-b6c8-70141d2cc701)

PR #1615, opened 2023-11-27
Updated the works controller to utilize DataCiteSerialization for XML validation

carolyncole:
Also capture exceptions during validation

refs #1613

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -267,14 +267,14 @@ class WorksController < ApplicationController
   def datacite_validate
     @errors = []
     @work = Work.find(params[:id])
-    datacite_xml = Nokogiri::XML(@work.to_xml)
-    schema_location = Rails.root.join("config", "schema")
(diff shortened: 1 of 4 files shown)


app/controllers/works_controller.rb
```diff
@@ -267,14 +267,14 @@ def datacite
...
-    datacite_xml = Nokogiri::XML(@work.to_xml)
-    schema_location = Rails.root.join("config", "schema")
-    Dir.chdir(schema_location) do
-      xsd = Nokogiri::XML::Schema(File.read("datacite_4_4.xsd"))
-      xsd.validate(datacite_xml).each do |error|
-        @errors << error
-      end
-    end
+    datacite_serialization = work.resource.datacite_serialization
+    datacite_serialization.valid?
```
hectorcorrea, line 271: Nice, I am glad you found that the `Datacite` class has a way to execute this without us having to duplicate the code.

hectorcorrea (approved):

PR #1634, opened 2023-12-18
Allow empty row in the creator and contributor tables to be ignored

### LINKED ISSUE #1628: ignore empty rows in the creators
If you add empty row to the contributor they get ignored.  Apply the same logic to the creator table.

Oddly when looking at the code I do not see the logic being applied to skip empty rows, but the code to check for an empty row is here: https://github.com/pulibrary/pdc_describe/blob/main/app/javascript/entrypoints/pdc/edit_required_fields.es6#L70

I wonder if there is old javascript around for the creator table causing issues...

carolyncole:
Also fixes a bug in the test, which is filling in the Family name and the expecting it to also be blank.  This no longer occurs because we are ignoring empty rows, but did occur in the past because the empty row was causing the family name issue.

fixes #1628


leefaisonr (approved):
Looks good, thank you!!

PR #1640, opened 2024-01-04
Total file size

### LINKED ISSUE #1581: Total size by submission
Display on the show page the total size of the submission. We currently display the total number of files (in the DataTables footer) but not the total size in bytes.

We probably also surface this information on the user's dashboard.

This is related to https://github.com/pulibrary/pdc_describe/issues/1541 but not the same.

claudiawulee:
Closes #1581 
image

### DIFF
--- a/spec/system/data_migration/sowingseeds_form_submission_spec.rb
+++ b/spec/system/data_migration/sowingseeds_form_submission_spec.rb
@@ -67,6 +67,7 @@ Download the README.txt for a detailed description of this dataset's content."
       click_on "Additional Metadata"
       click_on "Curator Controlled"
       fill_in "ark", with: ark
+      fill_in "publication_year", with: "2023"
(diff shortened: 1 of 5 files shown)


spec/system/data_migration/sowingseeds_form_submission_spec.rb
```diff
@@ -67,6 +67,7 @@
       click_on "Additional Metadata"
       click_on "Curator Controlled"
       fill_in "ark", with: ark
+      fill_in "publication_year", with: "2023"
```
hectorcorrea, line 70: FYI -- This is unrelated to the PR but the test was failing because the fixture is hard coded to 2023 and the system defaults the publication year to the current year to new submissions.

carolyncole (approved):
Looks great @claudiawulee and @hectorcorrea

PR #1645, opened 2024-01-09
Force README files to sort to the top.

### LINKED ISSUE #1643: Sort README first
When displaying the list of files in the Show page make sure the README shows first when using the default sort order (which is sort by file name). This will be particularly handy for works that have many files since it will make it easier for the researcher to make sure the README is up to date.

We should enforce the same logic in PDC Discovery.

hectorcorrea:
Forces any file with the word "README" to be sorted first.

We also fixed an issue with the sort by size.

Closes #1643

### DIFF
--- a/app/views/works/_s3_resources.html.erb
+++ b/app/views/works/_s3_resources.html.erb
@@ -83,8 +83,10 @@
             if (type == "display") {
               var html;
               if (row.filename_display.startsWith("*")) {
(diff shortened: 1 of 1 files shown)


bess (changes requested):
This looks like a great fix, but I'd really like to see a system test so we ensure this doesn't break in the future.

bess (approved):
🚀

PR #1678, opened 2024-02-27
Showing the full name in the curator drop down select on the work show page

### LINKED ISSUE #1670: Curator list should have full name
Per a request from one of the PPPL data reviewer and user, it is very cumbersome and confusing to use just first names. There are multiple first names now. They said:

> On that subject, it would be helpful if the curators' last names were visible (so that I know which Rob, Nathaniel, etc. they are)

I would agree. We have two Kevin's in there so it's already confusing.

carolyncole:
fixes #1670

Screenshot 2024-02-27 at 2 12 52 PM

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -168,6 +168,13 @@ class User < ApplicationRecord
     given_name.presence || uid
   end
 
+  # Returns a full name that always has a value
+  # This is needed because we have records in the Users table that are created automatically
+  # in which the only value we have for sure its their uid (aka NetID).
+  def full_name_safe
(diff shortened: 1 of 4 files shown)


hectorcorrea:
The tests caught an issue with null values...

```
undefined method `strip' for nil:NilClass
     #   
     #       full_name.strip.presence || uid
     #                ^^^^^^
```

carolyncole:
@hectorcorrea thank goodness for tests.  That also broke in the UI.  I just did not have as many administrators locally.  It should be good to go now...

hectorcorrea (approved):

PR #1692, opened 2024-03-14
Ensuring that the links for file downloads within the Work show and edit views are constructed and encoded for Rails routing

### LINKED ISSUE #1665: Bug: Unable to download files under Edit mode due to incorrect URL
One of the PPPL researchers point out that if you're in "Edit" mode on a record and you click on blue link for a data/file that was uploaded, it returns an error:

View from Edit mode:
Screen Shot 2024-01-31 at 2 18 07 PM

The error page:
Screen Shot 2024-01-31 at 2 19 10 PM

The issue, I think stems from the URI, with duplicate "246" for the record ID:
https://datacommons.princeton.edu/describe/works/246/246/download?filename=10.34770/dq61-gf89/246/Figure10

Note that this issue doesn't happen in the View mode for the record, which does permit downloading the file.

jrgriffiniii:
Resolves #1665

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -23,6 +23,7 @@ class WorksController < ApplicationController
   def new
     group = Group.find_by(code: params[:group_code]) || current_user.default_group
     @work = Work.new(created_by_user_id: current_user.id, group:)
+    @work_decorator = WorkDecorator.new(@work, current_user)
     @form_resource_decorator = FormResourceDecorator.new(@work, current_user)
     if wizard_mode?
       render "new_submission"
@@ -39,6 +40,7 @@ class WorksController < ApplicationController
       upload_service.update_precurated_file_list(added_files_param, deleted_files_param)
       redirect_to work_url(@work), notice: "Work was successfully created."
     else
+      @work_decorator = WorkDecorator.new(@work, current_user)
       @form_resource_decorator = FormResourceDecorator.new(@work, current_user)
       render :new, status: :unprocessable_entity
     end
@@ -96,8 +98,10 @@ class WorksController < ApplicationController
   end
 
   # GET /works/1/edit
+  # rubocop:disable Metrics/MethodLength
   def edit
     @work = Work.find(params[:id])
+    @work_decorator = WorkDecorator.new(@work, current_user)
     if current_user && @work.editable_by?(current_user)
       if @work.approved? && !@work.administered_by?(current_user)
         Honeybadger.notify("Can not edit work: #{@work.id} is approved but #{current_user.uid} is not admin")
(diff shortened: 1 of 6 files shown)


hectorcorrea (dismissed):

app/controllers/works_controller.rb
```diff
@@ -23,6 +23,7 @@ def index
   def new
     group = Group.find_by(code: params[:group_code]) || current_user.default_group
     @work = Work.new(created_by_user_id: current_user.id, group:)
+    @work_decorator = WorkDecorator.new(@work, current_user)
```
hectorcorrea, line 26: It does not seem that we actually using this variable in the method, do we need it?

app/controllers/works_controller.rb
```diff
@@ -39,6 +40,7 @@ def create
       upload_service.update_precurated_file_list(added_files_param, deleted_files_param)
       redirect_to work_url(@work), notice: "Work was successfully created."
     else
+      @work_decorator = WorkDecorator.new(@work, current_user)
```
hectorcorrea, line 43: It does not seem that we actually using this variable in the method, do we need it?

app/controllers/works_controller.rb
```diff
@@ -96,8 +98,10 @@ def resolve_ark
   end
 
   # GET /works/1/edit
+  # rubocop:disable Metrics/MethodLength
   def edit
     @work = Work.find(params[:id])
+    @work_decorator = WorkDecorator.new(@work, current_user)
```
hectorcorrea, line 104: It does not seem that we actually using this variable in the method, do we need it?

app/controllers/works_controller.rb
jrgriffiniii: I am sorry, yes, this should be needed. The tests were breaking for me locally, and I find them also breaking on CircleCI when I attempt to remove this from the `#create` action:
https://app.circleci.com/pipelines/github/pulibrary/pdc_describe/4808/workflows/08304c0b-6d44-4e0b-8f60-c609e6468fd9/jobs/17830

app/views/works/_s3_resources.html.erb
```diff
@@ -34,7 +34,13 @@
 <script type="text/javascript">
   $(function() {
     // work.id is nil when the form is rendered during work creation and the work has not been saved
-    var fileListUrl = '<%= work_file_list_path(@work.id || "NONE") %>';
+    var work = '<%= @work %>';
+    let fileListUrl;
+    if (work) {
+      fileListUrl = '<%= @work_decorator.file_list_path %>';
```
hectorcorrea, line 40: This might be why it is needed, the view seems to reference `@work_decorator`

hectorcorrea (approved):

Issue #1684, opened 2024-03-06
Refine the navigation buttons for the submission wizard


matthewjchandler:
### User Story
As a PDC submitter, I need to have clear and consistent navigation buttons throughout the submission wizard, so that I can be assured that I'm not missing anything and save myself from surprises and hassles through the curation process after I submit.

### Context
PRDS engaged several beta testers after PDC - Describe version 1.0 was released. The testers expressed deep misgivings with the current layout of the submission workflow, with particular confusion over navigating through the wizard. We can address the concerns by refining the action buttons at the bottom of each page/section of the submission wizard.

In particular, I propose a standard four-button layout: "Cancel" at the lower-left; and at the lower-right, "Previous", "Save", and "Next" (in that order). We would naturally deviate from the standard on the first and last steps.

So then the "Required Metadata" portion of the Describe step would keep the existing "Cancel" button in the lower-left, but revise the lower-right with "Save" and "Next" buttons:

![PDC-Describe_NewSubmission-Describe_ActionButtons_2024-03-06](https://github.com/pulibrary/pdc_describe/assets/66128387/1600c19e-5c56-4eba-9801-9e6824ce30d1)

Another repeated complaint is that submitters don't realize there's any opportunity to enter required metadata (addressed in part by #1635), which is a problem for curators. We can address those issues on the first page of the Describe step by making the (new) "Next" button take the submitter to the "Additional Metadata" portion of the Describe step.

The additional metadata page can keep the same body contents, but revise the action buttons at the bottom to match the new consistent layout:

![PDC-Describe_NewSubmission-Describe-Addl_ActionButtons_2024-03-06](https://github.com/pulibrary/pdc_describe/assets/66128387/37d342bb-1d82-4a54-9e25-a16ec7eef59a)

At the Readme and Upload steps, we not only need to revise the lower-right buttons, but also the existing "Go Back" buttons to become "Cancel". The Cancel button should have the same action on every page: exiting the wizard without saving whatever has changed since the last save.

![PDC-Describe_NewSubmission-Readme_ActionButtons_2024-03-06](https://github.com/pulibrary/pdc_describe/assets/66128387/0a40db31-2db6-477a-9807-d481b45e068b)

![PDC-Describe_NewSubmission-Upload_ActionButtons_2024-03-06](https://github.com/pulibrary/pdc_describe/assets/66128387/026e3043-a8f6-489e-ae9d-fd4dbec5c155)

[...]

PR #1710, opened 2024-03-22
Separating wizard new & edit and non wizard new & edit controller methods

carolyncole:
refs #1684
Also adding documentation to the top of the file for which methods get called when.
The goal is to move the wizard to a new controller so that we can more easily update it...

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -3,10 +3,33 @@
 require "nokogiri"
 require "open-uri"
 
+# Currently this controller supports Multiple ways to create a work, wizard mode, create dataset, and migrate
+# The goal is to eventually break some of these workflows into separate contorllers.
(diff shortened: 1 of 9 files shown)


app/controllers/works_controller.rb
```diff
@@ -532,5 +569,22 @@ def migrated?
 
       params[:submit] == "Migrate"
     end
+
+    # @returns false if an error occured
+    def handle_modification_permissions(uneditiable_message: "Can not edit work: #{@work.id} is not editable by #{current_user.uid}",
```
hectorcorrea, line 574: There is a typo on `uneditiable_message` but it's consistent so it works

hectorcorrea (approved):

app/controllers/works_controller.rb
carolyncole: Thanks I'll fix the typo in my next one...

PR #1715, opened 2024-03-26
Split out the work wizard into its own controller

carolyncole:
refs #1684


hectorcorrea (approved):
Fantastic job! and the code looks much more logical.

PR #1719, opened 2024-03-26
Uses Uppy for uploads in the Work edit screen

### LINKED ISSUE #1685: SPIKE: Redesign the New Submission Upload page
### User Story
As a PDC submitter, I need clear instructions for submitting my data files, with assurances that I am proceeding as I am supposed to, so that I can avoid a confusing and/or time-consuming conversation with the curators.

### Context

Multiple PDC testers criticized the file upload workflow: it gives no options for uploading files from multiple local sources; it does not spell out which files have been selected when there are multiple; when one tries to make a second selection, it just overwrites the first selection; and it does give a confirmation of successful upload.

PDC-Describe_NewSubmission-Upload_SelectedFiles_2024-03-06

Moreover, in PDC, after one advances to the Review step and then returns to the Upload, it appears that the uploaded files are removed. This really threw our testers through a loop.

They expected something more like what DataSpace provides: options of multiple uploads on one page, with immediate information on progress and success, and then a way to make revisions to the files after upload (including anything uploaded by mistake).

DataSpace_Upload_2024-03-07
DataSpace_Upload-Progress_2024-03-07
DataSpace_Upload-Success-Multi_2024-03-07
DataSpace_Upload-Revise_2024-03-07

Back to PDC. I'd recommend we build on the file viewer that exists in the curator/moderator view:

PDC-Describe_ModeratorReview-Files_2024-03-07

After selecting and uploading, submitters need to see what they have there and decide whether they are missing something or included something by mistake. This is a strong point made by multiple testers.

### Criteria for the solution
- [ ] On the "file-upload" page, add the following functionality:
  - [ ] The ability to select multiple files from multiple local sources and in multiple selection operations
  - [ ] Immediate confirmation with a list of all files successfully uploaded
  - [ ] The ability to remove some files after upload

## Criteria for this ticket
- [x] Hector is going to investigate what UI options are available to us and report back. 
- [x] Schedule a meeting with RDSS + Matt Chandler
- [x] Spend 1 hour discussing options and make tickets

hectorcorrea:
Integrates Uppy to handle file uploads in the Work edit screen (i.e. not in the Wizard)

Screenshot 2024-04-02 at 4 50 22 PM

Closes #1685 forward

Notice that Uppy is only used if we pass the `new_uploader=true` as a query string parameter. I am going to use this to demo the functionality to Matt.

There are a couple of additional tickets to implement this in the Wizard too: https://github.com/pulibrary/pdc_describe/issues/1731 and https://github.com/pulibrary/pdc_describe/issues/1732

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -225,6 +225,9 @@ class WorksController < ApplicationController
 
   def upload_files
     byebug
+    upload_service = WorkUploadsEditService.new(@work, current_user)
+    files = params["files"]
+    upload_service.update_precurated_file_list(files, [])
   end
 
   private
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -119,7 +119,7 @@ class WorksController < ApplicationController
   # GET /works/1/edit
   # only non wizard mode
   def edit
-    @new_uploader = true
+    @new_uploader = params[:new_uploader] == "true"
     @work = Work.find(params[:id])
     @work_decorator = WorkDecorator.new(@work, current_user)
     if handle_modification_permissions
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -119,6 +119,7 @@ class WorksController < ApplicationController
   # GET /works/1/edit
   # only non wizard mode
   def edit
+    @new_uploader = true
     @work = Work.find(params[:id])
     @work_decorator = WorkDecorator.new(@work, current_user)
     if handle_modification_permissions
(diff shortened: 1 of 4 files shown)


app/controllers/works_controller.rb
```diff
@@ -108,6 +108,8 @@ def resolve_ark
   # GET /works/1/edit
   # only non wizard mode
   def edit
+    @new_uploader = params[:new_uploader] == "true" || Rails.env.staging?
```
hectorcorrea, line 111: Notice that with this PR the new file uploader is only available in staging (you can force it with a query string `?new_uploader=true` if you want to test locally)

carolyncole (approved):
This is just spikecode, so the fact that the javascript is inline and there are no tests should be ok.  As this goes into production we should have tests and the javascript should live in app/javascript.

PR #1720, opened 2024-03-27
Updating works (wizard and normal) controllers to share code

carolyncole:
also clean up the wizard controller a bit to remove rubocop vialation


bess (approved):
This looks great. I added one line to notify honeybadger in the case of an error parsing embargo dates.

PR #1727, opened 2024-04-01
Allowing the new submission form to save and stay on page

carolyncole:
This was more complex than expected, because some times the work is persisted and sometimes not now that we can stay on the page.

For this reason I extracted some code into the WorkMetadataService.

The new button:
Screenshot 2024-04-01 at 11 23 51 AM

#1684

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -30,7 +30,8 @@ Rails.application.routes.draw do
 
   # The work wizard
   get "works/new-submission", to: "works_wizard#new_submission", as: :work_create_new_submission
-  post "works/new-submission", to: "works_wizard#new_submission_save", as: :work_new_submission
+  post "works/new-submission/(:id)", to: "works_wizard#new_submission_save", as: :work_new_submission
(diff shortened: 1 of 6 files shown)


config/routes.rb
```diff
@@ -30,7 +30,8 @@
 
   # The work wizard
   get "works/new-submission", to: "works_wizard#new_submission", as: :work_create_new_submission
-  post "works/new-submission", to: "works_wizard#new_submission_save", as: :work_new_submission
+  post "works/new-submission/(:id)", to: "works_wizard#new_submission_save", as: :work_new_submission
```
hectorcorrea, line 33: Curious, what do the parenthesis do?

hectorcorrea (approved):

config/routes.rb
carolyncole: Patch is what is sent for an update, and (:id) allows the id to be present or not

carolyncole: @hectorcorrea This allows for the save to occur for a brand new work, or one that was created because they hit save and returned to the page.

PR #1728, opened 2024-04-01
ensuring that the wizards buttons travers in the same order forwards and backwards

JaymeeH:
- testing the wizard in reverse order
- adding additional buttons to the work wizard

refs #1684


leefaisonr (approved):
Thanks!

PR #1741, opened 2024-04-04
Display confirm dialog before approving a work

### LINKED ISSUE #1650: Feature: Have a pop-up alerting of publication at approval
When the approve button is clicked, Describe immediately goes to the dashboard. There should be some fail safe in case someone clicks the wrong buttong (e.g., the "Edit" button is right next to it).

For cases without embargo, a pop-up message of "You're about to publish this dataset. There's no going back. Are you sure? (Yes/Cancel)"

For cases of embargo, it would be nice if the pop-up informs the approver of the embargo date and to confirm.

claudiawulee:
Display warning dialog before approving a work. Takes into account works with embargoes.

Closes #1650

### DIFF
--- a/app/views/works/show.html.erb
+++ b/app/views/works/show.html.erb
@@ -286,7 +286,7 @@
         <dt>Location notes</dt>
         <dd><%= @work.location_notes %></dd>
       <% end %>
-      <dt>Total Size</dt> 
+      <dt>Total Size</dt>
       <dd>
         <%= number_to_human_size(@work.total_file_size)%>
       </dd>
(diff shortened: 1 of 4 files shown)


app/views/works/show.html.erb
```diff
@@ -329,6 +329,21 @@
       $leftOfMessages.data("class", oldClass);
     });
 
+    $("#approve-button").on("click", (event) => {
+      var embargo = <%= @work.embargoed? %>;
+      var message = "";
+      if(embargo == true){
```
jrgriffiniii, line 335: 👍

jrgriffiniii (approved):

PR #1756, opened 2024-04-12
Handle file deletes through the Wizard

### LINKED ISSUE #1751: Delete files in the Wizard throws error
The delete file option in the Wizard (in the step adding files) does not work. 

This was introduced with PR https://github.com/pulibrary/pdc_describe/pull/1746 because now the user can view the files added and delete them as well.

hectorcorrea:
Closes #1751

### DIFF
--- a/app/controllers/works_wizard_controller.rb
+++ b/app/controllers/works_wizard_controller.rb
@@ -82,12 +82,12 @@ class WorksWizardController < ApplicationController
 
   # POST /works/1/file_upload
   def file_uploaded
-    files = pre_curation_uploads_param || []
-    if files.count > 0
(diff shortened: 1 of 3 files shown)


hectorcorrea:
This fixes the delete file bug, but in fixing it I noticed that we are handling file uploads in a way is not typical. I documented this "issue" here (https://github.com/pulibrary/pdc_describe/issues/1757) and will spend sometime digging into it.

JaymeeH (approved):
The delete functionality works, although there is a new quirk that I noticed. Not sure if this is intentional or not, but uploading a new file to the wizard requires the page to be saved or refreshed in order for it to appear in the file list. Whereas before — the file would appear immediately after being uploaded

Issue #1754, opened 2024-04-12
Layout issue for the cancel button


carolyncole:
I believe [adding the margin on all pages](https://github.com/pulibrary/pdc_describe/pull/1753/files#diff-0869e3045a3b5ca0e44ed193977d1568d686ca2373f4b93a8a868e4099fc932dR287) to fix the upload choice page is causing this issue:

<img width="1480" alt="Screenshot 2024-04-12 at 11 14 37 AM" src="https://github.com/pulibrary/pdc_describe/assets/1599081/1c36ee70-7c26-4357-a354-50544c9f9904">

leefaisonr:
I can look into this.

PR #1759, opened 2024-04-12
Adjust formatting for buttons in wizard

leefaisonr:
ref #1754


leefaisonr:
Buttons in wizard should now be aligned: 

image

hectorcorrea (approved):

Issue #1760, opened 2024-04-12
File upload in Wizard does not refresh file list immediately


hectorcorrea:
When uploading a new file via the Wizard the screen does not always display the newly added file on the screen. Notice how Uppy shows that `file_0001` has been uploaded, but the list of files in the back only shows the `README.txt`

![Screenshot 2024-04-12 at 1 44 21 PM](https://github.com/pulibrary/pdc_describe/assets/568286/3a37e5a2-43f3-4e99-856a-cd06e32bac96)

After adding a second file (`file_0002`) notice how the list in the back now shows all the files:

![Screenshot 2024-04-12 at 1 44 33 PM](https://github.com/pulibrary/pdc_describe/assets/568286/d71297ea-c59a-47b8-84a7-045653ad6fe6)

hectorcorrea:
Closes with #1762

Issue #1757, opened 2024-04-12
File upload endpoint for Wizard and Edit pages


hectorcorrea:
Currently the Wizard and the Edit page use the same endpoint for handling files uploaded via Uppy: they both go to `/works/1/upload-files` which is in the `WorksController`.

This works but it is a bit counterintuitive since typically Edit actions are handled on the `WorksController` and Wizard actions to be handled via the `WorksWizardController`. Having one of them (file upload) shared is probably not ideal.

Review if this needs to be refactored. I believe most of the functionality is already in a shared class, so having two end points call the shared code might be better since it will keep both controllers independent of each other.

hectorcorrea:
Fixed via #1762

PR #1762, opened 2024-04-12
Fixes Wizard delay in displaying uploaded files

hectorcorrea:
Uploading files added via the Wizard right away (instead of as a background job). This fixes the display issue reported in #1760. 

I also added Rails route to handle the Upload of files on the Wizard (instead of using the Upload process from the Work controller) as described in https://github.com/pulibrary/pdc_describe/issues/1757

The new code introduces a bug in the way we generate the UploadSnapshots, though. See below:

Screenshot 2024-04-15 at 11 46 09 AM

### DIFF
--- a/app/models/readme.rb
+++ b/app/models/readme.rb
@@ -24,21 +24,6 @@ class Readme
     end
   end
 
-  def attach_other(file)
-    extension = File.extname(file.original_filename)
-    filename = file.original_filename # TODO sanitize
-    puts "==> attach_other #{filename} started"
-    size = file.size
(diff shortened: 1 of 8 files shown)


app/models/readme.rb
```diff
@@ -25,6 +25,21 @@ def attach(readme_file_param)
     end
   end
 
+  def attach_other(file)
```
hectorcorrea, line 28: Hijacked the Readme class since it has a plain vanilla method to upload files.

carolyncole (approved):

PR #1775, opened 2024-04-18
Removing pre_curation_uploads attachment to active storage

carolyncole:
Also rename precuration_uploads_fast to precuration_uploads so that the _fast concept does not exists in the code base

Additionally making tests for publication more robust


app/models/work.rb
```diff
@@ -595,11 +594,7 @@ def s3_object_persisted?(s3_file)
     end
 
     def publish_precurated_files(user)
-      # An error is raised if there are no files to be moved
```
carolyncole, line 598: This is a private method that only gets called if the work is valid to approve.  That already makes sure there are files, so why check this again?  I could not write a test to test this code so I removed it.

spec/models/work_spec.rb
```diff
@@ -197,19 +202,32 @@
         work.reload
 
         expect(fake_s3_service_pre).to have_received(:publish_files).once
-        expect(work.pre_curation_uploads).to be_empty
```
carolyncole, line 200: All of these lines really just tested our mock.  No real test going on here.  Added some better tests below

spec/models/work_spec.rb
```diff
@@ -457,26 +475,6 @@
     end
   end
 
-  # Is this test providing any coverage?
```
carolyncole, line 460: This method was removed because we are not relying on active storage any more...  We have not for a long time

hectorcorrea:
oh my gosh this is such a welcomed PR!!!!! ❤️ I saw the `_fast` method yesterday and was wondering how come we have not refactored that code. Thanks @carolyncole

hectorcorrea (approved):
Awesome work. Tested Wizard and Edit screens in Staging and it seems to work.

leefaisonr (approved):
Thanks for the refactor!

PR #1781, opened 2024-04-23
Fixes file count

### LINKED ISSUE #1769: Incorrect file count on Uncurated Files section.
The file count on the Uncurated Files section of the Work Show page is off by one. Notice how it says "1 File or Directory" but there are in fact two files, in fact the label at the bottom says "showing 1 to 2 of 2 entries'

Screenshot 2024-04-17 at 8 35 44 AM

hectorcorrea:
Remove code that substracted one from the file count since it looks like we don't need it anymore.

Co-authored-by: Robert-Anthony Lee-Faison <leefaisonr@users.noreply.github.com>

Closes #1769

### DIFF
--- a/app/services/s3_query_service.rb
+++ b/app/services/s3_query_service.rb
@@ -325,16 +325,7 @@ class S3QueryService
 
   def count_objects(bucket_name: self.bucket_name, prefix: self.prefix)
     responses = s3_responses(bucket_name:, prefix:)
-    total_key_count = responses.reduce(0) { |total, resp| total + resp.key_count }
-    if total_key_count == 0
-      # if the bucket does not exist
-      # it will return 0
-      0
-    else
-      # if the bucket does exist
-      # s3 always sends back the bucket key as the first response, so we should not count it
-      total_key_count - 1
-    end
+    responses.reduce(0) { |total, resp| total + resp.key_count }
   end
 
   private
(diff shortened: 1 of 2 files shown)


app/services/s3_query_service.rb
```diff
@@ -325,16 +325,7 @@ def md5(io:)
...
-    if total_key_count == 0
-      # if the bucket does not exist
-      # it will return 0
-      0
-    else
-      # if the bucket does exist
-      # s3 always sends back the bucket key as the first response, so we should not count it
-      total_key_count - 1
-    end
+    responses.reduce(0) { |total, resp| total + resp.key_count }
```
hectorcorrea, line 328: AWS seems to return the correct key_counts and we don't need to subtract 1 from the total anymore. We tested with a bucket that does not exist and with a bucket that exists (and has files). It is not possible to have an empty bucket so that we could not test.

spec/services/s3_query_service_spec.rb
```diff
@@ -18,13 +18,6 @@
     {
       is_truncated: false,
       contents: [
-        {
-          etag: "\"#{s3_etag1}\"",
-          key: s3_bucket_key,
-          last_modified: s3_last_modified1,
-          size: 0,
-          storage_class: "STANDARD"
-        },
```
hectorcorrea, line 27: We are not seeing this on the actual AWS responses. The responses include files and directories but not the bucket_key as the first response.

bess (approved):

PR #1798, opened 2024-04-30
Switch to use gigabytes instead of gigibytes

### LINKED ISSUE #1793: Gigabytes vs Gigibytes display
We currently display units in Gigibytes (GiB) in PDC Describe but Globus uses Gigabytes (GB). This is an example of what Sarah noticed for work https://datacommons.princeton.edu/describe/works/303

Globus 
Screenshot 2024-04-26 at 4 13 58 PM

PDC Describe
Screenshot 2024-04-26 at 4 14 48 PM

For more info see [GB vs GiB—why the lowercase i](https://massive.io/file-transfer/gb-vs-gib-whats-the-difference/)

We should either switch to display file sizes in Gigabytes (GB) or update PDC Describe to use the correct abbreviation `GiB`.

Note this also applies to PDC Discovery -- we also display in Gigibytes there.

claudiawulee:
Closes #1793 

Rails does not specify a parameter to allow us to specify a specific base. So, we override the function to allow us to specify a base which by default is now 1000 instead of 1024.

### DIFF
--- a/config/initializers/number_to_human_size_converter.rb
+++ b/config/initializers/number_to_human_size_converter.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+# Source: https://github.com/rails/rails/issues/40054#issuecomment-674449143
+# Additional information: https://massive.io/file-transfer/gb-vs-gib-whats-the-difference/
+module ActiveSupport
+  module NumberHelper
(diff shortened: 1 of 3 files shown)


carolyncole (approved):
Thanks!  🐒

PR #1803, opened 2024-05-02
Handles abandoned works in the wizard

### LINKED ISSUE #1802: Works with an empty title cause error
A user was able to create an empty Work, a work with no data at all, not even a title. This results in an empty row show in the Dashboard.

Screenshot 2024-05-02 at 2 26 21 PM

You can inspect the page and figure out the URL for the work, which in this case it was: https://datacommons.princeton.edu/describe/works/307

Loading that page results in an exception on the code to produce the citation `app/helpers/application_helper.rb:33 in render_cite_as` 

## Steps to reproduce
* Click `Submit New`
* Check the `I understand this policy and wish to continue` box and click `Confirm`
* On the New Submission page close your browser tab (i.e. do not hit `Cancel` or `Create New`)
* Go back to the Dashboard and the empty record will show in the list of datasets

Notice that clicking `Cancel` on the New Submission page does clear the record correctly, but this issue happens when user don't hit `Cancel`

## Solutions?
* Fix the code that produces the citation to not throw an exception.
* We won't be able to completely prevent a user from creating an empty record so perhaps we should detect the empty record and (1) give it a temporary title (e.g. "Untitled") so that it can be clicked on, and then (2) throw back the user to the correct step in the Wizard to finish the data entry process.

hectorcorrea:
Displays the record as "untitled" and forces the user back to the wizard.

Closes #1802

### DIFF
--- a/app/views/users/_dataset_table.html.erb
+++ b/app/views/users/_dataset_table.html.erb
@@ -13,7 +13,11 @@
       <% datasets.each do |ds| %>
         <tr>
           <td>
-            <a href="<%= work_path(ds) %>"><%= ds.title %></a>
+            <% if ds.title.blank? %>
+              <a href="<%= edit_work_wizard_path(ds) %>">(untitled)</a>
+            <% else %>
+              <a href="<%= work_path(ds) %>"><%= ds.title %></a>
+            <% end %>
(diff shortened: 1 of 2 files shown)


leefaisonr (approved):
Looks good to me. Thanks!

PR #1804, opened 2024-05-03
Better logging for error updating snapshots

### LINKED ISSUE #1801: NoMethodError: undefined method `doi' for nil:NilClass
## Backtrace

    line 83 of [PROJECT_ROOT]/app/controllers/works_wizard_controller.rb: rescue in file_uploaded
    line 70 of [PROJECT_ROOT]/app/controllers/works_wizard_controller.rb: file_uploaded

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/99214/faults/106907891)

It is very strange that we got a `nil` work on [this method](https://github.com/pulibrary/pdc_describe/blob/main/app/controllers/works_wizard_controller.rb#L70-L88). Should investigate:

```
# POST /works/1/file_upload
  def file_uploaded
    upload_service = WorkUploadsEditService.new(@work, current_user)
    # By the time we hit this endpoint files have been uploaded by Uppy submmitting POST requests
    # to /works/1/upload-files-wizard therefore we only need to delete files here and update the upload snapshot.
    @work = upload_service.snapshot_uppy_and_delete_files(deleted_files_param)

    prepare_decorators_for_work_form(@work)
    if params[:save_only] == "true"
      render :file_upload
    else
      redirect_to(work_review_path)
    end
  rescue StandardError => active_storage_error
    Rails.logger.error("Failed to attach the file uploads for the work #{@work.doi}: #{active_storage_error}")
    flash[:notice] = "Failed to attach the file uploads for the work #{@work.doi}: #{active_storage_error}. Please contact rdss@princeton.edu for assistance."

    redirect_to work_file_upload_path(@work)
  end
```

hectorcorrea:
Added a slightly better logging to help us troubleshoot this issue if it happens again. Notice that the error only happened in Staging and while we were in the midst of refactoring the Wizard and the File Upload process so we might never seen this issue again. But if we do we'll have better logging.

Closes #1801

### DIFF
--- a/app/controllers/works_wizard_controller.rb
+++ b/app/controllers/works_wizard_controller.rb
@@ -80,9 +80,12 @@ class WorksWizardController < ApplicationController
     else
       redirect_to(work_review_path)
     end
-  rescue StandardError => active_storage_error
-    Rails.logger.error("Failed to attach the file uploads for the work #{@work.doi}: #{active_storage_error}")
-    flash[:notice] = "Failed to attach the file uploads for the work #{@work.doi}: #{active_storage_error}. Please contact rdss@princeton.edu for assistance."
+  rescue => ex
+    # Notice that we log the URL (rather than @work.doi) because sometimes we are getting a nil @work.
+    # The URL will include the ID and might help us troubleshoot the issue further if it happens again.
+    # See https://github.com/pulibrary/pdc_describe/issues/1801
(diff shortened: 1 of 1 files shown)


bess (approved):
@hectorcorrea This is an improvement, thank you! I added a line to also report the error to honeybadger, though, where I think we'll be more likely to notice it if it has happened.

bess (approved):

PR #1806, opened 2024-05-08
Fixes provenance bug in the Wizard

### LINKED ISSUE #1785: Strange provenance for files added/deleted/replaced
Neggin noticed a strange provenance display issue for the files on this record: https://datacommons.princeton.edu/describe/works/295

PDC Describe indicates that an "Unknown user" deleted two files at `17:33`, Neggin said the curators were not updating the data on AWS directly which normally would explain the "unknown user" message. It is curious that we logged that the user `Yuanyu` was making changes at that exact same time: 

Screenshot 2024-04-24 at 2 23 06 PM

This looks like a bug in out logic to determine changes in AWS, but I am not 100% sure where or how.

The log in AWS shows in fact several changes to the README, it looks like the file was replaced (deleted and readded or overwritten?) 

Screenshot 2024-04-24 at 2 27 36 PM

The fact that this happened on the README (which is a specific step on with Wizard with its own set of rules) might be something to keep an eye on if this happens again. We might need to revisit the logic to update this file when users select a new file via Uppy.

hectorcorrea:
Fixes the issue in which some files where not properly tracked during the upload process in the Wizard. 

The problem was that we were not updating the snapshot right away. We were only updating the snapshot if the user hit save but it was possible for the user to click the home page and bypass the "save" button after they have added files to the work. Those files will be picked up at a later time but by then the system will assign them to "Unknown user".

The new code updates the snapshot as soon as the file is updated to AWS and therefore, even if the user bypasses the "save" button we already updated the snapshot with the correct user associated to the action.

Closes #1785


app/controllers/works_wizard_controller.rb
```diff
@@ -10,6 +10,7 @@
 #     attachment_selected -> file_other ->                  review -> validate -> [ work controller ] show & file_list
 #                         \> file_upload -> file_uploaded -^
 
+# rubocop:disable Metrics/ClassLength
```
hectorcorrea, line 13: We exceeded by one line and I didn't want to refactor the code.

app/controllers/works_wizard_controller.rb
```diff
@@ -61,10 +61,10 @@ def file_upload
   end
 
   # POST /works/1/upload-files-wizard (called via Uppy)
-  # this puts the files into AWS, but does not capture anything in the Upload Snapshot.  THat occurs when the user hits next
   def upload_files
     @work = Work.find(params[:id])
-    params["files"].each { |file| upload_file(file) }
+    upload_service = WorkUploadsEditService.new(@work, current_user)
```
hectorcorrea, line 66: Having the upload go through our `WorkUploadsEditService` makes sure the correct logic is executed.

app/services/work_uploads_edit_service.rb
```diff
@@ -53,14 +53,30 @@ def delete_uploads(deleted_files)
...
       snapshot = BackgroundUploadSnapshot.new(work:)
       snapshot.store_files(added_files, pre_existing_files: last_snapshot&.files, current_user: @current_user)
       snapshot.save
+
+      # ...adds the file to AWS directly and mark them as complete in the snapshot
       added_files.map do |file|
-        new_path = "/tmp/#{file.original_filename}"
-        FileUtils.mv(file.path, new_path)
-        AttachFileToWorkJob.perform_later(file_path: new_path, file_name: file.original_filename, size: file.size, background_upload_snapshot_id: snapshot.id)
+        if upload_file(file)
```
hectorcorrea, line 64: Replaced the `AttachFileToWorkJob` since are now uploading the files to AWS directly using `s3_query_service.upload_file`

bess (approved):

PR #1813, opened 2024-05-14
Ensures that one and only one README file is uploaded and that there is one or more files uploaded for any given Work

### LINKED ISSUE #1772: Validation for Awaiting Approval (end of wizard) Need to check for Readme and Other files explicitly
The user should not be able to mark a submission as complete unless a README file exists AND some other file.

Completion Criteria:
- [x] When a user Click Grant License and Complete the controller should validate the uploaded files
  - [x] There Must be one and only one README.
  - [x] There Must be one or more other files if they are uploading locally
- [x] If the right files do not exists the user will see an error on the page and the work will not update state
  - [x] Readme error: "You Must include a README. [Please upload  one][link to readme step]"
  - [x] File error "You Must include at least one file. [Please upload  one][link to upload step]"

jrgriffiniii:
refs #1772

### DIFF
--- a/app/controllers/works_wizard_controller.rb
+++ b/app/controllers/works_wizard_controller.rb
@@ -110,6 +110,7 @@ class WorksWizardController < ApplicationController
   # PATCH /works/1/validate-wizard
   def validate
     @work.submission_notes = params["submission_notes"]
+
     if params[:save_only] == "true"
       @work.save
       render :review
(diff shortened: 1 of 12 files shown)


jrgriffiniii:
I suspect that the error messages do not directly match what is required, I am very sorry. Once I have these matching, I can open this and request review.

app/controllers/works_wizard_controller.rb
```diff
@@ -158,6 +167,19 @@ def file_location_url
   end
   helper_method :file_location_url
 
+  def rescue_aasm_error
```
carolyncole, line 170: @jrgriffiniii why was this redefined here?

app/services/work_validator.rb
```diff
@@ -161,6 +162,7 @@ def validate_metadata
     def validate_files
       return if @work.resource.migrated
       readme = Readme.new(work, nil)
-      errors.add(:base, "Must provide a README") if readme.blank?
+      errors.add(:base, "You must include a README. <a href='#{work_readme_select_path(work)}'>Please upload one</a>") if readme.blank?
+      errors.add(:base, "You must provide only one README file upload") if work.readme_uploads.length > 1
```
carolyncole, line 166: I believe we will allow the user to add other version of the PDF in files

carolyncole (changes requested):
@jrgriffiniii this PR looks good, but I believe the user can add additional readme formats in the files.  That means the check in the validation should be removed.  As long as there is one the work is valid.  If there are 3 the work should still be valid.

carolyncole (approved):

carolyncole:
@jrgriffiniii I'm happy to merge this PR.  I want to be certain you are aware that I can continue and submit even if I do not submit any other file than a readme.

PR #1825, opened 2024-05-28
Update production.rb to the new mail server

carolyncole:
Since this is a production only change, I ran it manually before making the PR and the mail went through...
Screenshot 2024-05-28 at 2 48 01 PM


kayiwa (approved):
OK! Thanks @carolyncole

Issue #1834, opened 2024-06-04
File provenance when deleting files


hectorcorrea:
When I delete a file the provenance register two records for the delete. One correctly assigned to the user that delete the file and one indicating "unknown user. The system should not log the "unknown user" provenance in this case since we know the user that deleted the file:

![Screenshot 2024-06-04 at 11 59 23 AM](https://github.com/pulibrary/pdc_describe/assets/568286/f2b356e7-e94b-46ac-9f12-2e37306176b3)

This is related to issue #1833 but not the same.

Look at the snapshots and figure out why the snapshot after the delete (if it is created) is not the same as the snapshot when you come back.

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/pdc_describe/1834?planning-poker) with Zenhub @hectorcorrea @JaymeeH @jrgriffiniii @leefaisonr

PR #1838, opened 2024-06-06
Fixes display issue when deleting a file

### LINKED ISSUE #1833: File name display when delete is clicked more than once
When I click on the "Delete File" link the file name is strike through but the system also displays a "#" symbol before the name and it should not:

Screenshot 2024-06-04 at 11 54 50 AM

If I then click "Undo delete" the file will keep the "#" before the name (it should not)

Screenshot 2024-06-04 at 11 54 57 AM

If I then click "Delete file" again, a new "#" symbol will be prefixed to the name:

Screenshot 2024-06-04 at 11 55 05 AM

### LINKED ISSUE #1835: Delete provenance count is wrong
If I click "Delete file" and then hit "cancel"

Screenshot 2024-06-04 at 12 03 30 PM

and then Edit the record again and click "Delete file" the display is incorrect (this is documented in #1833)

Screenshot 2024-06-04 at 12 04 09 PM

If I click "Save" the file will be deleted but the provenance will log an incorrect count for the number of files deleted:

Screenshot 2024-06-04 at 12 04 33 PM

Make sure the count is correct when deleting a file even in this scenario.

hectorcorrea:
Fixes the display issue when deleting a file **and** the file count of deleted file too.

Screenshot 2024-06-06 at 11 35 53 AM

Screenshot 2024-06-06 at 11 50 11 AM

Closes #1833 
Closes #1835

Issue #1834 is not fixed with this PR. That will be handled separately.

### DIFF
--- a/spec/system/work_edit_spec.rb
+++ b/spec/system/work_edit_spec.rb
@@ -31,13 +31,6 @@ RSpec.describe "Creating and updating works", type: :system, js: true do
       allow(fake_s3_service).to receive(:file_url).with(contents1.key).and_return("https://example-bucket.s3.amazonaws.com/#{contents1.key}")
       allow(fake_s3_service).to receive(:file_url).with(contents2.key).and_return("https://example-bucket.s3.amazonaws.com/#{contents2.key}")
       allow(fake_s3_service).to receive(:file_url).with(contents3.key).and_return("https://example-bucket.s3.amazonaws.com/#{contents2.key}")
-      # fake_aws_client = double(Aws::S3::Client)
-      # fake_s3_resp = double(Aws::S3::Types::ListObjectsV2Output)
-      # fake_aws_client.stub(:list_objects_v2).and_return(fake_s3_resp)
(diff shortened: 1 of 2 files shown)


spec/system/work_edit_spec.rb
```diff
@@ -69,6 +62,19 @@
...
     end
 
+    it "allows users to cancel the delete of one of the uploads" do
+      allow(ActiveStorage::PurgeJob).to receive(:new).and_call_original
+      allow(fake_s3_service).to receive(:client_s3_files).and_return([contents1, contents2], [contents2])
+
+      expect(page).to have_link "us_covid_2019.csv"
+      click_on "delete-file-#{contents1.safe_id}"
+      expect(page).not_to have_link "us_covid_2019.csv"
+      expect(page.html.include?("<s> us_covid_2019.csv</s>")).to be true
```
hectorcorrea, line 72: This is the line that test that the extra `#` is not added to the text.

jrgriffiniii (approved):

PR #1847, opened 2024-06-24
Fixes the bad provenance logged when deleting files

### LINKED ISSUE #1834: File provenance when deleting files

hectorcorrea:
Logs the correct provenance (who deleted the file) when deleting a file via the Edit screen:

Screenshot 2024-06-24 at 4 23 31 PM

Closes #1834

### DIFF
--- a/app/services/work_uploads_edit_service.rb
+++ b/app/services/work_uploads_edit_service.rb
@@ -11,9 +11,12 @@ class WorkUploadsEditService
   def update_precurated_file_list(added_files, deleted_files)
     delete_uploads(deleted_files)
     add_uploads(added_files)
-    if work.changes.count > 0
+    file_list_changed = deleted_files.count > 0 || added_files.count > 0
+    if file_list_changed
       s3_service.client_s3_files(reload: true)
-      work.reload # reload the work to pick up the changes in the attachments
+      # Pick up the changes in the attachments and assings them to the current user
+      work.reload
+      work.reload_snapshots(user_id: current_user.id)
     end
 
     work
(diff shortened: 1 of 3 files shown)


app/services/work_uploads_edit_service.rb
```diff
@@ -11,9 +11,12 @@ def initialize(work, current_user)
...
     delete_uploads(deleted_files)
     add_uploads(added_files)
-    if work.changes.count > 0
+    file_list_changed = deleted_files.count > 0 || added_files.count > 0
+    if file_list_changed
       s3_service.client_s3_files(reload: true)
-      work.reload # reload the work to pick up the changes in the attachments
+      # Pick up the changes in the attachments and assings them to the current user
+      work.reload
+      work.reload_snapshots(user_id: current_user.id)
```
hectorcorrea, line 19: This is the big change on this PR: I am forcing to reload the snapshots so that the changes are assigned to the current user.

app/services/work_uploads_edit_service.rb
```diff
@@ -45,9 +48,7 @@ def delete_uploads(deleted_files)
 
       deleted_files.each do |filename|
         s3_service.delete_s3_object(filename)
-        work.track_change(:removed, filename)
       end
-      work.log_file_changes(@current_user.id)
```
hectorcorrea, line 50: The extra logs are not needed since the `work.reload` logs the changes.

leefaisonr (approved):
LGTM - Thanks!

PR #1858, opened 2024-07-18
added additional testing

20LM22:
Advances #1855 by adding additional testing coverage.

PR #1861, opened 2024-07-19
Restoring the commented blocks and removing code comments

jrgriffiniii:
While you are correct in removing https://github.com/pulibrary/pdc_describe/pull/1858/files#diff-8c3fb4a34f9f014040e82fda94adff3d6344eacfbcea5bfd039afab491c205a3L15 and https://github.com/pulibrary/pdc_describe/pull/1858/files#diff-88f3a762d2fd74e42ef154e979619b890a12d146214366033da05a0af4357c26L18, my concern is that this must be explicitly confirmed before this is introduced into `main` (if it has not yet been removed, this might break any attempts to deploy `main`, or any branches rebased upon `main`).

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -278,28 +278,24 @@ class WorksController < ApplicationController
       params[:work] || {}
     end
 
-    # not a route, not called elsewhere
     def patch_params
       return {} unless params.key?(:patch)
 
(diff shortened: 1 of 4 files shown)


jrgriffiniii:
Closing this following a discussion for https://github.com/pulibrary/pdc_describe/pull/1858/files#

PR #1884, opened 2024-08-05
uri encoding the source key in dataspace file copy

### LINKED ISSUE #1874: Some objects are not migrating cleanly
We have quite a few migration failures with the error "the specified key does not exist."

I wonder if this is a case of AWS illegal characters in the filename, and we need to tweak [this class](https://github.com/pulibrary/pdc_describe/blob/977ce3e52d23b0b9d47fe3d41d6b5941d58849b5/app/services/file_rename_service.rb)? 
retries

## Backtrace

    line 239 of [PROJECT_ROOT]/app/services/s3_query_service.rb: copy_file
    line 11 of [PROJECT_ROOT]/app/jobs/dspace_file_copy_job.rb: perform

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/99214/faults/109875023)

carolyncole:
Turns out you need to url encode the key if you put it with the bucket name. 
'bucket/key+with+space' needs to be 'bucket/key%2Bwith%2Bspace' for copy_object 
even though aws will tell you the key is 'key+with+space'

Screenshot 2024-08-05 at 8 50 37 AM

fixes #1874

### DIFF
--- a/app/services/s3_query_service.rb
+++ b/app/services/s3_query_service.rb
@@ -236,7 +236,7 @@ class S3QueryService
     if size > part_size
       copy_multi_part(source_key:, target_bucket:, target_key:, size:)
     else
-      client.copy_object(copy_source: source_key, bucket: target_bucket, key: target_key, checksum_algorithm: "SHA256")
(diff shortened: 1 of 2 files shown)


leefaisonr (dismissed):

leefaisonr (approved):
Thanks @carolyncole!

PR #1894, opened 2024-08-13
Configuring sidekiq to allow for greater latency and size

### LINKED ISSUE #1893: Redis should be allowed to have many jobs
### What maintenance needs to be done?

Right now the redis health check errors when the job count greater than 100 and a latency greater than 30 second.
Screenshot 2024-08-13 at 8 49 08 AM

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?

We keep getting notified on slack when someone approves a large work in staging.  Once the production system is monitored by checkmk this will occur there too..

### Acceptance criteria
 - [ ] I should be allowed to have 10,000 jobs
 - [ ] The latency should be allowed to be 5 minutes

### Implementation notes, if any

Hopefully this is configurable in rails health check

carolyncole:
fixes #1893

### DIFF
--- a/config/initializers/health_monitor.rb
+++ b/config/initializers/health_monitor.rb
@@ -2,13 +2,13 @@
 HealthMonitor.configure do |config|
   config.cache
   config.redis
-  config.sidekiq
 
   # Make this health check available at /health
   config.path = :health
 
   config.sidekiq.configure do |sidekiq_config|
     sidekiq_config.latency = 3.hours
+    sidekiq_config.queue_size = 10_000
   end
 
   config.error_callback = proc do |e|


config/initializers/health_monitor.rb
```diff
@@ -2,13 +2,13 @@
 HealthMonitor.configure do |config|
   config.cache
   config.redis
-  config.sidekiq
```
carolyncole, line 5: This was creating a second entry for sidekiq on the page

leefaisonr (approved):
Thanks @carolyncole

PR #1897, opened 2024-08-15
Moved message input box to the top of the page

### LINKED ISSUE #1881: Messaging sidebar shows recent message at the top but you have to respond below
### Issues

For very extensive comments history, it became clear that the area to provide additional comments is way at the bottom while the most recent message is near the stop. Some discussions is needed as to what order the messages should appear (ascending vs descening). However, I think it would make sense for the comments box to appear next to the recent messages.

Screen Shot 2024-08-02 at 7 54 09 AM
...
...
Screen Shot 2024-08-02 at 7 54 16 AM

### URL

https://datacommons.princeton.edu/describe/works/301

### Acceptance criteria
- [ ] Move the box to enter messages to the top of the message section 
- [ ] Keep the messages in the current sort order
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools

### Notes

We may do additional user testing in the future on this item, but that is not part of this ticket.

hectorcorrea:
This is how the screen looks now:

enter-message

Closes #1881

### DIFF
--- a/app/views/works/_work_activity_messages.html.erb
+++ b/app/views/works/_work_activity_messages.html.erb
@@ -1,3 +1,21 @@
+<div id="new-message-section">
+  <%= form_with url: add_message_work_path(@work) do |f| %>
+    <textarea id="new-message" name="new-message" class="new-message" placeholder="leave a message" rows="5" cols="30"></textarea>
+    <br/>
+    <%= f.submit("Message", class: "btn btn-secondary") %>
+    <span id="new-message-help" class="new-message-help">
+      Simple Markdown is accepted.
+      <a href="#" id="new-message-help-more">See more.</a>
+    </span>
+    <br/>
+    <span id="new-message-help-part2" class="new-message-help hidden">
+      e.g. *italics*, **bold**, # Header 1, ## Header 2, ```code```.
+      Use @netid to refer to others.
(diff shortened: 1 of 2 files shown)


app/views/works/_work_activity_messages.html.erb
```diff
@@ -19,15 +37,21 @@
...
-      <br/>
-      <%= f.submit("Message", class: "btn btn-secondary") %>
-      <span class="new-message-help">
-        Simple Markdown is accepted, e.g. *italics*, **bold**, # Header 1, ## Header 2, ```code```. Use @netid to refer to others.
-      </span>
-    <% end %>
-  </div>
 </div>
+
+<script>
```
carolyncole, line 42: @hectorcorrea was there a reason you added this script?  Did the toggle button stop working?  It was working for me earlier in the week.

hectorcorrea: The help text wrapped funny so I had added the toggle so that it does not look as funny.

(that text always wrapped funny, but since it was a the very bottom of the page it was not obvious, now that it is at the top it was more prominent)

carolyncole (approved):

Issue #1610, opened 2023-11-17
Allow a Princeton Graduate Student utilize the PPPL work flow by default


carolyncole:
For the moment I am going to set `ar3676` default department to PPPL

In the future we have many options:
A UI for a curator to change a person's default department? 
The ability to choose the workflow (something like a "Submit New" "Submit New PPPL")? 
The ability for the user to choose their default department in their profile?  

Other options may exists.

carolyncole:
ran below to set the graduate student up for now
```
user = User.where(uid: 'ar3676').first
user.default_group_id = Group.plasma_laboratory.id
user.save
```

carolyncole:
The slack conversation: https://pulibrary.slack.com/archives/C01555TP6HE/p1700251479472559?thread_ts=1700249372.036709&cid=C01555TP6HE

hectorcorrea:
PR #1904 introduced a Rake task to do this, but we should update the UI so that admin can make this change directly from PDC Describe without having to SSH to the server. 

This will become even more important now that we are starting to support GAP accounts (see https://github.com/pulibrary/pdc_describe/issues/1901) and the default group will very likely be wrong for those users.

hectorcorrea:
PR #1911 takes care of most of it since it allows and admin (or a moderator) to change the default group for user without developer intervention. Still a manual process but at least available within the application.

PR #1904, opened 2024-08-20
Set PPPL user rake task

hectorcorrea:
Automated the steps to a rake task so it is easier to run. Stole the steps from the documentation that @carolyncole  had provided in the README.

Part of #1610


carolyncole (approved):
Thanks @hectorcorrea!

PR #1912, opened 2024-08-22
Implements support for providing users with the ability to revert Works awaiting approval to draft Works

### LINKED ISSUE #1852: As a PDC user, I like to be able to switch from awaiting_approval to draft
### User story

Users would go through the deposit process and may upload a README as a temporary solution and skip the dataset upload. Unfortunately when they get to the end, it puts it in a "awaiting_approval" state and they are not able to edit the submission further. This means they can't upload the data files or change the metadata. While those with Moderator status can, this imposes delays and third-party human intervention.

There are two needs:
- The ability for a Moderator to switch it back to "draft" state. This is analogous in DataSpace of not approving the submission and sending it back with instructions
- The ability for a data submitter to revert back to "draft" state so they can finish their submission

### Concrete example

An example of something being stuck is this one where it is awaiting approval but the depositor failed to upload files:
https://datacommons.princeton.edu/describe/works/304

jrgriffiniii:
Resolves #1852

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -148,6 +148,13 @@ class WorksController < ApplicationController
     redirect_to work_path(@work)
   end
 
+  def revert_to_draft
+    @work = Work.find(params[:id])
+    @work.revert_to_draft!(current_user)
+
+    redirect_to work_path(@work)
+  end
+
   def assign_curator
     work = Work.find(params[:id])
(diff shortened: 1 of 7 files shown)


hectorcorrea (approved):

hectorcorrea:
Nice work @jrgriffiniii -- this is such a nice new feature

PR #1915, opened 2024-08-28
Making sure the prefix has a slash in it

### LINKED ISSUE #1849: Bug in prefix_filename in background_upload_snapshot.rb
This method does not joins paths correctly if `work.prefix` does not end with forward slash. 

```
  def prefix_filename(filename)
    if filename.include?(work.prefix)
      filename
    else
       "#{work.prefix}#{filename}"
    end
  end
```

I don't think the running code has an issue with this, but we do have tests looking for the wrong path because of this bug:

```
WorksController valid user login with new file uploads for an existing Work without uploads handles the update page
     Failure/Error: expect(background_snapshot.files.first["filename"]).to eq "10.34770/123-abc/1us_covid_2019.csv"
     
       expected: "10.34770/123-abc/1us_covid_2019.csv"
            got: "10.34770/123-abc/1/us_covid_2019.csv"
```

carolyncole:
the real code always ends with a slash https://github.com/pulibrary/pdc_describe/blob/main/app/services/s3_query_service.rb#L81

fixes #1849

### DIFF
--- a/spec/support/s3_query_service_specs.rb
+++ b/spec/support/s3_query_service_specs.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-def stub_s3(data: [], bucket_url: nil, prefix: "10.34770/123-abc/1", bucket_name: "example-bucket")
+def stub_s3(data: [], bucket_url: nil, prefix: "10.34770/123-abc/1/", bucket_name: "example-bucket")
   @s3_client = instance_double(Aws::S3::Client)
   allow(@s3_client).to receive(:head_object)
   allow(@s3_client).to receive(:delete_object)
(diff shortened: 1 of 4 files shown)


spec/support/s3_query_service_specs.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-def stub_s3(data: [], bucket_url: nil, prefix: "10.34770/123-abc/1", bucket_name: "example-bucket")
+def stub_s3(data: [], bucket_url: nil, prefix: "10.34770/123-abc/1/", bucket_name: "example-bucket")
```
carolyncole, line 3: This is the fix.  The rest is just updating the tests to match the new prefix

hectorcorrea (approved):

hectorcorrea (approved):

PR #1946, opened 2024-09-25
Trying a new way to set relative url

hectorcorrea:


hectorcorrea:
Fixed via https://github.com/pulibrary/pdc_describe/pull/1947

PR #1962, opened 2024-10-09
Store the resource type and resource type general in the correct datacite fields

### LINKED ISSUE #1934: Resource Type and Resource Type General fields are behaving strangely
### Expected behavior
As a curator or moderator completing a new submission, I expect to be able to enter appropriate free-text values for Resource Type and select any option from the Resource Type General drop-down. For example, when I select "Software" from the dropdown for "Resource Type General" and enter a short descriptive term, such as "Source Code," in the "Resource Type" field, like so:

PDC-Describe_NewSubmission-Describe-CC_ResourceType_2024-09-19

...I expect to be able to complete the submission without issue.

### Actual behavior
Instead, I get an error message telling me that "Resource Type: General resource type cannot be nil," when I can plainly see that what I entered is still there (not nil):

PDC-Describe_NewSubmission-Describe_ResourceTypeMessage_2024-09-19

The message is especially confusing because it seems to conflate two fields: Resource Type and Resource Type General. According to [DataCite v4.4 documentation](https://schema.datacite.org/meta/kernel-4.4/doc/[DataCite-MetadataKernel_v4.4.pdf](https://schema.datacite.org/meta/kernel-4.4/doc/DataCite-MetadataKernel_v4.4.pdf)), Resource Type is meant to be a free-format text field (with guidance to choose "a single term of some detail"), while Resource Type General has a controlled vocabulary (which the current drop-down selector reflects).

### Steps to replicate
1. Start a new submission
2. After entering required metadata, go to the "Curator Controlled" tab
3. Enter something other than "Dataset" for Resource Type and select something other than "Dataset" for Resource Type General
4. Complete the submission process
5. Find yourself returned to the Describe step, Curator Controlled tab, with an error message at the top

### Impact of this bug
PRDS cannot complete migration without being able to accurately assign Resource Type and Resource Type General to existing records.

### Acceptance criteria
- [ ] The Resource Type field allows free text (no controlled vocabulary) that gets saved to the model on next or save
- [ ] All options from the drop-down selector for the Resource Type General field are available for submission
- [ ] In the XML version of the record for DataCite, the text entered in the PDC Resource Type field becomes the value for the element resourceType, and the selected value for the PDC field Resource Type General becomes the value for the attribute resourceTypeGeneral
- [ ] An error is thrown if Resource Type and/or Resource Type General do truly end up with nil values
- [ ] The error message is corrected as follows:
  - [ ] If the Resource Type field ends up empty, then the error is "Resource Type cannot be nil."
  - [ ] If the Resource Type General field ends up empty, then the error is "Resource Type General cannot be nil."
- [ ] Data sent to DataCite is still valid XML

carolyncole:
We were not storing the Resource Type General and Storing the Resource type in the Resource Type General location, which was what was causing the validation issues

fixes #1934


spec/models/pdc_serialization/datacite_spec.rb
```diff
@@ -143,63 +147,64 @@
...
           expect(resource_type.resource_type_general.value).to eq "InteractiveResource"
         end
         it "Model" do
-          resource_type = described_class.datacite_resource_type("Model")
+          resource_type = described_class.datacite_resource_type("Model", "aeroplane")
           expect(resource_type.resource_type_general.value).to eq "Model"
         end
         it "PhysicalObject" do
-          resource_type = described_class.datacite_resource_type("PhysicalObject")
+          resource_type = described_class.datacite_resource_type("PhysicalObject", "a rock")
```
hectorcorrea, line 183: 🤣

hectorcorrea (approved):

hectorcorrea:
@carolyncole wow...that was really broken, thanks for taking care of it!

PR #1982, opened 2024-10-28
Fixes time out issue on records with very large number of provenance messages

### LINKED ISSUE #1977: Error Pages for Large Approvals
### Expected behavior
When approving an item in PDC, we expect the page to load with the banner about how the files are moving now. 

### Actual behavior
When approving items containing more than 20k files and directories, it produces an error page saying "Looks like something went wrong" 
image

If you try to reload the describe page at this stage, it will either: 
a) not reload (timing out to a "looks like something went wrong page) or
b) it'll reload and it may or may not look like the approval was successful. 
However, (in cases where it looks like the approval did not work) if you wait 5-10 minutes and are able to reload the describe page again, it may change to "approved" even though no further action was taken (besides refreshing the page).  

### Steps to replicate
Approve a dataset with 20k or more files and directories. 

### Impact of this bug
We have to cross our fingers that everything is actually working ok even though we see a "something went wrong" message. 
There is no way to know if the files are continuing to transfer over, so we just have to check back by reloading the describe page and see if the counts are going up (though in one case the describe page would always time out, so we had to ask RDSS. File transfers of this size are currently taking more than a single business day to complete (when items contain more than 20k files and directories). 

### Acceptance criteria
- [ ] If this change touches any of the files referenced in `.rubocop_todo.yml` create a new ticket to address the rubocop  todo.
- [ ] The page will stop going to "looks like something went wrong" 
- [ ] The page in describe will show the normal banner for approvals 

NOTE: this may be an area that becomes a feature request to show something like a progress tracker of how long remains before a dataset is approved and the publication is fully accessible. 

### Honeybadger link and code snippet, if applicable

```
```

### LINKED ISSUE #1978: System provenance messages for large works
We are trying to display all system provenance messages on the page by default and this seems to be causing the page to grow too big and times out. 

For example, work https://datacommons.princeton.edu/describe/works/383 has 10,539 "system" notifications that we are rendering on the page which causes the "Looks like something went wrong" message to be rendered instead of the page. 

I suspect this is because either the page is too large or we take too long generating the page.

```
w = Work.find(383)
w.activities.count {|wa| wa.activity_type == "SYSTEM" }
=> 10539

w.activities.count
=> 10547
```

From that I can tell most of these "system" notifications are about empty files: "Warning: Attempted to publish empty S3 file"

This is very likely related to https://github.com/pulibrary/pdc_describe/issues/1977

We should implement a way to toggle these notifications, for example, perhaps we don't display SYSTEM notifications by default and allow the user to turn them on if they want to. Or some variation of this.

hectorcorrea:
Addresses the issue of loading a work with a large number of messages in the provenance log. 

The culprit was two fold:
1. The code to mark notifications as read iterates one by one on each notification which translates on individual SQL SELECT commands. This PR batches the SQL UPDATE which is much faster, particularly when there are thousand of notifications in the database.
2. There is also a problem with the way we determine the user name to display for each notification. The code issues a SQL SELECT for each notification to retrieve the user name to display (in addition to the user's netid). Even with Rails using its internal cache this is slow when there are thousands of SYSTEM generated notifications. This PR displays only the first 100 SYSTEM notifications and let's the user know if there are more to be aware of (see screenshot below).

Screenshot 2024-10-30 at 5 39 24 PM

Closes #1978 
Closes #1977

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -299,8 +299,12 @@ class Work < ApplicationRecord
     # Rails batching information:
     #   https://guides.rubyonrails.org/active_record_querying.html
     #   https://api.rubyonrails.org/classes/ActiveRecord/Batches.html
+
+    # Disable this validation since we want to force a SQL UPDATE.
+    # rubocop:disable Rails/SkipsModelValidations
     now_utc = Time.now.utc
     WorkActivityNotification.joins(:work_activity).where("user_id=? and work_id=?", user_id, id).in_batches(of: 1000).update_all(read_at: now_utc)
+    # rubocop:enable Rails/SkipsModelValidations
   end
 
   def current_transition
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -292,20 +292,15 @@ class Work < ApplicationRecord
   # In practice, the user_id is the id of the current user and therefore this method marks the current's user
   # notifications as read.
   def mark_new_notifications_as_read(user_id)
-    # https://guides.rubyonrails.org/active_record_querying.html
-    # TODO: Although we are batching the query (see batch_size parameter) we are still
-    # calling save on *each* notification and that might be too slow, is there a way to
-    # batch the save too?
-    WorkActivityNotification.joins(:work_activity).where("user_id=? and work_id=?", user_id, id).find_each(batch_size: 1000) do |unread_notification|
-      unread_notification.read_at = Time.now.utc
-      unread_notification.save
-    end
-
-    # TODO: Test if this work to save in batches
-    # ====
-    # now_utc = Time.now.utc
-    # WorkActivityNotification.joins(:work_activity).where("user_id=? and work_id=?", user_id, id).find_in_batches.update_all(read_at: now_utc)
-    # ====
+    # Notice that we fetch and update the information in batches
+    # so that we don't issue individual SQL SELECT + SQL UPDATE
+    # for each notification.
+    #
+    # Rails batching information:
+    #   https://guides.rubyonrails.org/active_record_querying.html
+    #   https://api.rubyonrails.org/classes/ActiveRecord/Batches.html
+    now_utc = Time.now.utc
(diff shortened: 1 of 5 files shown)


hectorcorrea:
This PR has been deployed to staging and you can see how fast it loads on this record that has 4000+ artificially generated notifications: https://pdc-describe-staging.princeton.edu/describe/works/649

app/models/work.rb
```diff
@@ -292,13 +292,19 @@ def new_notification_count_for_user(user_id)
   # In practice, the user_id is the id of the current user and therefore this method marks the current's user
   # notifications as read.
   def mark_new_notifications_as_read(user_id)
-    activities.each do |activity|
-      unread_notifications = WorkActivityNotification.where(user_id:, work_activity_id: activity.id, read_at: nil)
-      unread_notifications.each do |notification|
-        notification.read_at = Time.now.utc
-        notification.save
-      end
-    end
```
hectorcorrea, line 301: This was culprit number 1. We iterated over all activities for the work and then we issued another SQL SELECT to fetch unread notifications, so even when there were no unread notifications we were iterating individually. The new code does both at once (SQL SELECT and UPDATE) in batches of 1000 records.

app/views/works/_work_activity_provenance.html.erb
```diff
@@ -4,12 +4,15 @@
     No activity
   <% end %>
   <ul class="beads">
-    <% @work_decorator.changes.sort_by(&:created_at).each do |activity| %>
+    <% @work_decorator.changes.take(100).each_with_index do |activity| %>
```
hectorcorrea, line 7: This was culprit number 2. The code to display each notification issues a SQL SELECT to get the user information, even with Rails using its internal cache 1000 individual SQL SELECTS are slow. For now we are displaying only the first 100 notifications (if there are more they can contact us and we'll look closer)

leefaisonr (approved):

Issue #1984, opened 2024-11-04
How can I utilize capistrano to take a machine off the loadbalancer


carolyncole:
When we want to be certain a machine does not get served to the public, even when the rails environment is working, can we utilize capistrano machine off line

- [x] Create a custom health endpoint that looks at a database value (possibly flipflop)
- [ ] Create a rake task to change the value on/off (does flipflop already have these?)
- [ ] Create a capistrano task to change the value
- [ ] Verify that the loadbalancer believes the site is down and then up based on the cap task

carolyncole:
https://github.com/voormedia/flipflop/blob/master/lib/tasks/flipflop.rake Flip flop has rake tasks!

tpendragon:
@carolyncole did it! We've distributed it everywhere, closing this.

PR #1985, opened 2024-11-04
Utilize FileAbsenceCheck to force the loabalancer to remove a host

carolyncole:
refs #1984


config/initializers/health_monitor.rb
```diff
@@ -1,20 +1,26 @@
 # frozen_string_literal: true
-HealthMonitor.configure do |config|
-  config.cache
-  config.redis
 
-  # Make this health check available at /health
-  config.path = :health
+Rails.application.config.to_prepare do
```
carolyncole, line 3: This is explained: https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#use-case-1-during-boot-load-reloadable-code

app/models/file_check_failure.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+class FileCheckFailure < HealthMonitor::Providers::Base
```
carolyncole, line 2: FileAbsenceCheck

app/models/file_check_failure.rb
```diff
@@ -0,0 +1,10 @@
...
+  CHECK_FILE = "tmp/remove-from-nginx"
```
carolyncole, line 3: Make the file configurable

carolyncole: Needs to be in public/remove-from-nginx to survive deploys

config/deploy.rb
```diff
@@ -79,6 +79,28 @@
       end
     end
   end
+
+  # You can/ should apply this command to a single host
+  # cap --hosts=pdc-describe-staging1.princeton.edu staging application:remove_from_nginx
```
maxkadel, line 84: Would it be worthwhile/possible to add a check that the host parameter has been passed (like we do with the nginx playbook) to make sure you don't accidentally take both/all machines down?

carolyncole: Is it not ok to take everything off line? I guess I would agree I would not normally want to... I'm not entirely sure how to do it, but I can look into it...

maxkadel: Just a thought, definitely doesn't need to be a blocker!

carolyncole:
refs https://github.com/lbeder/health-monitor-rails/pull/127

bess (approved):

PR #1989, opened 2024-11-15
Significantly reduce load time on the Show page for large datasets

### LINKED ISSUE #1623: Works with more than 10,000 files
Currently when a work has more than 10,000 files the user interface struggles to display and load the information. 

As a stop gap measure we should update the application to display a banner indicating something along the lines of "This work has more than 10,000 files, file list is not available." and display the rest of the metadata and allow edits to it as normal.

hectorcorrea:
Although we already loaded the file list via AJAX on the Show page, we still had several calls to fetch the file list as we rendered the page (e.g. to calculate and show the total file size.) For large works (e.g. works with 60K files) this calculation takes 30+ seconds which mean the page took 30+ seconds to load to the user.

In this PR I moved all the code that depends on the file list to display those values after the page has been rendered and the file list has been fetched (via AJAX). 

While troubleshooting this issue I also noticed that our code to detect changes in the Upload Snapshots was incredibly slow on this kind of large datasets. Comparing 60,000 files 60,000 times for deletes + adds/modifications was very slow. Even thought this work is done in the background our implementation would take minutes to complete. So I updated the code that calculates the deletes + adds/modifications to use a binary search (instead of a regular Array `include?`) and that improved performance significantly (3-4 seconds instead of 1+ minute). 

Closes #1623 

This has been deployed so Staging and it can be tested with this work that has 40K files: https://pdc-describe-staging.princeton.edu/describe/works/650


app/controllers/works_controller.rb
```diff
@@ -415,5 +415,24 @@ def migrated?
 
       params[:submit] == "Migrate"
     end
+
+    # Returns a hash object that can be serialized into something that DataTables
+    # can consume. The `data` elements includes the work's file list all other
+    # properties are used for displaying different data elements related but not
+    # directly on the DataTable object (e.g. the total file size)
+    def file_list_ajax_response(work)
```
hectorcorrea, line 423: Notice how we return the file list plus a bunch of other values that depend on the file list

app/views/works/_s3_resources.html.erb
```diff
@@ -75,7 +68,18 @@
       select: true,
       ajax: {
         url: fileListUrl,
-        dataSrc: ''
+        dataSrc: function(json) {
+          // The JSON payload includes a few extra properties (fetched via AJAX because
+          // they be slow for large datasets). Here we pick up those properties and update
+          // the display with them.
+          $("#total-file-size").text(json.total_size_display);
```
hectorcorrea, line 75: We update the page with the extra values that now come on the AJAX response

app/models/upload_snapshot.rb
```diff
@@ -6,9 +6,15 @@ class UploadSnapshot < ApplicationRecord
...
+    s3_filenames_sorted = s3_filenames.sort
     existing_files.each do |file|
       filename = file["filename"]
-      unless s3_filenames.include?(filename)
+      # Use Ruby's Binary Search functionality instead of a plain Ruby Array `.include?`
+      # to detect missing values in the array because the binary search performs
+      # much faster when the list of files is large. Notice that the binary search
+      # requires that the list of files is sorted.
+      # See https://ruby-doc.org/3.3.6/bsearch_rdoc.html
+      if s3_filenames_sorted.bsearch { |s3_filename| filename <=> s3_filename }.nil?
```
hectorcorrea, line 17: This is the other optimization, notice the use of Binary Search `bsearch` instead of a normal `include?`

app/models/work.rb
```diff
@@ -629,5 +638,14 @@ def embargo_date_as_json
         embargo_date_iso8601.gsub(/\+.+$/, "Z")
       end
     end
+
+    def log_performance(start, message)
+      elapsed = Time.zone.now - start
+      if elapsed > 20
```
hectorcorrea, line 644: We can tweak the threshold, but I thought it would be good to start logging as warning slow requests

app/controllers/works_controller.rb
```diff
@@ -415,5 +415,24 @@ def migrated?
...
+      files = []
+      total_size = 0
+      unless work.nil?
+        files = work.file_list
+        total_size = work.total_file_size_from_list(files)
+      end
+      {
+        data: files,
+        total_size:,
+        total_size_display: ActiveSupport::NumberHelper.number_to_human_size(total_size),
```
jrgriffiniii, line 433: 👍

app/models/upload_snapshot.rb
jrgriffiniii, line 17: 💯

app/models/work.rb
```diff
@@ -349,17 +351,24 @@ def file_list
...
+    file_list.each do |file|
+      total_size += file[:size]
     end
+    total_size
+  end
+
+  # Calculates the total file size from a given list of files
+  # This is so that we don't fetch the list twice from AWS since it can be expensive when
+  # there are thousands of files on the work.
+  def total_file_size_from_list(files)
```
jrgriffiniii, line 370: 👍

jrgriffiniii (approved):

PR #2006, opened 2025-01-07
Fixes flaky test

### LINKED ISSUE #2002: Flaky test in CI
Test runs are failing intermittently on this error: 

Screenshot 2024-12-18 at 7 57 08 AM
Screenshot 2024-12-18 at 7 56 50 AM

hectorcorrea:
Fixes the flaky test for the Work Wizard.

I am not entirely sure how come this test passed sometimes but not others. If I am reading the spec correctly *it should had always failed* since (1) it does not add a data file to the work (it only adds the readme) and (2) the very last check for the "title" should always fail given that we don't display the file on the last step. ¯_(ツ)_/¯

Closes #2002

### DIFF
--- a/spec/system/work_wizard_spec.rb
+++ b/spec/system/work_wizard_spec.rb
@@ -83,6 +83,8 @@ describe "walk the wizard hitting all the buttons", type: :system, js: true do
 
     click_on "Next"
     expect(page).to have_css(validate_form_css)
+    # Force the work to have two files (readme + another file)
+    stub_s3 data: [FactoryBot.build(:s3_readme, work:), FactoryBot.build(:s3_file, work:)]
 
     click_on "Previous"
     expect(page).to have_css(file_upload_form_css)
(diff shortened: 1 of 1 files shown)


spec/system/work_wizard_spec.rb
```diff
@@ -96,7 +98,6 @@
     click_on "Grant License and Complete"
     page.driver.browser.switch_to.alert.accept
     expect(page).to have_content("5-10 business days")
-    expect(page).to have_content(work.title)
```
hectorcorrea, line 99: We don't display the title in https://github.com/pulibrary/pdc_describe/blob/main/app/views/works_wizard_submission_complete/show.html.erb (but how come this passed sometimes?)

spec/system/work_wizard_spec.rb
```diff
@@ -83,6 +83,8 @@
 
     click_on "Next"
     expect(page).to have_css(validate_form_css)
+    # Force the work to have two files (readme + another file)
+    stub_s3 data: [FactoryBot.build(:s3_readme, work:), FactoryBot.build(:s3_file, work:)]
```
hectorcorrea, line 87: This seems to be the culprit. We are not adding a data file (other than the readme) and the code expects one. That's why the validation error in CI is actually indicating. 

What I still don't understand is how come this passed sometimes but not others?

bess (approved):

PR #2032, opened 2025-02-21
Allow admin to remove users from groups on the Groups edit page

### LINKED ISSUE #2025: Error removing user from moderators list
Trying to remove a user from the moderators list from a group does not work. There is a JavaScript error reported on the browser's console but the user does not see any error, the delete option fails silently for them.

Image

carolyncole:
Do not remove all permissions, but remove based on role requested

fixes #2025

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -139,7 +139,10 @@ class Group < ApplicationRecord
     end
   end
 
-  def delete_permission(current_user, removed_user)
+  # @param [User] current_user - the currently logged in user
+  # @param [User] admin_user - the user who will be granted admin rights on this group
+  # @param [String] role to be deleted
+  def delete_permission(current_user, removed_user, role)
     if current_user.has_role?(:super_admin) || current_user.has_role?(:group_admin, self)
       if removed_user.nil?
         errors.add(:delete_permission, "User was not found")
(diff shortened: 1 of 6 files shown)


app/views/groups/edit.html.erb
```diff
@@ -89,14 +89,10 @@
...
-  ...pdc,
-  deleteUserFromGroupUrl: '<%= delete_user_from_group_url(id: @group.id, uid: "uid-placeholder") %>',
-  formAuthenticityToken: '<%= form_authenticity_token %>',
-  groupId: '<%= @group.id %>',
-  addSubmitterUrl: '<%= add_submitter_url(id: @group.id, uid: "uid-placeholder") %>',
-  addAdminUrl: '<%= add_admin_url(id: @group.id, uid: "uid-placeholder") %>',
-  userPath: "<%= user_path('user-placeholder') %>"
-}
-</script>
+<input type="hidden" id="deleteUserFromGroupUrl" name="deleteUserFromGroupUrl" value="<%= delete_user_from_group_url(id: @group.id, uid: "uid-placeholder") %>" />
```
carolyncole, line 92: This is the change that made the JS constantly work.  From what I could see the global variable sometimes had the correct information and sometimes did not.  Changing to a hidden input made the information consistently present.

app/models/group.rb
```diff
@@ -139,16 +139,18 @@ def add_submitter(current_user, additional_user)
...
     if current_user.has_role?(:super_admin) || current_user.has_role?(:group_admin, self)
       if removed_user.nil?
         errors.add(:delete_permission, "User was not found")
       elsif removed_user == current_user
         errors.add(:delete_permission, "Cannot remove yourself from a group. Contact a super-admin for help.")
       else
         errors.delete(:delete_permission)
-        removed_user.remove_role :group_admin, self
-        removed_user.remove_role :submitter, self
+        removed_user.remove_role role, self
```
carolyncole, line 153: For some odd reason no matter which delete you chose in the UI, the back end was removing the group admin and the submitter

hectorcorrea (approved):

Issue #2026, opened 2025-02-10
Permission enforcement when viewing datasets


hectorcorrea:
It looks like we need to enforce some logic to decide who can view a dataset in PDC Describe. We currently enforce authorization on editing but viewing is not as tight.

For example the `/works` URL shows a list of all the works. Likewise, going directly to the URL for a given dataset allows the user to view it (even if it does not allow the unauthorized user to edit it.)

It looks like the Dashboard enforce some logic to decide what dataset the user sees, but it's unclear whether it does the proper checks.

I believe we implemented authorization by groups (e.g. users that are submitters to the PPPL group can see datasets for the PPPL group but not for the PRDS group). We should revisit this logic.

hectorcorrea:
We also had to account for `/works.rss` since that page needs to remain available to the public because it's what PDC Discovery uses to get the list of datasets that need to be ingested.

PR #2048, opened 2025-03-14
Only allow for Super Admin users to access /works url path

leefaisonr:
ref #2026

### DIFF
--- a/app/views/works/index.rss.builder
+++ b/app/views/works/index.rss.builder
@@ -6,9 +6,7 @@ xml.rss version: "2.0" do
     xml.description "Princeton Data Commons RSS Feed of approved submissions"
     xml.link root_url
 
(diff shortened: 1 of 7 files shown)


app/views/works/index.rss.builder
```diff
@@ -6,9 +6,7 @@ xml.rss version: "2.0" do
     xml.description "Princeton Data Commons RSS Feed of approved submissions"
     xml.link root_url
 
-    @works.each do |work|
-      # Only include approved works in the RSS feed
-      next unless work.state == "approved"
+    @approved_works.each do |work|
```
hectorcorrea, line 9: We moved this logic to the controller to make it a bit more visible

hectorcorrea (approved):

PR #2052, opened 2025-03-17
Handle File Upload errors when the Load Balancer blocks the request

### LINKED ISSUE #2012: Error handling for file uploads
While troubleshooting issue #2009 we noticed that when there is a firewall/NGINX issue uploading a file the error is reported in the wrong place. This is how it is reported when the problem is uploading the README:

upload-error-js

The problem is that when this happen the Rails application does not even get the request and it is NGINX that reports back the error and it does it as an HTML page with status code 200 (see details in #2009). We could handle these errors by updating the JavaScript that process the response.

The [JavaScript code for the README](https://github.com/pulibrary/pdc_describe/blob/main/app/javascript/entrypoints/pdc/work_readme_file_upload.es6) is this:

```
uppy.use(Uppy.XHRUpload, {
      endpoint: uploadUrl,
      headers: { 'X-CSRF-Token': tokenContent },
      bundle: true, // upload all selected files at once
      formData: true, // required when bundle: true
      getResponseData(filename) {
        $('#new-readme').html(`File <b>${filename}</b> has been uploaded and set as the README for this dataset.`);
        $('#readme-upload').prop('disabled', false);
      },
    });
```

Notice how this code expected a filename but since the load balancer returned the error message instead we dump that message on the form.

The file upload for files other than the readme (https://github.com/pulibrary/pdc_describe/blob/main/app/javascript/entrypoints/pdc/work_edit_file_upload.es6) has the same problem except in that code we never display the error.

Acceptance Criteria

- [ ] Double check response to make sure it does not include "your Support ID" and if it does make that as an error for Uppy
- [ ] Happens both in the Readme and the file upload page.  Only the readme shows the error to the user.

hectorcorrea:
Detects when we get an error from the Load Balancer uploading a file. 

Closes #2012

### DIFF
--- a/app/javascript/entrypoints/pdc/work_edit_file_upload.es6
+++ b/app/javascript/entrypoints/pdc/work_edit_file_upload.es6
@@ -63,10 +63,46 @@ export default class WorkEditFileUpload {
       headers: { 'X-CSRF-Token': tokenContent },
       bundle: true, // upload all selected files at once
       formData: true, // required when bundle: true
-      getResponseData() {
-        // Reload the file list displayed
-        const fileTable = $('#files-table').dataTable();
-        fileTable.api().ajax.reload();
+      getResponseData(serverResponse) {
+        try {
+          // A normal response is a string that can be parsed as JSON
+          response = JSON.parse(serverResponse)
+          // Reload the file list displayed
+          const fileTable = $('#files-table').dataTable();
+          fileTable.api().ajax.reload();
+        } catch (ex) {
+
+          // Display the error message to the user
+          var errorMessage;
+          if ((serverResponse || "").toLowerCase().includes("your support id")) {
+            // We could not parse the serverResponse and it includes "Your support" ID
+            // we assume the problem is the Load Balancer.
+            errorMessage = "Error uploading file: Our load balancer rejected the request.";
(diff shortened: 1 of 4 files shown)


app/javascript/entrypoints/pdc/work_edit_file_upload.es6
```diff
@@ -63,10 +63,28 @@ export default class WorkEditFileUpload {
       headers: { 'X-CSRF-Token': tokenContent },
       bundle: true, // upload all selected files at once
       formData: true, // required when bundle: true
-      getResponseData() {
-        // Reload the file list displayed
-        const fileTable = $('#files-table').dataTable();
-        fileTable.api().ajax.reload();
+      getResponseData(filenames) {
+        const loadBalancerError = (filenames || '').toLowerCase().includes('your support id');
```
hectorcorrea, line 67: This line is the check. If the Load Balancer blocks the request we detect it because the "filenames" will include the error message "Your Support ID..."

hectorcorrea:
I could not figure out how to add a test to run the JavaScript and test the display of the message so I included a screenshot instead.

JaymeeH (approved):

PR #2057, opened 2025-03-19
Adds a custom health check for the Rails cache

### LINKED ISSUE #2014: [pdc_describe/production] HealthMonitor::Providers::CacheException: different values (now: 2025-01-16 13:24:44 +0000, fetched: 2025-01-16 13:24:45 +0000)
This happens in several of our applications and causes noise in Honeybadger.  We should address this at some point to reduce noise.

## Backtrace

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/99214/faults/115552105)

## Acceptance criteria
- [ ] Change the threshold to ignore if it is less than 10 seconds
- [ ] Hopefully change the check
- [ ] If not just don't tell honey badger

hectorcorrea:
Copied @carolyncole 's implementation of the Redis cache (https://github.com/pulibrary/pdc_describe/pull/2056) to workaround the bug in the default HealthMonitor check.

Closes #2014

### DIFF
--- a/app/services/pul_cache.rb
+++ b/app/services/pul_cache.rb
@@ -1,25 +1,28 @@
 # frozen_string_literal: true
 
-require 'health_monitor/providers/cache'
+require "health_monitor/providers/cache"
 
 class PULCacheException < StandardError; end
 
 class PULCache < HealthMonitor::Providers::Cache
   def check!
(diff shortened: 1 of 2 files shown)


carolyncole (approved):
looks good @hectorcorrea

Issue #2054, opened 2025-03-18
Improve automated messages for submitters


matthewjchandler:
### User story

As a submitter to PDC Describe, I need timely and accurate messages, so that I can take action as needed and be assured when no further action is needed.

### Context

Currently, the messages coming from PDC Describe regularly confuse and frustrate submitters, which the curators in PRDS manage with follow-up messages that correct the automated messages. It's not a good system.

The biggest recurring issues are that approval messages go out to submitters before datasets are actually publicly available and reverting a submission to draft re-sends the new submission message.

### Acceptance criteria

- [x] When a submitter begins a brand-new submission, send the new submission message the same as before
- [x] When a curator/moderator approves a submission, DELAY sending the approval message to the submitter by [90 minutes
](https://github.com/pulibrary/pdc_describe/blob/063211f82f431d7873dbcbd5f869e45548ab31e8/app/models/work_activity_notification.rb#L11)
- [x] When a curator/moderator returns a submission to draft:
  - [x] PDC does NOT send the new submission message to the submitter again
  - [x] Instead, PDC sends a message to the submitter in which:
    - [x] The subject line is "[pdc-describe] Submission Returned"
    - [x] The heading within the message body is "PDC Describe: Submission Returned"
    - [x] The rest of the auto-generated message is structured as follows:
      > Hello {submitter_name},
You have a new notification for your account with PDC Describe:
{moderator_name} ({moderator_id}) at {time_stamp} returned the following submission to you for revision:
{submission_title}
To view the notification, please browse {work_url}. Thank you very much for your support and time.
Gratefully, Research Data and Scholarship Services Princeton University Library

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/gh/pulibrary/pdc_describe/2054?planning-poker) with Zenhub @hectorcorrea @JaymeeH @jrgriffiniii @leefaisonr

PR #2062, opened 2025-03-24
Improve Automated Messages

JaymeeH:
Adding a delay to the work approval notifaction email 
Adding a new notificaiton email for the rejection of a work
ref #2054

### DIFF
--- a/app/mailers/notification_mailer.rb
+++ b/app/mailers/notification_mailer.rb
@@ -19,4 +19,22 @@ class NotificationMailer < ApplicationMailer
 
     mail(to: @user.email, subject: @subject)
   end
+
+  def reject_message
+    @user = params[:user]
+    @work_activity = params[:work_activity]
+
+    @subject = "[pdc-describe] Submission Returned"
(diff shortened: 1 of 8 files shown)


carolyncole (changes requested):
@JaymeeH I would love for there to be some tests around this code.  Additionally It might be easier to understand if you utilized the from state in both cases WorkActivityNotification AND WorkStateTransitionNotification.

For example. I might add a wait_time method
```
   def wait_time
       if work.state == "approved"
           90.minute
        else
           10.seconds
        end
     end
```
And then use that method to simplify your code a bit
```
mailer = NotificationMailer.with(user:, work_activity:)
work = work_activity.work
if  work.state == "draft" && work.from_state == "awaiting_approval"
  reject_message = mailer.reject_message
  reject_message.deliver_later(wait: wait_time) unless Rails.env.development?
else
  message = mailer.build_message
  message.deliver_later(wait: wait_time) unless Rails.env.development?
end
```

Issue #2053, opened 2025-03-18
Improve automated messages for curators/moderators


matthewjchandler:
### User story

As a curator and/or moderator for PDC, I need action-oriented messages to be highlighted, so that I can keep track of my assigned tasks and respond quickly when needed.

### Context

Currently, the messages coming from PDC Describe all look very similar. The subject heading is always "[pdc-describe] New Notification" and the header inside the message is always "PDC Describe." For messages sent to curators, within the message body are some details on the submitter and the submission, but the difference between an FYI that a submission has been created and an actionable message that a submission is ready for review is just a few words buried within the body of the message.

![Image](https://github.com/user-attachments/assets/95aec1b2-3d20-41d2-9ddb-4c698377833f)

![Image](https://github.com/user-attachments/assets/010d405c-4b79-42bc-8215-6573df57aadb)

This is a recurring pain point for curators. They need better messaging, especially when action is required of them.

### Acceptance criteria

- [x] When a brand-new submission has been created, PDC sends a message to the relevant curators/moderators in which:
  - [x] The subject line is "[pdc-describe] New Submission Created"
  - [x] The heading within the message body is "PDC Describe: New Submission Created"
  - [x] The rest of the auto-generated message remains as-is
- [x] When a submitter completes a submission, PDC sends a message to the relevant curators/moderators in which:
  - [x] The subject line is "[pdc-describe] Submission Ready for Review"
  - [x] The heading within the message body is "PDC Describe: Submission Ready for Review"
  - [x] The rest of the auto-generated message remains as-is
- [x] When a curator/moderator reverts a submission to draft:
  - [x] PDC does NOT send the new submission message to the relevant curators/moderators again
  - [x] Instead, PDC sends a message to the relevant curators/moderators in which:
    - [x] The subject line is "[pdc-describe] Submission Returned"
    - [x] The heading within the message body is "PDC Describe: Submission Returned"
    - [x] The rest of the auto-generated message is structured as follows:
      > Hello {receiver_name},
You have a new notification for your account with PDC Describe:
{moderator_name} ({moderator_id}) at {time_stamp} returned the following submission to the submitter:
{group}, {submitter_id} {submission_title}
To view the notification, please browse {work_url}. Thank you very much for your support and time.
Gratefully, Research Data and Scholarship Services Princeton University Library

carolyncole:
[...]

PR #2066, opened 2025-03-27
Improving messges for curators/moderators

JaymeeH:
Adding new submission and review message for curators
ref #2053

### DIFF
--- a/app/models/work_activity_notification.rb
+++ b/app/models/work_activity_notification.rb
@@ -9,7 +9,7 @@ class WorkActivityNotification < ApplicationRecord
       mailer = NotificationMailer.with(user:, work_activity:)
       work = work_activity.work
       delay = wait_time
-      from_state = check_from_state
+      from_state = check_from_state(work)
 
       if work.state == "draft" && from_state == :none # draft event
         new_submission_message = mailer.new_submission_message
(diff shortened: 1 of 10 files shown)


app/models/work_activity_notification.rb
```diff
@@ -10,9 +10,15 @@ class WorkActivityNotification < ApplicationRecord
       work = work_activity.work
       delay = wait_time
 
-      if work.state == "draft" && work.aasm.from_state == "awaiting_approval"
+      if work.state == "draft" && work.aasm.to_state == "none" # new_submission event
+        new_submission_message = mailer.new_submission_message
+        new_submission_message.deliver_later(wait: delay) unless Rails.env.development?
+      elsif work.state == "draft" && work.aasm.from_state == "awaiting_approval" # revert_to_draft event
         reject_message = mailer.reject_message
```
hectorcorrea, line 17: I have not been able to trigger this condition (even after rejecting a dataset). Let's review to make sure this logic is in the right place.

JaymeeH: I think the issue is with the way the `from_state` & `to_state` are being accessed. I can push a change shortly

app/models/work_activity_notification.rb
```diff
@@ -31,6 +38,17 @@ def wait_time
...
+      case work_activity.message
+      when /has been created/
+        :none
+      when /for revision/
+        :awaiting_approval
+      when /ready for review/
+        :draft
+      end
+    end
+
```
hectorcorrea, line 51: This logic works but it's a bit brittle, I would suggest grabbing the "from" state from the database, for example with code like this:

```suggestion
    def check_from_state(work) # check the previous state of the work
      work_states = UserWork.where(work_id: work.id).sort_by(&:created_at).reverse
      if work_states.count > 1
        # states[0] is the current state, [1] is the previous state
        work_states[1].state.to_sym
      else
        :none
      end
    end
```

hectorcorrea (approved):
Looks great! I left a suggestion for the "check_from_state" code.

hectorcorrea (approved):
Great work!

Issue #1925, opened 2024-09-12
Funding information is removed upon clicking the Save button


astrochun:
### Expected behavior
We were working with a researcher and upon reviewing the metadata we noticed that it was removed. I did it again separately and it happened. The log shows it being added and then removed. See screenshot.

This happened for: https://datacommons.princeton.edu/describe/works/523

### Actual behavior
![Screen Shot 2024-09-12 at 1 14 51 PM](https://github.com/user-attachments/assets/2817a0a2-35f4-4ea1-b71e-96c73089bcfb)

### Steps to replicate
The problem might have been caused by it being initially in PRDS and then migrated to PPPL.

hectorcorrea:
We could not reproduce this. We created a new dataset and modified various fields (including the funders) and the data was saved correctly in all cases. We even tried just tabbing through the fields without changing them to see if that would trigger the error to no avail.

Please let us know if this happens again and/or if there is a specific set of steps to make it happen.

hectorcorrea:
Yesterday the issue with the funders acting weird was reported again in https://github.com/pulibrary/pdc_describe/issues/2009 so I am re-opening this issue:

```
On the page 'Additional Metadata', additional funding references and related identifiers are not 
being saved after entering those, i.e. after hitting save, the new entries do not show up. The 
same applies to identifiers. One needs to add another (not needed) entry and then save in 
order for the previous entry to be saved.
```

carolyncole:
It looks like there were three simultaneous editors Promise, Chun and Sarah.  I wonder if the system got confused by updates from multiple people.
<img width="530" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNVJkQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--09a4d7da13c10e80d6e860fc7940459ec336d91c/Screenshot%202025-03-21%20at%2011.52.41%E2%80%AFAM.png" alt="Screenshot 2025-03-21 at 11.52.41 AM.png" />

carolyncole:
@astrochun I'm wondering if we should have some sort of check to see if the page was loaded before the last save and warn the user if they are submitting changes that could overwrite other changes?

hectorcorrea:
@carolyncole this is a great idea. I will go ahead and _at least_ log when this happens.

hectorcorrea:
[...]

PR #2067, opened 2025-03-28
Logs if a dataset updates will overwrite previous updates

hectorcorrea:
This does not fix the bug where funder information was removed during an update but *it will* help us determine if when saving a dataset the data will overwrite data written before. 

If we determine that the problem is users stepping on each other (or even the same user with two browser windows open over writting their own changes) we can decide how to handle this (e.g. warn a user that they will overwrite some changes)

Part of #1925 

This is an example of a Honeybadger entry for this kind of situation: https://app.honeybadger.io/projects/99214/faults/118747777

### DIFF
--- a/spec/controllers/works_controller_spec.rb
+++ b/spec/controllers/works_controller_spec.rb
@@ -1027,7 +1027,7 @@ RSpec.describe WorksController do
         context "when sending a nil group" do
           before do
             params[:group_id] = nil
-            params[:last_updated_at] = work.updated_at.to_s # Make sure we don't trigger the Honeybadger to stale data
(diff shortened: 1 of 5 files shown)


spec/controllers/works_controller_spec.rb
```diff
@@ -951,6 +951,28 @@
           end
         end
 
+        context "when updating a record" do
+          let(:form_params_stale) do
+            # Pretent the data submitted on the web form is older
```
carolyncole, line 956:
```suggestion
            # Pretend the data submitted on the web form is older
```

carolyncole (dismissed):
@hectorcorrea one tiny spelling mistake, otherwise looks great!  Thanks!

PR #2075, opened 2025-04-03
Update Capistrano to fix issue deploying

carolyncole:
error 'uninitialized constant Capistrano::SCM::Git::StringIO (NameError)' when deploying main

For some reason ansible tower still worked just fine...


JaymeeH (approved):
Thanks @carolyncole

PR #2084, opened 2025-04-07
Removing support and references for Globus URLS

### LINKED ISSUE #2037: Remove globus_uri_from_description
### What maintenance needs to be done?

There should not be any urls for globus in the decriptions.  Remove the code that creates the globus button when there is.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?
- clean up unused code
- Do not highlight the old globus location if it is present

### Acceptance criteria

 - [ ] If this change touches any of the files referenced in `.rubocop_todo.yml` create a new ticket to address the rubocop  todo.
 - [ ] The code has been removed from app/models/solr_document
 - [ ] All tests still pass

jrgriffiniii:
Resolves #2037

### DIFF
--- a/app/models/s3_file.rb
+++ b/app/models/s3_file.rb
@@ -48,11 +48,6 @@ class S3File
     size == 0
   end
 
-  def globus_url
-    encoded_filename = filename.split("/").map { |name| ERB::Util.url_encode(name) }.join("/")
-    File.join(Rails.configuration.globus["post_curation_base_url"], encoded_filename)
-  end
-
   delegate :s3_query_service, to: :@work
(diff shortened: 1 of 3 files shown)


hectorcorrea (changes requested):
Wait...I don't think we want this. I think this will blow the URL for each file that we expose in Discovery.

hectorcorrea:
@jrgriffiniii I don't think we need to implement this ticket because we don't want to remove the Globus URL for each of the files that we expose in Discovery. That URL is how users download the files:

Screenshot 2025-04-07 at 3 43 21 PM

PR #2087, opened 2025-04-08
Adds a method to find released embargo in a standard fashion

carolyncole:
released are approved works with an embargo day of yesterday 
embargoed are currently embargoed today or in the future

Standardized the functions to use sql current_date function.


hectorcorrea (approved):
🆒

PR #2094, opened 2025-04-10
Move files to embargo bucket for approved embargoed works

hectorcorrea:
Moves files to the the embargo bucket (instead of post-curation) for embargoed works.

Closes #2055 

Co-authored-by: Robert-Anthony Lee-Faison <leefaisonr@users.noreply.github.com>

PR #2104, opened 2025-04-15
Log the correct location in the work activity for embargoed works

hectorcorrea:
Fixes a message that we missed in PR https://github.com/pulibrary/pdc_describe/pull/2094

### DIFF
--- a/app/models/approved_upload_snapshot.rb
+++ b/app/models/approved_upload_snapshot.rb
@@ -4,7 +4,8 @@ class ApprovedUploadSnapshot < BackgroundUploadSnapshot
     new_files.each do |file|
       work.track_change(:added, file["filename"])
     end
-    WorkActivity.add_work_activity(work.id, "#{new_files.count} #{'file'.pluralize(new_files.count)} were moved to the post curation bucket", new_files.first["user_id"],
+    target_location = work.embargoed? ? "embargo" : "post-curation"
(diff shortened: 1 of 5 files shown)


app/models/approved_upload_snapshot.rb
```diff
@@ -4,7 +4,8 @@ def finalize_upload
     new_files.each do |file|
       work.track_change(:added, file["filename"])
     end
-    WorkActivity.add_work_activity(work.id, "#{new_files.count} #{'file'.pluralize(new_files.count)} were moved to the post curation bucket", new_files.first["user_id"],
+    target_location = work.embargoed? ? "embargo" : "post-curation"
```
carolyncole, line 7: @hectorcorrea can this code utilize `work.files_mode_human`?

app/models/work.rb
```diff
@@ -539,7 +539,7 @@ def embargoed?
   # Returns a human friendly name for the bucket where the files for the work are located.
   # Notice that we don't use the values from PULS3Client because those are not human friendly
   # (e.g. the lack dashes between words)
-  def files_bucket_name_human
+  def files_mode_human
```
carolyncole, line 542: ❤️

app/models/approved_upload_snapshot.rb
hectorcorrea: @carolyncole sadly no, the work state is not yet approved here so I had to just check the embargo here (and not the state)

carolyncole (approved):
Thanks @hectorcorrea

PR #2109, opened 2025-04-17
Fixes problem adding submitters that are not Users

### LINKED ISSUE #2107: Error adding new submitters to a group
Anya reported that she had issues adding new submitters to the PPPL group. 

Around the same time Honeybadger reported an error related to a problem trying to add user to with a duplicated email: https://app.honeybadger.io/projects/99214/faults/119506666/01JRZSB3SYR5749VKZC2GWVGRY?page=0

```
ActiveRecord::RecordNotUnique: PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint
"index_users_on_email" DETAIL:  Key (email)=() already exists.
```

It turns out we are trying to add the actual user when a new submitter is added if the `netid` entered is for a new user. This is causing problems because when this happens we add users to the User table without an email address. The very first time this works (even though it shouldn't) but as soon as a second user is added the "duplicated email" error is raised because we cannot have two users with an empty email. 

The current workaround is for Anya to ask the users to login to PDC Describe so that a User record is created for them with their actual email from CAS. Once their user record has been created adding them as submitters works because the User record already exists. This is awkward because from the user's perspective is pointless to login to PDC Describe if they cannot submit to anything.

hectorcorrea:
Prevent adding users without emails and triggering database constraint errors.

Closes #2107

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -213,19 +213,7 @@ class Group < ApplicationRecord
   def default_user(uid)
     user = User.find_by(uid:)
     return user if user.present?
-
-    begin
-      user = User.new(uid:, default_group_id: id)
-      user.save!
-      user
-    rescue ActiveRecord::RecordNotUnique => unique_error
-      # If adding a submitter to an existing group
-      Rails.logger.error("Failed to created a new user for #{self}: #{unique_error}")
-      user = User.new(uid:)
-      user.default_group_id = id
-      user.save!
-      user
-    end
+    User.new_for_uid(uid, id)
   end
 end
 # rubocop:enable Metrics/ClassLength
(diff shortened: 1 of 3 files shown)


app/models/group.rb
```diff
@@ -213,19 +213,7 @@ def default_community
...
-      user
-    rescue ActiveRecord::RecordNotUnique => unique_error
-      # If adding a submitter to an existing group
-      Rails.logger.error("Failed to created a new user for #{self}: #{unique_error}")
-      user = User.new(uid:)
-      user.default_group_id = id
-      user.save!
-      user
-    end
+    User.new_for_uid(uid, id)
```
hectorcorrea, line 216: Our existing `new_for_id` method did most of what we needed, I just updated it to accept a default group id and that fixes the problem.

PR #2111, opened 2025-04-17
Fixes bug on Wizard Save

### LINKED ISSUE #2110: Creators not saved when clicking "Save" instead of "Next" in the Wizard
The contributors entered by the user are not saved when the user just clicks "Save" in the "Wizard" instead of clicking "Next".

In the following screenshot notice how the user has selected "Jodi Venables". If the user the "Save" button right now the data for Jodi Venables" will not be saved and instead whatever value was in the contributor box before will be kept.

Image

hectorcorrea:
Closes #2110

### DIFF
--- a/app/controllers/works_wizard_controller.rb
+++ b/app/controllers/works_wizard_controller.rb
@@ -170,6 +170,7 @@ class WorksWizardController < ApplicationController
 
   private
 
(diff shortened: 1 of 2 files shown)


carolyncole (approved):
Thanks @hectorcorrea!

Issue #2127, opened 2025-05-09
Remove Data Dog


kelynch:
### What maintenance needs to be done?
Remove PDC Describe from Data Dog. 

### Level of urgency
- [x] High
- [ ] Moderate
- [ ] Low

### Why is this maintenance needed?
Data Dog is expensive and we aren't really using it. 

### Acceptance criteria

 - [ ] All PDC Discovery systems are removed from Data Dog

Referenced in https://github.com/pulibrary/princeton_ansible/issues/6126

PR #2129, opened 2025-05-13
Remove DataDog

bess:
Ref #2127


hectorcorrea (approved):
Deployed to staging with no issues.

Issue #2254, opened 2026-04-15
Upgrade ruby & rebuild prod


bess:
## Acceptance criteria
- [x] PDC Describe can be rebuilt in production (currently the staging rebuild works but the prod one does not)

PR #2261, opened 2026-04-20
Fix vite deploy

bess:
Works!
`HOSTS=pdc-describe-prod1.princeton.edu BRANCH=fix_vite_deploy be cap production deploy`

Ref #2254


vite.config.mts
```diff
@@ -1,22 +1,8 @@
 import { defineConfig } from 'vite'
 import RubyPlugin from 'vite-plugin-ruby'
 
-export default ({ command, mode }) => {
```
carolyncole, line 4: @bess is this really needed?  It will minify the JS in development and make it harder to debug.

carolyncole (changes requested):
@bess If the change to remove the non minify in development is REALLY needed this CAN go forward, but I would love not to...

vite.config.mts
bess: @carolyncole Good catch! No, that change was not intended, it was a side effect of re-installing vite. I put that file back the way it was, and deploy still works. Thank you!

carolyncole (dismissed):
Thanks @bess

Issue #2204, opened 2026-02-19
draft DOI should go to a separate page indicating that the work is still under review


astrochun:
### User story

One of our data curator stated the following:

> Feedback: It seems that our authors are putting in links to the PDC data sets in their papers, which is great, but the links just show up as dead links until the data set is published. Since there may be some delay between journal publication and PDC data set publication, and the link is already fixed, is it possible to ask the PDC people to change the 'link doesn't exist' error to a 'this data set is being finalized for publication, please come back later' message, or something similar?

### Acceptance criteria

- [ ] DOIs that are drafted go to a page that indicates the above message
- [ ] Incorrect URL will still go to a 404.

### Concrete example

- [ ] https://doi.org/10.34770/hmgp-km09 (Would go to this new pending review stage)
- [ ] https://doi.org/<something_else> (if not an already minted draft DOI, to go to 404)

matthewjchandler:
Does the DOI need to be in the "Registered" state (not "Draft") for this to work?

matthewjchandler:
> Does the DOI need to be in the "Registered" state (not "Draft") for this to work?

Yes

matthewjchandler:
Would the "Discovery URL" in the Describe work change to go to the landing page, too?

astrochun:
> Would the "Discovery URL" in the Describe work change to go to the landing page, too?

I don't think we should have that URL available in Describe. It will allow authors to use it over the DOI, and that's not a best practice. We had this issue a number of times with ARKs and folks using links that are now outdated and will need to be redirected.

matthewjchandler:
> > Would the "Discovery URL" in the Describe work change to go to the landing page, too?
> 
> I don't think we should have that URL available in Describe. It will allow authors to use it over the DOI, and that's not a best practice. We had this issue a number of times with ARKs and folks using links that are now outdated and will need to be redirected.

OK, I made a separate ticket for that: #2274

matthewjchandler:
I double-checked with Chun and the PRDS team, and we all approve of proceeding to set DOIs as "Registered" in DataCite when a submitter/depositor clicks "Create New" on the "New Submission" screen

PR #2283, opened 2026-05-04
Improving documentation for Works index method

leefaisonr:
ref #2204


bess (changes requested):
@leefaisonr I agree with your changes, but this PR title does not describe the changes contained. (I know that's because we changed directions, but it still needs fixing.)

Please fix the PR title so that later when we go back through the change log it'll be clear what change was made in these commits.

Thanks!

bess:
All of these changes were also in the other PR.

PR #2284, opened 2026-05-04
Create an RSS feed to enable harvesting of draft DOIs

leefaisonr:
Co-authored-by: Bess Sadler <bess@users.noreply.github.com>
Co-authored-by: Precilla Prempeh <precillap@users.noreply.github.com>

ref #2204 

image

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -48,13 +48,12 @@ class WorksController < ApplicationController
     end
   end
 
-  # This method allows any user to visit /pending-works.rss
+  # This method allows any user to visit /works/pending.rss
   # and a list is generated for all pending works in an RSS format
   # so that the works are harvestable by PDC Discovery.
   # In order to support landing pages for not yet approved works in PDC Discovery
   # (see https://github.com/pulibrary/pdc_describe/issues/2204) we will add a pending works RSS feed.
   def pending
-    byebug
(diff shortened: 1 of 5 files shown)


app/controllers/works_controller.rb
```diff
@@ -50,6 +48,17 @@ def index
...
 
+  # This method allows any user to visit /works/awaiting-approval.rss
+  # and a list is generated for works with a draft DOI in an RSS format
+  # so that the works are harvestable by PDC Discovery.
+  # In order to support landing pages for not yet approved works in PDC Discovery
+  # (see https://github.com/pulibrary/pdc_describe/issues/2204).
+  def awaiting_approval
+    if rss_awaiting_approval_request?
+      rss_awaiting_approval_index
+    end
```
carolyncole, line 63:
```suggestion
    else
      # If a user who is not a super admin attempts to visit /awaiting_approval they will be redirected to the root_path and receive a flash notice that they do not have access to that page.
      flash[:notice] = "You do not have access to this page."
      redirect_to root_path
    end
```

carolyncole (changes requested):
I'm just noting that the system returns nothing if they are not trying to get a rss feed.  Possibly use the code I suggested to include some kind of response?

carolyncole (approved):
Thanks @leefaisonr

Issue #2299, opened 2026-05-13
Messages addressed to curators/moderators incorrectly


matthewjchandler:
### Expected behavior

If a message intended for a submitter/depositor is also sent to curators/moderators, the body text should be exactly the same for all recipients. The copy sent to curators/moderators should be like a bcc email.

See #2209

### Actual behavior

Copies of messages sent to curators/moderators are being addressed to them by name by mistake, e.g.:

<img width="1004" height="424" alt="Image" src="https://github.com/user-attachments/assets/3afefbe5-3a29-4608-943c-336642b0917a" />

### Impact of this bug

Curators/moderators are getting confused about what messages are intended for them vs. what is merely a carbon copy, and they are not able to confirm what message exactly was sent to the submitter/depositor.

### Acceptance criteria

- [ ] When a message intended for a submitter/depositor is also sent to curators/moderators, then body text should be exactly the same for all recipients
  - [ ] The copy sent to curators/moderators should behave like a bcc email
- [ ] There is a test for this

PR #2324, opened 2026-06-02
Updating code to allow for multiple WorkActivity classes based on the transition

carolyncole:
We know that we need to be able to change the behavior of the notifications based on what state transition we are going through.

Instead of creating more branches as this gets more complex I decided to create classes for each transition.

You will note that there are very few changes to the tests as this is an effort to reorganize our code and create a seam for #2299.  Some changes were to make the tests more explicit, or just needed based on the organization changes.

Generally Work State Notifications were being handled by the WorkActivityNotification class and every notification that was sent out was based on the user tags set in the WorkStateTransitionNotification.

There are still some states (Withdrawn, Resubmitted, and Deletion Marker) that just use the base ActivityNotification but can easily be updated to implement their own send_message as the need arrises.

### DIFF
--- a/app/models/work_activity_notification.rb
+++ b/app/models/work_activity_notification.rb
@@ -6,6 +6,15 @@ class WorkActivityNotification < ApplicationRecord
 
   after_create do
     if send_message?
+      send_messages
+    else
+      update_email_sent("unsent")
+    end
+  end
+
+  private
+
+    def send_messages
       mailer = NotificationMailer.with(user:, work_activity:)
       # messages sent from the work
       if work_activity.activity_type == WorkActivity::MESSAGE
@@ -15,8 +24,8 @@ class WorkActivityNotification < ApplicationRecord
 
       # state transition notifications
       elsif work_activity.activity_type == WorkActivity::NOTIFICATION
-        message = build_state_transition_message(mailer)
-        message.deliver_later(wait: wait_time) unless Rails.env.development?
+        update_email_sent("state_transition")
+        mailer.build_message.deliver_later(wait: wait_time) unless Rails.env.development?
 
       # Error publishing DOI or Curator self assigned and assigned
       else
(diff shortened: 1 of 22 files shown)


app/models/work_activity_notification.rb
```diff
@@ -15,46 +24,16 @@ class WorkActivityNotification < ApplicationRecord
 
       # state transition notifications
       elsif work_activity.activity_type == WorkActivity::NOTIFICATION
-        message = build_state_transition_message(mailer)
-        message.deliver_later(wait: wait_time) unless Rails.env.development?
+        update_email_sent("state_transition")
```
carolyncole, line 27: Send a message for one of the classes that have no overridden messages

app/models/work_activity_notification.rb
```diff
@@ -15,46 +24,16 @@ class WorkActivityNotification < ApplicationRecord
...
         message.deliver_later
       end
-    else
-      update_email_sent("unsent")
-    end
-  end
-
-  private
-
-    # this method is a bit long but it is handling a lot of different cases and I think it's clearer to have it all in one place rather than trying to break it up into multiple methods
```
carolyncole, line 34: We no longer have two places where we are looking at the state.  Now the WorkStateTrasitionNotification class handles the decision on what notification class gets created to send messages about the state change

app/models/work_state_transition_notification.rb
```diff
@@ -25,32 +25,38 @@ def initialize(work, current_user_id)
...
-    WorkActivity.add_work_activity(id, notification, current_user_id, activity_type: WorkActivity::NOTIFICATION)
+    class_for_transition.add_work_activity(id, current_user_id, user_tags)
   end
 
     private
 
-      def notification_for_transition
+      # rubocop:disable Metrics/MethodLength
+      # I want the factory to be all in one method
+      def class_for_transition
```
carolyncole, line 38: Choose a class not just a message

spec/models/work_activity_notification_spec.rb
```diff
@@ -19,9 +19,9 @@
     end
 
     it "enqueues an e-mail message to be delivered for the notification" do
-      work_activity_notification = described_class.create(user:, work_activity:)
+      work_activity_notification = WorkStateTransition::WithdrawnNotification.create(user:, work_activity:)
```
carolyncole, line 22: These notifications are no longer just the base class and can not just be called.

spec/models/work_state_transition_notification_spec.rb
```diff
@@ -0,0 +1,125 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+describe WorkStateTransitionNotification, type: :model do
```
carolyncole, line 4: Test the factory works as expected

bess (approved):
This refactor is a huge improvement, thank you @carolyncole

PR #2335, opened 2026-06-10
Works that are preserved are really awaiting_approval

carolyncole:
This was a distinction that was not needed util we started chnaging the json based on the state

Also deliver_later without a wait so mailers work in development rails


leefaisonr (approved):
Thanks @carolyncole

